PR #2, opened 2021-03-04
chore: move unity project to folder

BrianAmadori:
The basic idea is to have a folder structure like this, so the JS stuff doesn't get mixed with the Unity project:

```
.
unity-renderer\
unity-interface\
docs\
```


menduz:
I won't see the 5000 files changed, if you make it pass the build I will happily approve it

menduz (approved):

PR #11, opened 2021-03-10
fix: merge from explorer repo

BrianAmadori:
This PR recreates the unity-client folder from explorer repo preserving its history.

All LFS files had to be re-uploaded.

Procedure used:

```
git remote add explorer <explorer-url>
git merge -s ours --allow-unrelated-histories --no-commit explorer/master 
git read-tree --prefix=unity-client/ -u explorer/master:unity-client
git commit -m "first commit" -S
```

This is fine and dandy, but the LFS pointers are still pointing to invalid resources, so we do...

```
git lfs fetch --all explorer
git lfs push origin
```

And lastly we push (if we do this before the LFS upload, it will get stuck):

```
git push
```

Sauces used to come up with this solution:
 
https://stackoverflow.com/questions/6426247/merge-git-repository-in-subdirectory
https://stackoverflow.com/questions/1214906/how-do-i-merge-a-sub-directory-in-git
https://stackoverflow.com/questions/49349974/move-git-directory-from-one-lfs-enabled-repo-to-another-also-lfs
http://gbayer.com/development/moving-files-from-one-git-repository-to-another-preserving-history/


menduz (approved):
cheto

PR #54, opened 2021-03-11
fix: update CI to use 2019.4.19f1 unity image

BrianAmadori:


ci-editmode-test.sh
```diff
@@ -2,13 +2,12 @@
...
-${UNITY_EXECUTABLE:-xvfb-run --auto-servernum --server-args='-screen 0 640x480x24' /opt/Unity/Editor/Unity } \
+xvfb-run --auto-servernum --server-args='-screen 0 640x480x24' $UNITY_PATH/Editor/Unity \
         -batchmode \
-        -logFile "$PROJECT_PATH/editmode-logs.txt" \
-        -projectPath "$PROJECT_PATH" \
+        -logFile "$PROJECT_PATH/playmode-logs.txt" \
         -runTests \
         -testPlatform EditMode \
-        -testResults "$PROJECT_PATH/editmode-results.xml" \
+        -testResults "$PROJECT_PATH/playmode-results.xml" \
```
menduz, line 10:
```suggestion
        -testResults "$PROJECT_PATH/editmode-results.xml" \
```

ci-playmode-test.sh
```diff
@@ -1,17 +1,13 @@
...
 source ci-setup.sh
 
-${UNITY_EXECUTABLE:-xvfb-run --auto-servernum --server-args='-screen 0 640x480x24' /opt/Unity/Editor/Unity } \
+xvfb-run --auto-servernum --server-args='-screen 0 640x480x24' $UNITY_PATH/Editor/Unity \
         -batchmode \
         -logFile "$PROJECT_PATH/playmode-logs.txt" \
         -runTests \
         -testPlatform PlayMode \
         -testResults "$PROJECT_PATH/playmode-results.xml" \
-        -enableCodeCoverage \
```
menduz, line 11: can we use left the coverage turned on?

menduz (changes requested):
I left two comments

ci-playmode-test.sh
BrianAmadori: Can we add this later? I saw that the coverage output files are not deployed or copied anywhere, so it might make no sense right now.

menduz: They are supposed to be uploaded to circle, using the store_artifacts https://circleci.com/docs/2.0/code-coverage/ if not, we can remove it. I thought that it would be nice to have that info available since the beginning 

https://github.com/decentraland/unity-renderer/blob/8a1a4ccd7dfdaa0d2f5d3deaf2da7a1c471c6d94/.circleci/config.yml#L113-L115

menduz (approved):

ci-playmode-test.sh
menduz: @BrianAmadori

PR #190, opened 2021-03-23
feat: publish WebGL artifacts

### LINKED ISSUE #14: Clean logs for preview mode
Right now, the console is full on clutter when working with preview mode. 

For the Unity initialization log, we can override `console.log` on the startup phase of the application to clean it up.

menduz:
Publishes the built artifacts ready to be consumed by the explorer in the URL `https://renderer-artifacts.decentraland.org/branch/${BRANCH}/${PATH}`. Examples: https://renderer-artifacts.decentraland.org/branch/feat/publish-renderer/package.json

### DIFF
--- a/browser-interface/scripts/build.ts
+++ b/browser-interface/scripts/build.ts
@@ -117,6 +117,12 @@ async function createPackageJson() {
         tag: process.env.CIRCLE_TAG,
         commit: process.env.CIRCLE_SHA1,
         branch: process.env.CIRCLE_BRANCH,
+        author: "Decentraland Contributors",
+        license: "Apache-2.0",
+        repository: {
+          type: "git",
+          url: "https://github.com/decentraland/unity-renderer.git",
+        },
       },
       null,
       2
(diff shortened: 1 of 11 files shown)


browser-interface/scripts/build.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  for (let file of glob.sync("**/*", { cwd: basePath, absolute: true })) {
+    copyFile(file, path.resolve(DIST_PATH, file.replace(basePath + "/", "./")))
+  }
+}
+
+async function buildRollup() {
+  ensureFileExists(DIST_PATH, "UnityLoader.js")
+  console.log("> reading UnityLoader.js")
+  const banner = readFileSync(path.resolve(DIST_PATH, "UnityLoader.js")).toString()
```
BrianAmadori, line 58:
```suggestion
  ensureFileExists(DIST_PATH, "DCLUnityLoader.js")
  console.log("> reading DCLUnityLoader.js")
  const banner = readFileSync(path.resolve(DIST_PATH, "DCLUnityLoader.js")).toString()
```

If here we are including the used loader, the correct name would be `DCLUnityLoader.js`.
Note that in the build folder we also have `hls.min.js` for the video streaming plugin. Not sure if that'd break if we don't include it here.

BrianAmadori (approved):

PR #279, opened 2021-04-20
chore: merge explorer repo changes since 3/10/21

BrianAmadori:
After multiple rebase, filter-repo and read-tree failed attempts with a few master -f pushes in-between, just decided to copy-paste the contents of updated explorer repo changes in this PR.


.circleci/config.yml
```diff
@@ -68,7 +68,7 @@ jobs:
             while true; do git reset --hard HEAD && git clean -fd && git lfs pull && break; done
       - save_cache:
           name: Store large files in cache
-          key: git-lfs-{{ .Branch }}
+          key: git-lfs-{{ .Branch }}-{{ epoch }}
```
menduz, line 71: Should this `{{epoch}}` be merged to master? It would make the clone very heavy in size. lfs caché is content-addressed s there should not be any conflicts

.circleci/config.yml
```diff
@@ -217,7 +217,7 @@ jobs:
           destination: Builds
       - save_cache:
           name: Store library
-          key: library-{{ .Branch }}
+          key: library-2020-3-{{ .Branch }}-{{ epoch }}
```
menduz, line 220: Same as line 71

menduz (changes requested):

.circleci/config.yml
BrianAmadori: Epoch is 100% necessary here. If we don't put this the Library cache will not be regenerated when we push new assets on the same branch.

.circleci/config.yml
BrianAmadori: Same as below, but maybe we can pass of this epoch as the lfs overhead is quite light.

PR #309, opened 2021-04-28
fix: renderer publish fix

BrianAmadori:
This PR changes the CI so we publish NPM packages for all branches, not only master.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -310,12 +310,7 @@ workflows:
             - build-unity
 
       - publish-renderer:
-          filters:
-            branches:
-              only:
-                - master
-            tags:
-              only: /^\d+\.\d+\.\d+(.*)/ # npm tag + latest
+          <<: *all_branches_and_tags
           requires:
             - build-unity
             - build-js-interface
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,10 +78,7 @@ jobs:
           paths:
             - "*"
 
-  ############
-  ## THIS IS TEMPORARILY DISABLED AND RUNS ONLY IN THE "editmode-branch" branch
-  ############
-  editmode-tests:
+  build-unity:
     <<: *working_directory_root
     resource_class: xlarge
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -98,11 +95,62 @@ jobs:
       - restore_cache:
           name: Restore hashed files if they exist
           keys:
-            - unity-editmode-2020-3-{{ checksum "../.unitysources-checksum" }}
+            - unity-build-{{ checksum "../.unitysources-checksum" }}
       - restore_cache: &RESTORE_LIBRARY_CACHE
           name: Restore library if exists
           keys:
             - library-2020-3-{{ .Branch }}
+      - run:
```
BrianAmadori, line 103: Some of these steps were being referenced to steps in the `editmode-tests` job. Changed it so the "master" steps are in `unity-build`.

browser-interface/scripts/publish.ts
```diff
@@ -7,8 +7,18 @@ const DIST_ROOT = resolve(__dirname, "../dist")
...
+
+  if (branchName === "master") {
     await publish(["latest"], "public", DIST_ROOT)
+  } else {
+    const circleBuildNum = process.env.CIRCLE_BUILD_NUM;  
+    let match = /[^a-zA-Z0-9-]/g;
+    const cleanedBranchName = branchName!.replace(match, "-")
+    const tagName = `1.0.0.${circleBuildNum}${cleanedBranchName}`
+
+    await publish([tagName], "public", DIST_ROOT);
```
menduz, line 21: What is the need behind this change? it will overload the package registry. To test the explorer this _wont_ be needed as the artifacts are uploaded to S3 (which is cheaper and easier to manage than NPM)

BrianAmadori: As we recently changed the `update-renderer` make command kernel side, I had a bobazo and forgot we changed the overall approach. Closing in 3...2...1...

PR #333, opened 2021-04-30
chore: Create an AppMode to manage several states of the game

sandrade-dcl:
### Why?
We often need to implement specific behaviours depending on different situations that happens in our game (for example: enable/unable certain buttons of the taskbar when the Builder In World editor is open/closed). Currently we are managing these cases implementing very specific logics in specifics places, making it difficult to trace and also making it difficult remember that we possibly should update this logic in future changes.

To avoid this, we would need some kind of system that let us set in any moment a "state" for the game. For example: I open the Builder In World editor, then I would like to set the game state as something that tell us "_We are executing the Builder In World editor_". In this way we would be able to know in any moment what is the game state and apply a more generic logic in the side of all those consumers that need to do something depending on specific states.

### Implementation of the `appMode` variable
We have created a new variable `appMode` in our `DataStore`. The type of this new variable is a simple enum that contains the different modes (or states) of our application. For now we have created only 2 modes (we should add new modes to the enum as we need it):
- `DEFAULT`: to define the "normal" state of the game.
- `BUILDER_IN_WORLD_EDITION`: to define when we are editing a scene with the Builder In World feature.

In the consumers side (for example: the `TaskbarButton` class) the only thing they should do is to subscribe to `DataStore.i.appMode.OnChange` and implement the needed logic (for the case of the `TaskbarButton`: enable or unable the button when we enter in specifics game modes).

### Testing
1. Go to: https://play.decentraland.zone/?index.html&renderer=urn:decentraland:off-chain:renderer-artifacts:chore/Create-an-AppMode-to-manage-several-states-of-the-game&ENV=zone&ENABLE_BUILDER_IN_WORLD&position=100,100
3. Notice that the explore and tutorial buttons are been deactivated in the taskbar.
image
4. Exit from the BiW editor.
5. Notice that the explore and tutorial buttons are been activated in the taskbar.
image


pbosio (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldController.cs
```diff
@@ -550,6 +539,7 @@ public void ExitEditMode()
         RenderSettings.skybox = previousSkyBoxMaterial;
 
         OnExitEditMode?.Invoke();
+        DataStore.i.appMode.Set(AppMode.DEFAULT);
```
Suduck, line 542: Maybe we should take into account for the future that we should be able to go back to the previous mode so we can do a "Pile"  of states. 

For example Defautl state -> BuilderInWorld -> Preview Mode

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/TaskbarHUD/TaskbarButton.cs
```diff
@@ -26,6 +29,26 @@ public void Initialize()
...
+    {
+        bool isInteractable = true;
+        foreach (var mode in modesWhereNotInteractable)
+        {
+            if (current == mode)
+            {
+                isInteractable = false;
+                break;
+            }
+        }
```
Suduck, line 49: Is not better to implement the counterpart logic here? You should specify in which modes it should work instead of in which modes it doesn't work.

This way in the future if you are developing a new feature you only need to take care about the things that needs to work and not about the things that souldn't work.

In the future if we have like 34 systems they will probably work independent of each other, so it is more scalable the other way

Maybe is only my opinion!

Suduck (approved):
LGTM! Only a suggestion that you can feel free to implement it or not!

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/TaskbarHUD/TaskbarButton.cs
pbosio: I like this suggestion, It could minimize a lot of possible issues in the future

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldController.cs
sandrade-dcl: Hmmm maybe in the future we could need something like this, but for now I don't see problem to give the responsibility of set each specific mode to each module (BiW, preview mode, etc.). At the moment I wouldn't like to overimplement the system, if we need it in the future it will not be something difficult to adapt.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/TaskbarHUD/TaskbarButton.cs
sandrade-dcl: Yeah, I guess it can be interpreted in both ways. For me the fact of disable a feature is always the exception, in fact in the case of the taskbar we only need to disable some buttons in a very specific case (when the BiW editor is open) and the rest of situations should be actived. But I understand what you say... I don't know, I think we can let each consumer decide how to handle it.

sandrade-dcl: Well if you both agree, I'm going to change it 👍

pbosio: Yeah, the main reason for me about this could be that if you are implementing a new feature, it would be easy to notice that something is not working when it should than noticing that something is working when it shouldn't :P

sandrade-dcl: But take into account that, with your approach, when we add any new AppMode, we are going to need to go to each system (taskbar and anye others that is using the enum) and add the new mode in its "list of compatible modes".

sandrade-dcl: Done in https://github.com/decentraland/unity-renderer/pull/333/commits/52af5e303ad1ed2f17db2974b2f73db3e6df7359. Thanks guys!

PR #338, opened 2021-05-03
fix: BiW scale tool is not working correctly

### LINKED ISSUE #316: Scale gizmos on all axis is not working correctly
Scale gizmos is working strange when you try to scale in the three axis at the same time

### LINKED ISSUE #317: Scale on Z axis with multiple entities
If you have multiple entities selected and scale in the Z axis, they move as well when they should only scale

sandrade-dcl:
## Fixed bugs
- [x] **Scale gizmos on all axis is not working correctly** (Fixes #316)
_Scale gizmos is working strange when you try to scale in the three axis at the same time_
- [x] **Scale on Z axis with multiple entities** (Fixes #317)
_If you have multiple entities selected and scale in the Z axis, they move as well when they should only scale_

## Testing
1. Go to: https://play.decentraland.zone/?index.html&renderer=urn:decentraland:off-chain:renderer-artifacts:fix/BiW-scale-is-not-working-correctly&ENV=zone&ENABLE_BUILDER_IN_WORLD&position=100,100

### DIFF
--- a/unity-renderer/Assets/Builder/Scripts/Gizmos/DCLBuilderScaleGizmo.cs
+++ b/unity-renderer/Assets/Builder/Scripts/Gizmos/DCLBuilderScaleGizmo.cs
@@ -27,7 +27,7 @@ namespace Builder.Gizmos
             if (axis == axisProportionalScale)
             {
                 scaleDirection = Vector3.one;
-                float inputDirection = lastMousePosition.y - initialMousePosition.y;
+                float inputDirection = (lastMousePosition.x - initialMousePosition.x) + (lastMousePosition.y - initialMousePosition.y);
                 if (inputDirection < 0)
                 {
                     scaleDirection = -Vector3.one;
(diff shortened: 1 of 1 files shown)


Suduck (approved):

Suduck:
Good job!

unity-renderer/Assets/Builder/Scripts/Gizmos/DCLBuilderScaleGizmo.cs
```diff
@@ -22,12 +24,38 @@ public override void Initialize(Camera camera, Transform cameraTransform)
         public override void SetSnapFactor(DCLBuilderGizmoManager.SnapInfo snapInfo) { snapFactor = snapInfo.scale; }
 
         public override float TransformEntity(Transform entityTransform, DCLBuilderGizmoAxis axis, float axisValue)
+        {
+            // In order to avoid to make the scale of each selected entity dependent of the 'entityTransform' parent,
+            // we temporally move all entities to the local position (0,0,0) before calculate the new scale.
+            foreach (Transform entity in entityTransform)
+            {
+                entitiesOriginalPositions.Add(entity, entity.transform.position);
+                entity.transform.localPosition = Vector3.zero;
```
pravusjif, line 33: this should be `position` instead of `localPosition` right? let's use the same position term to be consistent, even though probably in this case `poition` and `localPosition` do the same thing if the entity doesn't have a parent GO

pravusjif (approved):

unity-renderer/Assets/Builder/Scripts/Gizmos/DCLBuilderScaleGizmo.cs
sandrade-dcl: But in this case I want to set its local position, not the world one, because the entity has a parent and I want to place temporally the child to the (0,0,0) position relative to the parent.

PR #349, opened 2021-05-03
refactor: builder projects panel scene cards, sections and search handler

pbosio:


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Scripts/SectionController/SearchInfo/SearchInfo.cs
```diff
@@ -0,0 +1,74 @@
...
+    private string[] keywords;
+
+    public SearchInfo() { keywords = new string[2]; }
+
+    void ISearchInfo.SetId(string id) { this.id = id; }
+
+    void ISearchInfo.SetName(string name)
+    {
+        this.name = name;
+        keywords[0] = name;
```
Suduck, line 44: This is a little strange, any reason for array usage instead of a list?

Suduck (approved):
Good job! I will like to test the panel but I guess that the task button is hidden, should we enable it if the flag is present on the URL?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Scripts/SectionController/SearchInfo/SearchInfo.cs
pbosio: I choose to use an array cause it was a fixed size, I didn't have to heavily manipulate data stored there and because it wasn't clear at the moment if I should use an enum or constant to name the indices like `keywords[NAME] = name` which would be super strange to do using a list :P

pbosio:
> Good job! I will like to test the panel but I guess that the task button is hidden, should we enable it if the flag is present on the URL?

you can test it here: https://play.decentraland.zone/branch/feat/builderprojectspanel/index.html?ENABLE_BUILDER_IN_WORLD&ENV=org&position=-149%2C-144&realm=baldr-amber&renderer=urn%3Adecentraland%3Aoff-chain%3Arenderer-artifacts%3Afeat%2Fbuilderprojectspanel

pravusjif (approved):

PR #350, opened 2021-05-03
fix: pool extra clones on initialization

pravusjif:
original PR: https://github.com/decentraland/explorer/pull/2329

* Added mechanism to avoid re-creating OnPointerEvent Colliders on shapes that dont change their mesh
* Added missing tests


sandrade-dcl (approved):
LGTM! Be careful with the tests, that are currently failing.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/PoolManager/Pool.cs
```diff
@@ -72,18 +71,7 @@ public void ForcePrewarm()
...
-                {
-                    Instantiate();
-                }
-
-                Instantiate();
-            }
-            else if (unusedObjects.Count == 0)
+            if (unusedObjectsCount == 0)
             {
                 Instantiate();
```
BrianAmadori, line 76: This was added so we have additional prewarmed objects before the camera is activated. If this is breaking the tests we should just set the `PoolManager.i.initializing` to true in the tests instead of removing this code.

BrianAmadori (changes requested):

Suduck (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/PoolManager/Pool.cs
pravusjif: It's not breaking the tests, it's introducing the bug of more cloned instances while the rendering is off

BrianAmadori: That's not a bug, its a (badly documented) feature. `PoolableComponentFactory` prewarms only the entity component pools. 

This feature adds additional pooled instances of whichever assets are added to the pool, with the heuristic of "this object is used a lot, so we may have to instantiate more after the loading screen finishes. Let's instantiate them now to save the hiccups later".

Example 1:
Let's say you spawn in a zone with a lot of avatars that use base wearables. In that case the heuristic would kick in, and as base wearables are quite common, there's a high probability of needing more after the loading finishes. When that's the case, the wearables would be already instanced ahead of time.

Example 2:
You spawn in 64,-64 or in a zone that makes heavy use of builder assets. When you walk around, the most used builder assets would be already have additional items prewarmed ahead of time. When you go to neighboring scenes that also use builder assets there's high probability of needing more of those, and instantiate time would be saved.

pravusjif: Thx for the info, I added a comment regarding that and re-added the removed code. Check it out.

BrianAmadori (approved):
LGTM

PR #358, opened 2021-05-04
fix: Floor loading placeholder is visible after the floor is loaded s…

### LINKED ISSUE #342: Floor loading placeholder is visible after the floor is loaded sometimes
If the floor is not loaded and you try to load another floor, the loading placeholder of the first floor will stay

We should ensure that there is no loading placeholder after all the floors has been loaded

sandrade-dcl:
## What this PR includes?
- **BUG FIX:** "Floor loading placeholder is visible after the floor is loaded sometimes" (Fixes #342)
_If the floor is not loaded and you try to load another floor, the loading placeholder of the first floor will stay._

## How to test it?
1. Go to: https://play.decentraland.zone/?index.html&renderer=urn:decentraland:off-chain:renderer-artifacts:fix/BiW-Floor-loading-placeholder-is-visible-after-the-floor-is-loaded-sometimes&ENV=zone&ENABLE_BUILDER_IN_WORLD&position=100,100
3. Open the catalog panel (right side of the screen).
4. Select different floors several times in a row without letting them finish loading.
5. Notice that the loading spinner is always being hidden after the last floos has been loaded.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Controllers/BIWFloorHandler.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Controllers/BIWFloorHandler.cs
@@ -33,9 +33,21 @@ public class BIWFloorHandler : BIWController
     private CatalogItem lastFloorCalalogItemUsed;
     private readonly Dictionary<string, GameObject> floorPlaceHolderDict = new Dictionary<string, GameObject>();
 
-    private void Start() { meshLoadIndicator.SetCamera(Camera.main); }
+    private void Start()
+    {
+        builderInWorldEntityHandler.OnEntityDeleted += BuilderInWorldEntityHandler_OnEntityDeleted;
 
-    private void OnDestroy() { Clean(); }
+        meshLoadIndicator.SetCamera(Camera.main);
+    }
+
+    private void OnDestroy()
+    {
+        builderInWorldEntityHandler.OnEntityDeleted -= BuilderInWorldEntityHandler_OnEntityDeleted;
+
+        Clean();
+    }
+
+    private void BuilderInWorldEntityHandler_OnEntityDeleted(string entityId) { dclBuilderMeshLoadIndicatorController.HideIndicator(entityId); }
 
     public void Clean()
     {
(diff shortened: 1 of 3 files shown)


pravusjif (approved):

BrianAmadori (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Controllers/BIWFloorHandler.cs
```diff
@@ -92,6 +92,7 @@ public void CreateFloor(CatalogItem floorSceneObject)
         {
             DCLBuilderInWorldEntity decentralandEntity = biwCreatorController.CreateCatalogItem(floorSceneObject, WorldStateUtils.ConvertPointInSceneToUnityPosition(initialPosition, parcel), false, true);
             decentralandEntity.rootEntity.OnShapeUpdated += OnFloorLoaded;
+            dclBuilderMeshLoadIndicatorController.HideAllIndicators();
```
Suduck, line 95: This won't work in a scene with severals parcels since it will hide the loading indicators of the previous instanciated placeholders. 

I think a better approach will be to delete the complete loading placeholder when deleting the previous floor entities, this way we are not only hiding the indicators but also removing the placeholder that will stay in the world

Suduck (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Controllers/BIWFloorHandler.cs
sandrade-dcl: You're right, thanks for the warning! I have changed the approach, you can see it in https://github.com/decentraland/unity-renderer/pull/358/commits/d4125b1ce8409eb8166fce5e8b46d716a4fe546b

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Controllers/BuilderInWorldEntityHandler.cs
```diff
@@ -709,6 +710,7 @@ public void DeleteEntity(DCLBuilderInWorldEntity entityToDelete, bool checkSelec
         hudController?.RefreshCatalogAssetPack();
         EntityListChanged();
         builderInWorldBridge?.RemoveEntityOnKernel(idToRemove, sceneToEdit);
+        OnEntityDeleted?.Invoke(entityToDelete.rootEntity.entityId);
```
Suduck, line 713: I think this event should be before the line 787, since we destroy the entity gameObject it can be null, also, if someone want to do changes on the entity before it gets finally delete it should be able to do it!

Suduck (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Controllers/BuilderInWorldEntityHandler.cs
sandrade-dcl: Ok, done in https://github.com/decentraland/unity-renderer/pull/358/commits/ea850879045eb077e80380241672a91a48801b4b

PR #395, opened 2021-05-07
fix: ssao being seen behind transparencies

BrianAmadori:
## What this PR includes?

This PR fixes the issue in which SSAO could be seen behind seemingly opaque objects.

## How to test it?

1. Go to: https://play.decentraland.zone/?index.html&renderer=urn:decentraland:off-chain:renderer-artifacts:{branch_name}&{desired_url_params}


pravusjif (approved):
Awsome!

It can also be tested in Space Office windows glass at:
https://play.decentraland.zone/index.html?ENV=zone&position=64%2C-67&renderer=urn%3Adecentraland%3Aoff-chain%3Arenderer-artifacts%3Afix%2Fssao-being-seen-behind-transparencies

and master (bugged) would be:
https://play.decentraland.zone/index.html?ENV=zone&position=64%2C-67

kuruk-mm:
image

image


https://user-images.githubusercontent.com/12563266/117680915-fa1f7900-b187-11eb-883a-83d41e7deb8b.mp4

I'm seeing that little image behind that wall (is like a rat). I don't know if it is related with this.

It only renders when we're in the scene, so I don't know exactly what it is.

BrianAmadori:
@kuruk-mm that artifact is something related to the material that's used to render the NFT collectibles, not the SSAO. Maybe we can file another issue for that one.

kuruk-mm (approved):

PR #397, opened 2021-05-07
fix: builder projects panel the graph and catalyst fetch cache

pbosio:
* **add**: class `DataCache<T>` to store data that will be cleaned after a set period of time
* **refactor**: `Catalyst` and `TheGraph` classes to be able to set a custom time for caching apis' response
* **refactor**: `BuilderProjectsPanelController` class to reduce the time of deployed scenes cache and fetch for updates on interval
* **refactor**: promote `DeployedScene` and `LandWithAccess` classes to their own file script.
* **fix**: `SceneCardView` trying to fetch thumbnail even if it was the same as the previous thumbnail

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Scripts/BuilderProjectsPanelController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Scripts/BuilderProjectsPanelController.cs
@@ -1,5 +1,5 @@
 ﻿using System;
-using System.Collections.Generic;
+using System.Collections;
 using System.Linq;
 using DCL;
 using DCL.Helpers;
@@ -14,6 +14,10 @@ public class BuilderProjectsPanelController : IHUD
     private const string TESTING_ETH_ADDRESS = "0x2fa1859029A483DEFbB664bB6026D682f55e2fcD";
     private const string TESTING_TLD = "org";
     private const string VIEW_PREFAB_PATH = "BuilderProjectsPanel";
+
+    private const float CACHE_TIME_LAND = 5 * 60;
+    private const float CACHE_TIME_SCENES = 1 * 60;
+    private const float REFRESH_INTERVAL = 2 * 60;
 
     internal readonly IBuilderProjectsPanelView view;
 
@@ -30,6 +34,8 @@ public class BuilderProjectsPanelController : IHUD
     private ICatalyst catalyst;
 
     private bool isInitialized = false;
+    private bool isFetching = false;
+    private Coroutine fetchDataInterval;
     private Promise<LandWithAccess[]> fetchLandPromise = null;
 
     public BuilderProjectsPanelController() : this(
@@ -43,6 +49,8 @@ public class BuilderProjectsPanelController : IHUD
 
     public void Dispose()
     {
+        StopFetchInterval();
+
         DataStore.i.HUDs.builderProjectsPanelVisible.OnChange -= OnVisibilityChanged;
         view.OnClosePressed -= OnClose;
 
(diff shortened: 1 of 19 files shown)


sandrade-dcl (approved):
LGTM!

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Scripts/BuilderProjectsPanelController.cs
```diff
@@ -116,8 +124,13 @@ private void OnVisibilityChanged(bool isVisible, bool prev)
...
+            StartFetchInterval();
             sectionsController.OpenSection(SectionId.SCENES_DEPLOYED);
         }
+        else
+        {
+            StopFetchInterval();
+        }
     }
 
     private void OnClose() { SetVisibility(false); }
```
BrianAmadori, line 136: Maybe we can call `StopFetchInterval()` when the window is closed or disposed?

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/DeployedScenesFetcher/LandWithAccess.cs
```diff
@@ -0,0 +1,25 @@
...
+    public string id => rawData.id;
+    public LandType type => rawData.type;
+    public LandRole role => rawData.role;
+    public int size => rawData.size;
+    public string name => rawData.name;
+    public string owner => rawData.owner;
+
+    public List<DeployedScene> scenes;
+    public Vector2Int[] parcels;
+    public Vector2Int @base;
```
BrianAmadori, line 16: Can't we rename this to `baseCoords` to avoid the nasty verbatim operator?

unity-renderer/Assets/Scripts/MainScripts/DCL/ServiceProviders/Catalyst/Catalyst.cs
```diff
@@ -41,15 +50,15 @@ public interface ICatalyst : IDisposable
...
 public class Catalyst : ICatalyst
 {
-    private const float CACHE_TIME = 5 * 60;
+    private const float DEFAULT_CACHE_TIME = 5 * 60;
     private const int MAX_POINTERS_PER_REQUEST = 90;
 
     public string contentUrl => realmContentServerUrl;
 
     private string realmDomain = "https://peer.decentraland.org";
     private string realmContentServerUrl = "https://peer.decentraland.org/content";
```
BrianAmadori, line 59: Remember that we have `ContentServerUtils.cs` for this kind of stuff. Maybe we should unify this code with that one somehow.

pravusjif (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/DeployedScenesFetcher/LandWithAccess.cs
pbosio: renamed https://github.com/decentraland/unity-renderer/pull/397/commits/1dcf87e94bcc0287a2e5753f3b525382af74fb5f

unity-renderer/Assets/Scripts/MainScripts/DCL/ServiceProviders/Catalyst/Catalyst.cs
pbosio: Those url are just placeholder it real value is set on class constructor or/and when user moves to other realm https://github.com/decentraland/unity-renderer/blob/3274c0ab77663f65eeb40c68c0498eb5ab88e9d0/unity-renderer/Assets/Scripts/MainScripts/DCL/ServiceProviders/Catalyst/Catalyst.cs#L63-L71
I leave it there cause I though it would be easy to understand what those string are for instead of just leaving them as `null`, but I can change them back to `null` if you like

PR #423, opened 2021-05-11
chore: clean AB Converter core code

pravusjif:
Cleaned up some code and added 1 helpful menu item for dumping scene ids

### DIFF
--- a/unity-renderer/Assets/ABConverter/AssetBundleMenuItems.cs
+++ b/unity-renderer/Assets/ABConverter/AssetBundleMenuItems.cs
@@ -45,9 +45,10 @@ namespace DCL
 
         public static void DumpEmptyParcels(string folderName = "empty-scenes")
         {
+            // The paths in this method work if the unity-renderer repo is at the same location as the explorer repo
             string indexJsonPath = Application.dataPath;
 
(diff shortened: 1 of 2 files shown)


pbosio (approved):

unity-renderer/Assets/ABConverter/AssetBundleMenuItems.cs
```diff
@@ -45,9 +45,10 @@ public class EmptyParcels
 
         public static void DumpEmptyParcels(string folderName = "empty-scenes")
         {
+            // The paths in this method work if the unity-renderer repo is at the same location as the explorer repo
             string indexJsonPath = Application.dataPath;
 
-            indexJsonPath += $"/../../kernel/static/loader/{folderName}/index.json";
+            indexJsonPath += $"/../../../explorer/kernel/static/loader/{folderName}/index.json";
```
BrianAmadori, line 51: As this is a local path and should be set manually each time we want to use it, I'd add a constant with the `/../../../explorer/kernel` bit, and clarify that in a comment.

pravusjif: Addressed, check it out!

PR #425, opened 2021-05-11
fix: BiW Editor dropping an entity in the scene should place it right there

### LINKED ISSUE #383: BiW Editor dropping an entity in the scene should place it right there
If you drag and drop an entity from the collection menu the entity is not created when you drop but when you click again in the scene. The entity should be created right when you drop it in the scene.

sandrade-dcl:
## What this PR includes?
If you drag and drop an entity from the collection menu the entity is not created when you drop but when you click again in the scene. The entity should be created right when you drop it in the scene (Fixes #383).

## How to test it?
1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/BiW-Editor-dropping-an-entity-in-the-scene-should-place-it-right-there&ENV=zone&ENABLE_BUILDER_IN_WORLD&position=100,100
3. Drag some items from the Catalog to the scene.
4. Notice that the item is directly placed in the floor.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Controllers/BIWCreatorController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Controllers/BIWCreatorController.cs
@@ -1,11 +1,8 @@
 using DCL;
 using DCL.Components;
 using DCL.Configuration;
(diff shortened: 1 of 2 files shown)


pravusjif (approved):
also tested in-build

Issue #416, opened 2021-05-10
Move environment to its own assembly


olavra:
We need to move the new environment class to a new assembly  to avoid cyclic reference

- [x] Create mock branch and refactor relentlessly to make plan and start again
- [x] Move needed auxiliar systems to new assemblies (Promise, ContentServerUtils and some Utils methods)
- [x] Move all ECS and SceneController interfaces to new assemblies (IDCLEntity, IComponent, IParcelScene, ISceneController, etc)
- [x] Move all WorldRuntimeContext interfaces to new assemblies 
- [x] Move all PlatformContext interfaces to new assemblies 
- [x] Move Environment to new assembly
- [ ] Apply fixes and review feedback

PR #429, opened 2021-05-12
refactor: move SceneController and ECS interfaces to own assembly

BrianAmadori:
## What this PR includes?

This PR is part of the following series to address https://github.com/decentraland/unity-renderer/issues/416:
* refactor: move SceneController and ECS interfaces to own assembly https://github.com/decentraland/unity-renderer/pull/429
* refactor: put all runtime context interfaces into own assembly https://github.com/decentraland/unity-renderer/pull/432
* refactor: put all platform context interfaces into own assembly https://github.com/decentraland/unity-renderer/pull/431
* refactor: put environment into own assembly https://github.com/decentraland/unity-renderer/pull/433

---

This PR manages to put all interfaces related to scene runtime (`IParcelScene`, `IDCLEntity`, `ISceneController`, `IComponent`, etc) into a single assembly: `WorldRuntimeInterfaces`. This is a step needed in order to move `Environment` to its own assembly.

- Logic changes are almost non-existent. This is pretty much code and architecture reorganization.
- To better review this PR, please do it commit-per-commit. The commit order was planned ahead for review easiness and some commits have explanations about the changes.
- A few files were moved to new folders. The commits that move files just include the movement diff to work well with git move detection.
- A lot of `.asmdef` files were changed. To refactor them faster, the `.asmdef` files were edited by hand. (seems that Rider still can't modify them through the refactor commands...).

## How to test it?
1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:refactor/move-world-runtime-interfaces-to-own-assembly&ENV=org
2. The client should run correctly. If the world loads then most likely this PR is working correctly.

## Our Code Review Standards


unity-renderer/Assets/ABConverter/Utils.cs
```diff
@@ -16,6 +16,27 @@
 
 namespace DCL.ABConverter
 {
+    public static class MeshUtils
```
BrianAmadori, line 19: Duplicated on purpose so the ABConverter doesn't have any reference from our main project.

sandrade-dcl (approved):
What a huge PR! After making a general review I have not seen anithing weird in the code. I love the new approach of separated assemblies. I feel more structuring and less caos in our project now, so thanks!

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/TaskbarHUD/TaskbarHUDController.cs
```diff
@@ -551,8 +551,10 @@ public void ShowTutorialOption(bool isActive)
             view.moreMenu.ShowTutorialButton(isActive);
     }
 
-    private void SceneController_OnNewPortableExperienceSceneAdded(GlobalScene newPortableExperienceScene)
+    private void SceneController_OnNewPortableExperienceSceneAdded(IParcelScene scene)
     {
+        GlobalScene newPortableExperienceScene = (GlobalScene)scene;
```
AjimenezDCL, line 556: I'd add some checks here:

`if(!scene is GlobalScene newPortableExperienceScene) return;`

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Interfaces/ISceneController.cs
```diff
@@ -0,0 +1,45 @@
+﻿using System;
+using DCL.Controllers;
+using DCL.Models;
+
+namespace DCL
+{
+    public delegate void OnOpenNFTDialogDelegate(string assetContractAddress, string tokenId, string comment);
```
AjimenezDCL, line 7: Not super happy with this delegate here, but it wouldnt make sene to tackle down this technical debt in this PR.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Scene/LoadingFeedbackController.cs
```diff
@@ -47,7 +47,11 @@ private void OnDestroy()
         CommonScriptableObjects.rendererState.OnChange -= RendererState_OnChange;
     }
 
-    private void SceneController_OnNewSceneAdded(ParcelScene scene) { scene.sceneLifecycleHandler.OnStateRefreshed += Scene_OnStateRefreshed; }
+    private void SceneController_OnNewSceneAdded(IParcelScene scene)
+    {
+        var parcelScene = (ParcelScene)scene;
```
AjimenezDCL, line 52: I'd add some checks here:
`if (!scene is ParcelScene parcelScene)`

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/TaskbarHUD/TaskbarHUDController.cs
AjimenezDCL: On the other hand something feels out of place here (same in LoadingFeedbackController).

I find strange that the listener to the `OnNewPortableExperienceSceneAdded` event has to know that the `IParcelScene` provided is also a `GlobalScene`. Maybe we should have an `IGlobalScene` as well as part of our WorldRuntime interfaces. 

In any case this issue can be addressed in other PR, if you agree with the concern, of course.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
```diff
@@ -142,14 +142,15 @@ private void OnWorldReposition(Vector3 current, Vector3 previous)
             MinimapMetadataController.i?.UpdateMinimapUserInformation(avatarUserInfo);
         }
 
-        private void OnEntityTransformChanged(DCLTransform.Model updatedModel)
+        private void OnEntityTransformChanged(object newModel)
         {
-            lastAvatarPosition = updatedModel.position;
+            DCLTransform.Model newTransformModel = (DCLTransform.Model)newModel;
```
AjimenezDCL, line 147: We should ensure the type is correct here as in my previous requestes.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldController.cs
```diff
@@ -321,17 +324,18 @@ public VoxelEntityHit GetCloserUnselectedVoxelEntityOnPointer()
...
-    private void NewSceneAdded(ParcelScene newScene)
+    private void NewSceneAdded(IParcelScene newScene)
     {
         if (newScene.sceneData.id != sceneToEditId)
             return;
 
         Environment.i.world.sceneController.OnNewSceneAdded -= NewSceneAdded;
 
-        sceneToEdit = Environment.i.world.state.GetScene(sceneToEditId) as ParcelScene;
+        sceneToEdit = (ParcelScene)Environment.i.world.state.GetScene(sceneToEditId);
```
AjimenezDCL, line 338: I'm a bit concerned of all these castings we have to make regarding the scenes. It feels like something is missing somewhere, as if our extracted interfaces for the scenes were a bit incompleted.

AjimenezDCL (approved):
Great job! I really appreciate the distribution by commits it ended up being way easier that I could expect for a 205 files PR.

I left some concerns regarding these explicit casts on the `ParcelScene` interfaces. I wont block the PR but I would like to know if it's techincal-debt you have under your radar for a future refactor or I'm missing something.

pravusjif (approved):
We should update these PRs with master since something in the wearables support changed and now these PRs fail to load unless we use an updated kernel branch like the followwing:
https://play.decentraland.zone/branch/fix/register-name-for-scene-storage-controller/index.html?ENV=org&realm=baldr-amber&renderer=urn%3Adecentraland%3Aoff-chain%3Arenderer-artifacts%3Arefactor%2Fmove-world-runtime-interfaces-to-own-assembly

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/TaskbarHUD/TaskbarHUDController.cs
BrianAmadori: Yes, I think this issue should be addressed on another PR. The code already relied on the `GlobalScene` cast to succeed and wasn't doing any checks, so the code would fail in any case. I saw that and realized that just using static casting would have the same behaviour, that's the reasoning that led me to the change.

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Interfaces/ISceneController.cs
BrianAmadori: If the interface needs it, it has to be here. Not sure of what we could do better in this case as moving it to another assembly seems overkill.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
BrianAmadori: The only dispatcher of this event (`DCLTransform`) will always come with `DCLTransform.Model`. 

I took the shortcut of changing the event signature to pass `object` type because keeping the reference to the specific component from the entity will give problems in the future. This problems will materialize when trying to put `DCLTransform` in a single assembly for example.

This exposes a design flaw: we shouldn't be having component-specific events at the entity layer, and those events should belong elsewhere.

BrianAmadori: Adding some tech debt to address this in the future.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldController.cs
BrianAmadori: This was another shortcut. `BuilderInWorldController` needs the `ParcelScene` type deep down in the logic to call the scene messages manually. Ideally, it should use some kind of interface to do this because messages processing shouldn't belong in `ParcelScene`.

Ty for reminding me of this issue, will add it as tech debt because the fix is not trivial.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/TaskbarHUD/TaskbarHUDController.cs
BrianAmadori: Added the additional check just in case.

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Interfaces/ISceneController.cs
AjimenezDCL: Yeah but at the same time, it seems strange that the `ISceneController`  interface needs an `OpenNFTDialgo` delegate, doesnt it?

PR #430, opened 2021-05-12
feat: BiW - Clicking on an entity of the collection hides the collection menu

### LINKED ISSUE #382: BiW Editor clicking on an entity of the collection hides the collection menu
While creating new entities using the Collections menu, it usually happens that entities get created behind the panel. To avoid this, the panel will be hidden as soon as the player clicks on an entity.

This won't be the behaviour while dragging and dropping.

Make 2 tests:
- [ ] One hiding the panel
- [ ] One making the panel opacity something like 0.2

sandrade-dcl:
### Fixes #382

**PROBLEM:** While creating new entities using the Collections menu, it usually happens that entities get created behind the panel.

**SOLUTION:** To avoid this, the panel will become semi-transparent as soon as the player clicks on an entity. This won't be the behaviour while dragging and dropping.

## How to test it?
1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:feat/BiW-Clicking-on-an-entity-of-the-collection-hides-the-collection-menu&ENV=zone&ENABLE_BUILDER_IN_WORLD&position=100,100
3. Select items in the catalog.
4. Notice that the entity information panel enters in "transparency mode" to allow you move the item behind.

## Our Code Review Standards


AjimenezDCL (approved):
Looks good to me! Nice test suite!

pravusjif (approved):

PR #457, opened 2021-05-18
fix: Fixes on Undo/Redo and Shortcuts

### LINKED ISSUE #441: BiW Editor change shortcuts
Duplicate = Shift + D
Reset Camera = Shift + C
Reset  = Shift + R

### LINKED ISSUE #442: Undo/redo in the toolbar is not working
I can trigger the undo/redo by using the shortcuts but not using the buttons on the toolbar.

sandrade-dcl:
### Fixes #442
We can trigger the undo/redo by using the shortcuts but not using the buttons on the toolbar.

### Fixes #441
- [ ] Duplicate = Shift + D
- [ ] Reset Camera = Shift + C
- [ ] Reset  = Shift + R
- [ ] Controls panel shortcut needs to be C instead of N

image

- [ ] Snap Icon does not have a tooltip with text + shorcut

image

- [ ] + to zoom in, - to zoom out

## How to test it?

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/BiW-undo-redo-and-shortcuts-fixes&ENV=zone&ENABLE_BUILDER_IN_WORLD&position=100,100

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldController.cs
@@ -241,6 +241,7 @@ public class BuilderInWorldController : MonoBehaviour
         biwFloorHandler.Init();
         bIWInputHandler.Init();
         biwSaveController.Init();
+        actionController.Init();
     }
 
     private void StartTutorial() { TutorialController.i.SetBuilderInWorldTutorialEnabled(); }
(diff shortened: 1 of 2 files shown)


pbosio (commented):
Codewise it looks good. I didn't approved it cause I'm not entirely sure if the following issue is a blocker for this PR.
those zoom shortcuts are not going to work in several keyboard layouts.
as an example, on my mac-english-no-numpad keyboard I can zoom out but I can't zoom in (cause I need to press SHITF+KEY for the "+").
so I take this opportunity to, yet again, recommend that we should replace unity's input system for the new version :P
EDIT: I've just added an issue to the board about this :)

sandrade-dcl:
> Codewise it looks good. I didn't approved it cause I'm not entirely sure if the following issue is a blocker for this PR.
> those zoom shortcuts are not going to work in several keyboard layouts.
> as an example, on my mac-english-no-numpad keyboard I can zoom out but I can't zoom in (cause I need to press SHITF+KEY for the "+").
> so I take this opportunity to, yet again, recommend that we should replace unity's input system for the new version :P
> EDIT: I've just added an issue to the board about this :)

@pbosio After ask Alvaro about what keys to use as ZoomIn/Out shortcuts, we're going to use only the numpad ones (commited here: https://github.com/decentraland/unity-renderer/pull/457/commits/9facd1e32f6a37971ab0da6f126023491786dd86). Obviously this will limit to all these keyboards that don't have the numpad side, but take on account that this shortcuts are only extra aids for the user, I mean he will always be able to zoom in/out with the mouse wheel or with the AWSD keys in the keyboard.

Apart of that, yeah, you're right that we should replace the Unity's input system by the new one. Thanks for add it to the backlog!

In order to not to block this specific PR, and confirming that with Álvaro, what do you think if we approve it and separate the input system refactor from here?

pbosio (approved):

Issue #299, opened 2021-04-28
Add generic Metrics flow to send data through segment


olavra:
In order to add any metrics for the explorer, to measure the users activity, we want to create a generic channel that allows us to send any data flag with any data structure to segment.

This will allow us to create panels in metabase or graphana, and have insights about the performance and adoption of features.

Please agree this feature with Pablo Seibelt

ie: For the Quests project, we want to start measuring when a player progresses on a quest, so relevant data will be:

flag: user_quest_progress

data:
- user_id
- quest_id
- task_id
- progress amount
- taks_status
- timestamp

kuruk-mm:
I suggest to add a `custom metric`.

We can call it in Unity with:

`DCL.Interface.WebInterface.CustomMetric(metricName, jsonData);`

And then to send to Segment we call it like:
`track("custom {metricName}", jsonData)`

So for example you can use:

```
[Serializable]
public class QuestProgressPayload
{
	public int user_id = 0;
	public int quest_id = 0;
	public int task_id = 0;
	public int progress_amount = 0;
	public int taks_status = 0;
	public long timestamp = 0;
}

[...]

QuestProgressPayload payload = new QuestProgressPayload
{
	user_id = 1,
	quest_id = 2,
	task_id = 5,
	progress_amount = 50,
	taks_status = 1,
	timestamp = 123123123
};
DCL.Interface.WebInterface.CustomMetric("user_quest_progress", payload);
```

and we will receive something like this in segment:

```
eventName -> "custom user_quest_progress"

payload:
{
	user_id: 1
	quest_id: 2
	task_id: 5
	progress_amount: 50
	taks_status: 1
	timestamp: 123123123
	sessionId: "27ad6e01-9150-40e8-a6ef-aac81cbbd46d"
	time: "2021-05-06T02:20:54.384Z"
	version: "70e47eeade1720a903b33f60cf48e9f912279016"
}
```

Let me know if we want something like this :)

sicarul:
I think the idea is great, but i would remove the "Custom" name from it, it'll make all metrics read as "custom ..." instead of just the metric. In your example, i'd make it so the name reads:

eventName -> "user_quest_progress"

The rest seems good. The event was just an example, if not, what's that user id? (We already get automatically the wallet if they have it in the current tracking on kernel), also, those quest ids and task ids, will we be able to get them somewhere to give more context rather than numbers? (Also please change taks_status to tasks_status)

Thanks!

PR #458, opened 2021-05-18
feat: analytics service provide

AjimenezDCL:
#299 

Goes in pair with this `explorer` PR: https://github.com/decentraland/explorer/pull/2369

It cannot be tested since no implementation is done yet. A smoke test should be enough


unity-renderer/Assets/Scripts/MainScripts/DCL/WebInterface/Interface.cs
```diff
@@ -1186,5 +1194,7 @@ public static void RequestUserProfile(string userId)
         }
 
         public static void ReportAvatarFatalError() { SendMessage("ReportAvatarFatalError"); }
+
+        public static void SendGenericAnalytic(string eventName, Dictionary<object, object> data) { SendMessage("GenericAnalytic", new GenericAnalyticPayload { eventName = eventName, data = data }); }
```
pbosio, line 1198: There is already a `ReportAnalyticsEvent` method in this class.
Is replacing/removing that planned for a following PR?

pbosio (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/ServiceProviders/Analytics/IAnalytics.cs
```diff
@@ -0,0 +1,21 @@
...
+
+public class Analytics : IAnalytics
+{
+    //Remove this once environment is on its own assembly and can be accessed properly
+    public static IAnalytics i;
+    public Analytics() { i = this; }
+
+    public void SendAnalytic(string eventName, Dictionary<object, object> data) { SendToSegment(eventName, data); }
+
+    internal void SendToSegment(string eventName, Dictionary<object, object> data) { WebInterface.SendGenericAnalytic(eventName, data); }
```
pbosio, line 18: Could we add a `VERBOSE` flag to be able to look at console's log what analytics are being send?

unity-renderer/Assets/Scripts/MainScripts/DCL/WebInterface/Interface.cs
AjimenezDCL: I didnt know about that method. Gonna dig a bit further, maybe we can reuse this and the `kernel` PR is not even needed.

AjimenezDCL: You were right @pbosio , the existent endpoint works as well here. I'm gonna use it!

unity-renderer/Assets/Scripts/MainScripts/DCL/ServiceProviders/Analytics/IAnalytics.cs
AjimenezDCL: That's a great idea, thanks!

pravusjif (approved):

PR #483, opened 2021-05-23
fix: BiW - Several fixes on thumbnails management

### LINKED ISSUE #466: BiW Editor: Draged entity preview should not show the background or favorite icon
When we drag an entity from the catalog panel to the scene, we should see the dragged icon only with the entity sprite, removing the box background and any other icon that is not the own entity.

### LINKED ISSUE #469: BiW Editor: Quickbar thumbnails are sometimes missing
When we drag an entity thumbnail from a quickbar slot to other one and we change the tab in the catalog panel, the moved thumbnail is missing its sprite.

sandrade-dcl:
### Fixes #469 (**Quickbar thumbnails are sometimes missing**)
- [x] When we drag an entity thumbnail from a quickbar slot to other one and we change the tab in the catalog panel, the moved thumbnail is missing its sprite.

### Fixes #466 (**Draged entity preview should not show the background or favorite icon**)
- [x] When we drag an entity from the catalog panel to the scene, we should see the dragged icon only with the entity sprite, removing the box background and any other icon that is not the own entity.

## How to test it?

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/BiW-fixes-on-catalog-item-thumbnails&ENV=zone&ENABLE_BUILDER_IN_WORLD&position=80,154

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Catalog/UI/Adapters/CatalogItemAdapter.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Catalog/UI/Adapters/CatalogItemAdapter.cs
@@ -108,8 +108,13 @@ public class CatalogItemAdapter : MonoBehaviour, IBeginDragHandler, IEndDragHand
         RectTransform newAdapterRT = GetComponent<RectTransform>();
         canvasGroup.blocksRaycasts = false;
         canvasGroup.alpha = 0.6f;
-        favImg.enabled = false;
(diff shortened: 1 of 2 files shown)


AjimenezDCL (approved):
I'm approving but I think this system needs a refactor.

It seems as if an entity `QuickBarSlot` would be missing. Therefore the `QuickBarController` would only fill these slots with a certain model and the `QuickBarSlot` itself would take care of the rest (thumbnails included)

pravusjif (approved):

PR #507, opened 2021-05-27
feat: [BiW Editor] Create a new Feedback Tooltip View

### LINKED ISSUE #474: BiW editor disabled Publish button should give feedback
The Publish button may be disabled in a few circunstances:
- If the scene is saving in the background
- If the scene limits are reached
- If there are elements outside of the scene boundaries or contain errors.

These messages are not handled and the Creator does not received feedback, leading to think its broken.

A tooltip with a message should appear displaying this information:

- [x] Scene limits reached
- [x] There are entities outside of the scene boundaries.

sandrade-dcl:
### Fixes #474 (**BiW editor disabled Publish button should give feedback**)
A tooltip with a message should appear displaying this information:
- [x] Scene limits reached

image

- [x] There are entities outside of the scene boundaries.

image

## How to test it?

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:feat/BiW-disabled-Publish-button-should-give-feedback&ENV=zone&ENABLE_BUILDER_IN_WORLD&position=80,154

## Our Code Review Standards


pbosio (approved):
LGTM!
Just, as a suggestion:
I've placed an entity outside the scene and I can see the tooltip, but the tooltip doesn't hide until I put the entity inside the scene again, I was expecting to automatically hide after a couple of seconds. So I have the tooltip there stealing me screen space (also is drawn over the exit tooltip)
It would be cool if the tooltip hides automatically after a couple of seconds and appear again when mouse is over the "publish" button

sandrade-dcl:
> LGTM!
> Just, as a suggestion:
> I've placed an entity outside the scene and I can see the tooltip, but the tooltip doesn't hide until I put the entity inside the scene again, I was expecting to automatically hide after a couple of seconds. So I have the tooltip there stealing me screen space (also is drawn over the exit tooltip)
> It would be cool if the tooltip hides automatically after a couple of seconds and appear again when mouse is over the "publish" button

Good point! I'm going to confirm the behaviour with Álvaro and decide about it. Thanks Pato!

PR #508, opened 2021-05-27
fix: loading when wearables fail

AjimenezDCL:
If any wearable in the player's profile has no valid representation for the current bodyshape, the avatar loading process fails and therefore the rendering startup.

This PR addresses the issue by using a default wearable in that case.

A better approach would be to always fallback to default wearables when the last retry has been done. The development time required to implement that solution is higher than this one and therefore it will be done afterwards.


sandrade-dcl (approved):
Awesome work, thanks!

pbosio (approved):

pravusjif (approved):

PR #516, opened 2021-05-28
fix: [BiW Editor] Error while reseting scale values from details panel

### LINKED ISSUE #380: BiW editor error while reseting scale values to 1:1:1
If you modify the scale of an entity in different axes and then you reset it to 1:1:1 in the details panel, the scale is corrupted.

sandrade-dcl:
### Fixes #380 
- [x] If you modify the scale of an entity in different axes and then you reset it to 1:1:1 in the details panel, the scale is corrupted.

image

## How to test it?

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/BiW-Editor-Error-while-reseting-scale-values&ENV=zone&ENABLE_BUILDER_IN_WORLD&position=80,154

## Our Code Review Standards


pravusjif (approved):
Also tested and it works fine, good job!
I believe this can also be fixed using `entityToUpdate.transform.lossyScale` but your approach is more reliable.

AjimenezDCL (approved):

PR #518, opened 2021-05-28
refactor: NFT fetch from OpenSea

pbosio:


sandrade-dcl (approved):
I have not been able to find any bug or any weird thing in your refactor, so confratulations! :) I think it's very clear.

pravusjif (approved):

PR #531, opened 2021-05-31
fix: [BiW Editor] Scene Limits calculation is wrong on number of Materials

### LINKED ISSUE #527: BiW: Scene Limits calculation is wrong on number of Materials
Elements like Meshes and Materials, are shared between different Entities and duplicated copies, and are considered as only 1 independently of the number of entities referencing it.

Scene limits in BIW is currently broken while counting duplicated materials.

**Steps to reproduce:**
1. Start editing a scene in BiW
2. Instantiate multiple assets and materials.
3. Make sure that you duplicate the same entity several times, observing the Material count does not increase for each duplicated entity
4. Exit the scene
5. Re-enter the editor and check the Materials counter. Counting is also considering materials of duplicated entities

sandrade-dcl:
### Fixes #527 
- [x] Scene limits in BIW is currently broken while counting duplicated materials.

image

## How to test it?

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/BiW-Editor-Scene-Limits-calculation-is-wrong-on-number-of-Materials&ENV=zone&ENABLE_BUILDER_IN_WORLD&position=80,154
3. Instantiate multiple assets and materials.
4. Make sure that you duplicate the same entity several times, observing the Material count does not increase for each duplicated entity.
5. Exit the scene.
6. Re-enter the editor and check the Materials counter.

## Our Code Review Standards


pbosio (commented):
Why aren't we using `SceneMetricsController` class instead of redoing scene metrics count for builder in world?

sandrade-dcl:
> Why aren't we using `SceneMetricsController` class instead of redoing scene metrics count for builder in world?

@pbosio we continue using `SceneMetricsController` -> `CalculateMaterial()`: https://github.com/decentraland/unity-renderer/blob/68ba7d6783bb66fad08c7d4313d0426fc319957e/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneMetricsController.cs#L303

But from there, it goes to `SceneBoundsChecker` -> `GetOriginalMaterials()`: https://github.com/decentraland/unity-renderer/blob/68ba7d6783bb66fad08c7d4313d0426fc319957e/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneBoundariesController/SceneBoundsChecker.cs#L36

And this `feedbackStyle` variable (`ISceneBoundsFeedbackStyle`) is specific depending on the system we are running (`SceneBoundsFeedbackStyle_Simple`, `SceneBoundsFeedbackStyle_RedFlicker` or `SceneBoundsFeedbackStyle_BIW`) and each one has its own `GetOriginalMaterials()` function. In the case of BiW: https://github.com/decentraland/unity-renderer/blob/68ba7d6783bb66fad08c7d4313d0426fc319957e/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneBoundariesController/SceneBoundsFeedbackStyle_BIW.cs#L54-L64

pbosio (approved):
👍

pravusjif (approved):

PR #538, opened 2021-06-01
feat: bots tool

pravusjif:
Bots Tool

How to use:
1. enter https://play.decentraland.zone/branch/feat/BotsTool/index.html?ENV=org&LOS=0&position=-110%2C-110&renderer=urn%3Adecentraland%3Aoff-chain%3Arenderer-artifacts%3Afeat%2FBotsTool
2. open browser console
3. Call a Bots Tool method:
- A. `clientDebug.InstantiateBotsAtCoords(amount: number, xCoord: number, yCoord: number, areaWidth: number, areaDepth: number)`: Instantiate bots in an area positioned at the base position of the parcel for which coords where provided, for example: `clientDebug.InstantiateBotsAtCoords(10)` will instantiate 10 bots in the user's current parcel with a default spawning area of 16x16 as default values apply when some parameters are not provided.
- B. `clientDebug.RemoveBot(botId: string)`: removes an instantiated bot. Every bot has its ID as its avatar name.
- C. `clientDebug.ClearBots()`: removes all the instantiated bots.

Note: The first batch takes a bit more because it fetches the catalogue of wearables.

PDR: https://docs.google.com/document/d/1Acc5yWt_Q_2Zc-jNkGm_W7mBRILnXBtKm6_GjXlwHLs/edit#
ISSUE: https://app.zenhub.com/workspaces/unity-6047bad476c3c0001942ee91/issues/decentraland/unity-renderer/298
Kernel-side: https://github.com/decentraland/explorer/pull/2403


sandrade-dcl (approved):
Reviewed and tested. Very useful tool! Thanks!

BrianAmadori (commented):
I had a few issues with usability, what `xCoord: number, yCoord: number`  is for? Tried the command a bunch of times with those values set to random values and nothing happened. I suggest to add some verbosity to the command, so it outputs something like "instancing X avatars in X,Y pos".

Also, the `bots` naming makes sense? We couldn't just name the command `InstantiateAvatarsAtPos` (for example). As they aren't really bots, they are just avatar components.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BotsController/BotsController.cs
```diff
@@ -0,0 +1,291 @@
...
+        private IEnumerator EnsureGlobalSceneAndCatalog()
+        {
+            if (globalScene != null)
+                yield break;
+
+            globalScene = Environment.i.world.state.loadedScenes[Environment.i.world.state.globalSceneIds[0]] as ParcelScene;
+
+            CatalogController.wearableCatalog.Clear();
+
+            yield return GetAllWearableItems(ALL_WEARABLES_FETCH_FIRST_URL);
```
AjimenezDCL, line 49: Do we really need the whole catalog? This won't scale very well with the new wearables editor.

AjimenezDCL (approved):
Approved with a concern

pravusjif:
@BrianAmadori 
> I had a few issues with usability, what `xCoord: number, yCoord: number` is for? Tried the command a bunch of times with those values set to random values and nothing happened. I suggest to add some verbosity to the command, so it outputs something like "instancing X avatars in X,Y pos".

Cool, I just added a bunch of log calls that report the status of the instantiation with those values.

> Also, the `bots` naming makes sense? We couldn't just name the command `InstantiateAvatarsAtPos` (for example). As they aren't really bots, they are just avatar components.

In the future we could add some behaviours to these plain-avatars that we use now, like some movement/waypoints pattern, emojis, etc. I prefer to keep the name as Bots so that we can build on this and make them more bot-like and less plain-avatar-like, what do you think?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BotsController/BotsController.cs
pravusjif: True, I have to refactor and change how we fetch the collections for the catalogue because o an API change so I'll also change it to randomize the collections it'll download and use the randomized collections locally to populate the catalogue (instead of populating it with ALL the collections)

BrianAmadori:
>In the future we could add some behaviours to these plain-avatars that we use now, like some movement/waypoints pattern, emojis, etc. I prefer to keep the name as Bots so that we can build on this and make them more bot-like and less plain-avatar-like, what do you think?

👍

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BotsController/BotsController.cs
pravusjif: Done!

BrianAmadori (approved):
LGTM! Looks great!

PR #543, opened 2021-06-02
fix: mouse lock enabled on launch screen

BrianAmadori:
## What this PR includes?

When any user logs for first time, the mouse will be locked on decentraland splash screen. This is caused by the `SetVisibility` method on `QuestsPanelHUDController`. Couldn't put the renderer visibility check on the `LockCursor()` method itself because cyclic references (the mouse locking code is awful and should be in its own assembly).

## How to test it?
To reproduce the bug, go into https://play.decentraland.zone in incognito mode. You will notice that the mouse will lock by itself at landing page as soon as Unity finishes loading. The bug can also be reproduced going to https://play.decentraland.org/?QUESTS_ENABLED 

## Our Code Review Standards


pravusjif (approved):

pbosio (approved):
Maybe in the future we could add the renderer state check inside `Utils.LockCursor` to avoid something like this happening again?

Issue #511, opened 2021-05-27
Integrate Polygon Mana balance in user Profile Card


olavra:
Player need to have visibility of the Mana that they have in their wallet both for Ethereum and Polygon networks when clicking on the Profile Picture inside the explorer.

![image.png](https://images.zenhubusercontent.com/603cdb58edb53386a3d0d8d5/e90c23c0-23c8-4108-ba68-f5d4e02827ce)

PR #554, opened 2021-06-08
feat: polygon mana balance in the profile

AjimenezDCL:
#511 
Polygon MANA balance is now shown in the profile.

I'm not really happy with the implementation but the `ProfileHUD` needs quite the refactor. I tried to adjust to the current "paradigm" and created a ticket to solve it in the future: https://app.zenhub.com/workspaces/unity-6047bad476c3c0001942ee91/issues/decentraland/unity-renderer/549


unity-renderer/Assets/Scripts/MainScripts/DCL/ServiceProviders/TheGraph/TheGraph.cs
```diff
@@ -90,6 +92,32 @@ public Promise<List<Land>> QueryLands(string tld, string address, float cacheMax
         return promise;
     }
 
+    public Promise<double> QueryMana(string address)
```
pbosio, line 95: Could we add something to explicitly point out that this is for matic/polygon chain?
Not sure what would be the best way to go, changing the method name or adding a new argument to the method like `bool usePolygonChain` and use one url or the other depending on that.

pbosio (approved):
LGTM

unity-renderer/Assets/Scripts/MainScripts/DCL/ServiceProviders/TheGraph/TheGraph.cs
AjimenezDCL: I'm gonna rename it to `QueryPolygonMana` and once we port the retrieval of `MainNet/Ropsten` to Unity we can create another query (similar to how Lands are requested)

pravusjif (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/ServiceProviders/TheGraph/TheGraph.cs
```diff
@@ -90,6 +92,32 @@ public Promise<List<Land>> QueryLands(string tld, string address, float cacheMax
...
+            .Then(resultJson =>
+            {
+                try
+                {
+                    JObject result = JObject.Parse(resultJson);
+                    JToken manaObject = result["data"]?["accounts"].First?["mana"];
+                    if (manaObject == null || !double.TryParse(manaObject.Value<string>(), out double parsedMana))
+                        throw new Exception($"QueryMana response couldn't be parsed: {resultJson}");
+
+                    promise.Resolve(parsedMana / 1e18);
```
BrianAmadori, line 110: Why are we dividing the value by `1e18`?

BrianAmadori (approved):

AjimenezDCL:
> I'm approving this, BUT, there's no way of knowing what the icons stand for, and there's no good explanation in the tooltip:
> 
> image
> 
> I think we should hold this PR until we get a better tooltip or at least some way of improving the usability of this.

@BrianAmadori You are completely right about this, We are in the process of finding a more appropriated tooltip and hover text

kuruk-mm (approved):
Looks great

PR #566, opened 2021-06-09
feat: builder in world unpublish scene

pbosio:
## What this PR includes?
Add "Unpublish" option to scene context menu in Builder Projects Panel

## How to test it?
1- go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:feat/builderinworld-unpublish-scene&ENV=zone&ENABLE_BUILDER_IN_WORLD
2- create a scene using builder in world, if you don't have any, and publish it
3- open builder projects panel from the taskbar
4- find your scene on the list, open it context menu by pressing the three dots on the top-right of the card
5- press "Unpublish"
6- follow the "Unpublish" flow
7- profit?

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Bridges/DataStoreBridge/DataStoreBridge.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+using DCL;
+using UnityEngine;
+
+public class DataStoreBridge : MonoBehaviour
```
AjimenezDCL, line 5: Not sure about this name. Shouldn't this be something related to `BuilderInWorld`  rather than the broad concept of `DataStore`?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Scripts/Unpublish/UnpublishPopupController.cs
```diff
@@ -0,0 +1,80 @@
...
+        view.Hide();
+        CoroutineStarter.Stop(fakeProgressRoutine);
+    }
+
+    IEnumerator ProgressRoutine()
+    {
+        float progress = 0;
+        while (true)
+        {
+            progress = Mathf.Clamp(progress + UnityEngine.Random.Range(0.01f, 0.03f), progress, 0.99f);
```
AjimenezDCL, line 60:
```suggestion
            progress = Mathf.Clamp(progress + UnityEngine.Random.Range(0.01f, 0.03f), 0, 0.99f);
```
Shouldnt it be like this?

AjimenezDCL (approved):
Approving with a few minor comments

sandrade-dcl:
@pbosio it is working fine! Good job.

Testing your PR I have noticed that I see different information in the panel related to my created scenes:
- In your PR I only see my **already published** scenes:

image

- In master I see both, the **already published** scenes and **other projects** that I had created without still publish:

image

Is it the expected behaviour for this panel? or should we see both kind of scenes (published and not published)? @olavra

sandrade-dcl (changes requested):
Code reviewed, good work!

Testing the build I have noticed that there is something that is not working correctly: when we unpublish a scene, after the unpublish process is end, the land in the world should be updated (in this case it should disappear) and not having to restart the browser to see the applied changes. For example, for the case of "publish" a scene, when the publication process ends, the land is automatically updated and you see the new scene at the moment.

olavra:
> @pbosio it is working fine! Good job.
> 
> Testing your PR I have noticed that I see different information in the panel related to my created scenes:
> 
> * In your PR I only see my **already published** scenes:
> 
> image
> 
> * In master I see both, the **already published** scenes and **other projects** that I had created without still publish:
> 
> image
> 
> Is it the expected behaviour for this panel? or should we see both kind of scenes (published and not published)? @olavra

Unpublished Scenes must disappear from that panel,

sandrade-dcl:
> Unpublished Scenes must disappear from that panel,

Ok! Then @pbosio forget what I said about it :)

unity-renderer/Assets/Scripts/MainScripts/DCL/Bridges/DataStoreBridge/DataStoreBridge.cs
pbosio: Truth be told. I didn't wanted to add a MonoBehaviour just for this.
I was thinking... How many times did I had to create a MonoBeaviour just to set a variable from the DataStore with data coming from kernel...
So... Why not creating a script where we can do just that?
That's the main reason for this. What do you think? Is it a good idea or should I change it?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuilderProjectsPanel/Scripts/Unpublish/UnpublishPopupController.cs
pbosio: It's kind of the same thing case `progress` should never decrease... but I'll change that, I agree is more confusing this way 🤣

pbosio:
> Code reviewed, good work!
> 
> Testing the build I have noticed that there is something that is not working correctly: when we unpublish a scene, after the unpublish process is end, the land in the world should be updated (in this case it should disappear) and not having to restart the browser to see the applied changes. For example, for the case of "publish" a scene, when the publication process ends, the land is automatically updated and you see the new scene at the moment.

I will tackle this on another PR since is a kernel issue and this PR is just about UI flow

sandrade-dcl:
> I will tackle this on another PR since is a kernel issue and this PR is just about UI flow

Ok, then I'm going to approve this PR 👍

sandrade-dcl (approved):

PR #577, opened 2021-06-10
fix: third person camera polish

BrianAmadori:
## Test Link
https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/camera-polish&ENV=org

## What this PR includes?
This PR includes a general polish of our third person camera. Most of the changes proposed in the RFC were implemented, and more. (https://docs.google.com/document/d/1iRUr7-wiZHA3S8gMvUfLqNiPqO5YXrSRePhb0PvAErA/edit#heading=h.g1uhsmsys485)

* Looking above the avatar doesn't place the cursor behind the avatar.
* Changed general framing for bottom, mid, and top looking angles.
* When walking forward and backwards, the camera is not tilting anymore.
* When walking backwards, the avatar doesn't get too close of the camera anymore.
* Walking over bumpy ground--like genesis plaza fields with small rocks and stuff--doesn't make the camera altitude jitter. This also works for preventing jittering on stairs that lack a ramp collider. The damping that fixes this is deactivated when the avatar is airborne and standing on moving platforms, so when jumping on moving platforms the feel should keep working as before.
* Camera distance and fov is incremented gradually when running forward. When standing, the distance/fov springs back slowly.
* Collider detection wasn't working correctly. Now all colliders block the camera correctly.
* Added special behaviour for the genesis plaza hole. We detect falling from high places with a time threshold (this means the avatar has to be falling for at least 2 seconds). When that happens the camera blends into a special framing.
* Added a very minuscule damping to rotation input.

Technical Stuff:
* Cinemachine package was updated to the latest version

## Our Code Review Standards


AjimenezDCL:
I feel like the camera tilts a little bit when aiming from below, is that an artistic decision or a bug? IMHO feels a bit weird.

image

Suduck:
I have the same opinion as Alex, also isn't the top limit angle too much? As you can see in the screenshot this is the further point in the top that I can reach in 3rd person, you can select further events whicht feel strange
image

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
```diff
@@ -375,6 +376,7 @@ void Jump()
         ResetGround();
 
         velocity.y = jumpForce;
+        //cameraTargetProbe.damping.y = dampingOnAir;
```
Suduck, line 379: Please remove commented code!

Suduck (approved):
Just a very small request!

Overall I think that our current camera implementation is a little over-complicated ( I could be missing a lot, so take this with a grain of salt). In a future we should look to overhaul the camera completly

sandrade-dcl (approved):
Overall I feel the camera behaviour has improved a lot, good job 👍

Anyway, I agree with Adri that our camera system seems to be too complex for what it currently does. Too much code logic for maintain something that, in theory, tools as cinemachine should provide us as something configurable.

BrianAmadori:
Just a heads-up: The current additional logic had to be made because it does things that cinemachine CANNOT do with just configuration. Yes, cinemachine has limitations. The complexity is justified (at least for now). cc: @sandrade-dcl @Suduck

Issue #487, opened 2021-05-24
Launch Explorer Desktop using Guest Mode by default


olavra:
While we implement a launch screen with selection between Wallet Connect or Guest, we will automatically launch the desktop Explorer in Guest Mode.

- [x] Mockup Sign-up Screen with JUST "Play as Guest" button to trigger the workflow
- [x] First launch will trigger the Sign-up Screen
- [x] Next time player info will be retrieved from local store
- [ ] ~~Login out will clear local store and redirect to Sign-up screen~~

Log out tips
-> Refresh headless chromium
-> Reload Desktop Scene

Additional information in Notion
https://www.notion.so/decentraland/Desktop-Client-V1-94bc953821724c6e921385f661afc323#1b05138ef50443ba8895c708aeafd401

PR #611, opened 2021-06-15
feat: message to authenticate from the renderer as a guest or using w…

kuruk-mm:
…allet connect

Related to #487

## What this PR includes?
Add a way to authenticate from the renderer as a guest or using wallet_connect

## Why
Desktop client needs to decide itself when will authenticate

## Our Code Review Standards


pravusjif (approved):

sandrade-dcl (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/WebInterface/Interface.cs
```diff
@@ -954,6 +954,18 @@ public class SaveAvatarPayload
             public AvatarModel avatar;
         }
 
+        public enum RendererAuthenticationType
+        {
+            GUEST = 0,
+            WALLET_CONNECT = 1
+        }
```
pbosio, line 961: I would suggest not using an enum here, instead create a string constants and send a string as payload for the message.
That way we don't need to maintain two enums for this since kernel only needs a string for the provider. In this case "wallet_connect"

unity-renderer/Assets/Scripts/MainScripts/DCL/WebInterface/Interface.cs
```diff
@@ -953,7 +953,19 @@ public class SaveAvatarPayload
             public bool isSignUpFlow;
             public AvatarModel avatar;
         }
-
+        
+        public static class RendererAuthenticationType
+        {
+            public static string Guest => "guest";
+            public static string WalletConnect => "WalletConnect";
```
pbosio, line 960: this string is not matching what is expected on kernel side, which is `'wallet_connect'`
other than that the looks great

kuruk-mm, line 960:
```suggestion
            public static string WalletConnect => "wallet_connect";
```

kuruk-mm: ups :fearful:

pbosio (approved):

PR #615, opened 2021-06-15
chore: remove fast moving platforms safeguard

pravusjif:
### **WHY**
Since we already let the parcel devs to teleport the users, we might as well remove this safeguard that releases the character from a moving platform when it moves too fast, this way **we improve the experience for users on long-distance moving platforms when the framerate drops**.

**Why does the framerate drop in these long-distance moving platforms?**
Mostly it's because of all the scenes loading and unloading all around the character while being moved by the moving platform.

### **WHAT**
Removed fast moving platforms safeguard to improve experience on long distance moving platforms

### **TEST**
This can be easily tested at the new tram that NicoE is implementing, unless you are running on a high-end pc you'll get dropped off while riding on it and it completely breaks the experience:

Prod (wait for the wagon to arrive) 
https://play.decentraland.org/?CATALYST=peer-testing.decentraland.org&position=0%2C12&realm=testing-amber

This PR (wait for the wagon to arrive) 
https://play.decentraland.org/?CATALYST=peer-testing.decentraland.org&position=0%2C12&realm=testing-amber&renderer=urn%3Adecentraland%3Aoff-chain%3Arenderer-artifacts%3Achore/RemoveMovingPlatformsSafeguard


pbosio (approved):
Tested. Working. :)

sandrade-dcl:
I'm not sure what should be the improved behaviour, but testing with 2 avatars at the same time (one logged with my wallet and the other one as guest) I'm seeing that when both avatars go inside the wagon and it starts moving, each avatar is seeing the other one moving further and further away (playing the waling animation) until each avatar is off the wagon and keep there "flying". Is it the correct behaviour?

image
image

pravusjif:
> I'm not sure what should be the improved behaviour, but testing with 2 avatars at the same time (one logged with my wallet and the other one as guest) I'm seeing that when both avatars go inside the wagon and it starts moving, each avatar is seeing the other one moving further and further away (playing the waling animation) until each avatar is off the wagon and keep there "flying". Is it the correct behaviour?
> 
> image
> image

Santi this PR has nothing to do with 'other avatars', it's only local... in my macbook pro, every long distance platform that I use drops me off half-way instead of being able to complete the travel distance, this PR avoids that.

sandrade-dcl (approved):
Code reviewed! LGTM!

AjimenezDCL (approved):

PR #631, opened 2021-06-18
fix: set AvatarCollider to the correct layer to detect the collisions…

### LINKED ISSUE #557: AvatarModifiers.HIDE_AVATARS no longer works
This example scene works fine on 6.6.4, but doesn't on 6.6.5

https://github.com/decentraland-scenes/avatar-swap

The player's avatar should be removed when entering the swap area, this worked fine on 6.6.4,  but in 6.6.5 the original avatar never goes away

kuruk-mm:
… with BoxTriggerArea for AvatarModifierArea

## What this PR includes?
Fix #557

The AvatarCollider was in an incorrect layer, so it wasn't detecting an Avatar entering a Trigger Area. And the side-effect is #557.

## How to test it?
Go to one of the following position:
- https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/trigger-collider-layer&position=-24,22&ENV=org
- https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/trigger-collider-layer&position=-46,-99&ENV=org
- https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/trigger-collider-layer&position=150,63&ENV=org

And check if you're avatar hide properly.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/Resources/Prefabs/CharacterController.prefab
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/Resources/Prefabs/CharacterController.prefab
@@ -233,7 +233,7 @@ GameObject:
   m_Component:
   - component: {fileID: 4710899382381041556}
   - component: {fileID: 7115442889188564318}
-  m_Layer: 16
+  m_Layer: 21
   m_Name: AvatarCollider
   m_TagString: Untagged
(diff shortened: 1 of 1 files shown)


AjimenezDCL (approved):

nearnshaw (approved):
I have checked the broken scenes and looks like they all work now with this change :D

PR #634, opened 2021-06-18
fix: tutorial robot is culled

AjimenezDCL:
The culling controller was culling the tutorial robot due to a race condition.
We fix it by having the robot disabled until needed (and the CullingController's ignore mask is ready as well).

#588

### DIFF
--- a/unity-renderer/Assets/Rendering/Culling/Interfaces/CullingControllerSettings.cs
+++ b/unity-renderer/Assets/Rendering/Culling/Interfaces/CullingControllerSettings.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using UnityEngine;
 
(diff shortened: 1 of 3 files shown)


sandrade-dcl (approved):

menduz:
May I suggest commenting in the code why are those lines there? It is a love letter to your future selves

AjimenezDCL:
@menduz We changed the approach to a more broad solution. We also decided to do a small research on how can we optimize the use of layers. https://app.zenhub.com/workspaces/unity-6047bad476c3c0001942ee91/issues/decentraland/unity-renderer/639

BrianAmadori (approved):

sandrade-dcl (approved):

Issue #641, opened 2021-06-18
BIW: If you are using the keys to move the camera and the newsletter popup appear, the keys will stuck in active


Suduck:
If you are using the editor while moving right the camera ( for example) and the newsletter pop up appear, the camera will continue moving to the right because the keys will be stuck activated

olavra:
remove the newsletter popup

PR #651, opened 2021-06-21
fix: remove email request prompt

pbosio:
#641 
remove email request prompt hud since we are already asking for user's email during sign-up


Suduck (approved):
Is this PR the heaven?

AjimenezDCL (approved):

sandrade-dcl (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/HUDController.cs
```diff
@@ -159,16 +157,15 @@ public enum HUDElementID
...
-        COUNT = 30
+        USERS_AROUND_LIST_HUD = 21,
+        GRAPHIC_CARD_WARNING = 22,
+        BUILDER_IN_WORLD_MAIN = 23,
+        BUILDER_IN_WORLD_INITIAL = 24,
+        QUESTS_PANEL = 25,
+        QUESTS_TRACKER = 26,
+        BUILDER_PROJECTS_PANEL = 27,
+        SIGNUP = 28,
+        COUNT = 29
```
menduz, line 168: Please do not break compatibility at any moment with kernel, enums in protocols are constant and forever. You could leave the old unused enum element, it could be documented to avoid usage. 

Removing and reassigning values would create an unnecessary friction in the merge of both pull request simultaneously. And that is something we can avoid.

menduz (changes requested):

menduz (approved):
We can now merge this to master without breaking .zone, thanks

PR #664, opened 2021-06-23
refactor: scene boundarie checker

### LINKED ISSUE #691: SceneBoundaries Feedback change refactor

Suduck:
## What this PR includes?

Fixes #691 

This PR has made a little refactor to the scene boundaries checker to include a clean feedback function to improve consistency while changing the feedback k style

## How to test it?

1.  Create a new feedback style and change it at runtime while there is another feedback applying 

## Our Code Review Standards


pravusjif (changes requested):
The code looks good but I just tested with the local test scene at 0, 90 (`0.90.scene-boundaries`), because this is a behaviour that is also used in preview mode, and I saw that using this PRs renderer we lost the wireframe box that used to identify which submesh is out of bounds, inside an affected mesh:

Master:
https://user-images.githubusercontent.com/1031741/123330939-059ae980-d515-11eb-80a4-1f3ac94d814d.mov

This PR:
https://user-images.githubusercontent.com/1031741/123330991-13506f00-d515-11eb-916b-c6cbb75c7072.mov

We need to keep that functionality because with giant meshes in a scene it is difficult to identify which submesh is the culprit for being out of bounds, and that wireframe mesh is for that.

AjimenezDCL (approved):
Have not tested it but code looks good.

pravusjif (approved):
I tested the last build with the latest changed and now the Scene Boundaries Checker seems to be correct, good job!

PR #670, opened 2021-06-23
feat: biw analytics

Suduck:
## What this PR includes?

This PR includes analytics for builder in world

## How to test it?

1. Go to: https://play.decentraland.zone/?index.html&renderer=urn:decentraland:off-chain:renderer-artifacts:{branch_name}&{desired_url_params}
2. ...

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Analytics/BIWAnalytics.cs
```diff
@@ -0,0 +1,201 @@
...
+    public static void NewObjectPlaced(CatalogItem catalogItem, string source)
+    {
+        Dictionary<string, string> events = new Dictionary<string, string>();
+        events.Add("name", catalogItem.name);
+        events.Add("assetPack", catalogItem.assetPackName);
+        events.Add("category", catalogItem.category);
+        events.Add("category Name", catalogItem.categoryName);
+        events.Add("source", source);
+        events.Add("type", catalogItem.itemType.ToString());
+        AddLandInfoSendEvent("new_object_placed", events);
```
sicarul, line 127: Could this be a little bit heavy if the user is putting e.g. a wall around? Maybe we should only send this if it's different to the last object placed? So we don't get 100x the same object because they are planting flowers or building walls in a scene?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Analytics/BIWAnalytics.cs
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    public static void QuickAccessAssigned(CatalogItem catalogItem, string source)
```
sicarul, line 130: What is this? (It's missing comments and i can't deduce it)

sicarul (commented):
Please check my two comments about some of the events @Suduck thanks!

pravusjif (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Analytics/BIWAnalytics.cs
Suduck: In builder in world we have a quick access buttons  ( from 1 to 9) where you can set assets to quickly drop them in the scene, each time a asset is assigned to this part, an event is fired with the item assigned

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Analytics/BIWAnalytics.cs
Suduck: If the user is putting severals wall this will be fired each time.  I think that most of the user won't set more than 100 assets per scene, but this is only a "non-exact prediction".

 If we only send 1 event for each asset we lose the ability to track the most used asset in total. Also, we are losing how many times a user is using an asset per scene but this may be overkill. 

Another option is to add a counter with the number of times that the user has set the asset and only send the event if X seconds has pass without set the same asset in the scene, but we will lose some information if the user disable biw or exit from the explorer

@olavra what do you think about that?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Analytics/BIWAnalytics.cs
sicarul: ty <3

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Analytics/BIWAnalytics.cs
sicarul: >  we are losing how many times a user is using an asset per scene 
We'll still have this data in the finished scene right? I plan on downloading all scenes anyway to the DWH, so you don't really need this data sent by Segment. I was just thinking if you can save in a variable the last object sent, and if last_asset = current_asset, then don't send, otherwise send. That way if someone does:

Place 100 walls
Place 50 flowers
Place 25 walls

We'd only get 3 events (place wall, place flower, place wall) - Makes sense? I don't think we need 175 events for this scenario above

Suduck: Yep but when do you send that event to be sure that we are only sending 3 events? Users can close the web any time without exiting from the editor and the others parts of biw will work.

We can send the event if nothing has changed in X seconds. Does that work for you? Which time will you propose?

AjimenezDCL: I like the idea of sending batches every few seconds.

Another alternative is reporting when you change your selected asset or close the BiW. There's the downside of losing track of the lasts placed items if the user were to close the tab, but I think it's a fair trade-off.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Analytics/BIWAnalytics.cs
```diff
@@ -0,0 +1,201 @@
...
+    public static void AddSceneInfo(Vector2Int sceneCoords,  string sceneOwnership, Vector2Int sceneSize)
+    {
+        coords = sceneCoords;
+        ownership = sceneOwnership;
+        size = sceneSize;
+    }
+
+    #endregion
+
+    private static void AddLandInfoSendEvent(string eventName, Dictionary<string, string> events)
```
AjimenezDCL, line 192: It might add a bit of boilerplate code, but I would suggest only adding data to the dictionary in this method.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Utils/BuilderInWorldUtils.cs
```diff
@@ -12,13 +12,29 @@
...
 {
+    public static LandRole GetLandOwnershipType(List<LandWithAccess> lands, ParcelScene scene)
+    {
+        LandWithAccess filteredLand = lands.Where( land => scene.sceneData.basePosition == land.baseCoords)
+                                           .Select(land => land)
+                                           .FirstOrDefault();
+        return GetLandOwnershipType(filteredLand, scene);
+    }
+
+    public static LandRole GetLandOwnershipType(LandWithAccess land, ParcelScene scene)
```
AjimenezDCL, line 31: `ParcelScene scene` is not needed here.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Utils/BuilderInWorldUtils.cs
```diff
@@ -12,13 +12,29 @@
...
 using DCL.Controllers;
 using UnityEngine.Networking;
 using UnityEngine.Events;
 
 public static partial class BuilderInWorldUtils
 {
+    public static LandRole GetLandOwnershipType(List<LandWithAccess> lands, ParcelScene scene)
+    {
+        LandWithAccess filteredLand = lands.Where( land => scene.sceneData.basePosition == land.baseCoords)
+                                           .Select(land => land)
```
AjimenezDCL, line 26: Isn't this `Select` useless? xD

AjimenezDCL (approved):
Left a couple of requests, but approving regardless

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Analytics/BIWAnalytics.cs
Suduck: I don't understand what you mean here, Each function has his own parameters and this function adds the generic one for every event

AjimenezDCL: `AddLandInfoSendEvent` -> `AddLandInfo`. So we dont have a method that does two things (adding info to the dictionary and sending the event).

Suduck: This is a common pattern in analytics. The reason behind this decission is to left the ability to add more layers without having to know much about the structure, in a deeper layer more generic information is added like timestamp, user id...etc. 

For example: In this case, I don't need to know about what are the data that we are collecting from the user, I just can wrap my data into the event and send to the next layer 

The name is probably confusing and I will rename it to `SendEditorEvent `  because we want to add this information to every editor event. 

Does this work for you?

AjimenezDCL: Sure, but then why dont we just use the `SendEvent` method to add this common data?

Suduck: When we are sending the data from the project panel, we still don't have a land chosen to edit. So we only add the information of the land when we have already chosen a land ( when we enter in editor mode).

We could have 3 methods:
-SendEditorEvent
-SendPanelEvent
-SendEvent

Since the panel don't have common data to share the SendPanelEvent is not added

AjimenezDCL: That seems clear enough

PR #677, opened 2021-06-23
feat: avatar lods (feature-flagged)

pravusjif:
### Avatar LODs
* Implemented first iteration of **Avatar LODs** based on a **LOD Distance** (default 16) and a **Max non-lod avatars** (default 20), both configurable in the settings panel
* Current implementation uses a generic **placeholder** LOD texture
* The feature is feature-flagged with `ENABLE_AVATAR_LODS`
* The top screens won't be showing any video streaming in ZONE realms because of CORS (in a local test scene those work fine)
* Disregard bugs on bots wearables, the endpoint we hit for the collections also brings non-published collections that may be bugged.
* Profiling results document: https://www.notion.so/decentraland/Bots-Profiling-for-Massive-Events-7471e8cb14c048b38152968d073cad40

### Testing
1. enter https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:test/AvatarQuadLODs&ENV=zone&position=91,151&ENABLE_AVATAR_LODS
2. open browser console

https://user-images.githubusercontent.com/1031741/124167165-1e0d8580-da7a-11eb-960b-923299ef4a54.mov


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarsLODController.cs
```diff
@@ -0,0 +1,147 @@
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace DCL
+{
+    public class AvatarsLODController : MonoBehaviour
+    {
+        public static AvatarsLODController i { get; private set; }
```
sandrade-dcl, line 8: Just out of curiosity: why are we using a singleton for this? Shouldn't we add this `AvatarsLODController` into our `Environment`?

sandrade-dcl (approved):
Reviewed and tested, goog job! I approved but I left some doubts.

sandrade-dcl:
Testing the build I noticed that the configured distance for swtiching between different LOD versions of the avatar maybe be a little small. Will this be the final configuration?

And other doubt: Will this dark shadow be the final visual for the distant avatars? or is it something temporal?

image

pravusjif:
> Testing the build I noticed that the configured distance for swtiching between different LOD versions of the avatar maybe be a little small. Will this be the final configuration?
> 
> And other doubt: Will this dark shadow be the final visual for the distant avatars? or is it something temporal?
> 
> image

In this first iteration there are no different LOD versions, there is only that **placeolder texture** that is not the final visual, it's a placeholder XD

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarsLODController.cs
pravusjif: I can try making that change

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarsLODController.cs
```diff
@@ -0,0 +1,147 @@
...
+                lodGO.transform.forward = lookAtDir;
+            }
+
+            if (applyVerticalMovement)
+                lastLODsVerticalMovementTime = Time.timeSinceLevelLoad;
+        }
+
+        void OnMainPlayerReposition(Vector3 newPos, Vector3 previousPos) { UpdateAllLODs(); }
+
+        public void UpdateAllLODs()
```
pbosio, line 78: Is it possible to make this method private and call it when `lodDistance` or `maxNonLODAvatars` are changed? (converting them into properties)

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarsLODController.cs
```diff
@@ -0,0 +1,147 @@
...
+                                CommonScriptableObjects.playerUnityPosition.OnChange += OnMainPlayerReposition;
+                            }
+                            else
+                            {
+                                enabled = false;
+                            }
+                        });
+        }
+
+        public void LateUpdate()
```
pbosio, line 49: this method can be made private

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarsLODController.cs
```diff
@@ -0,0 +1,147 @@
...
+
+                ToggleLOD(currentAvatar.avatarRenderer, isLOD);
+            }
+        }
+
+        private void ToggleLOD(AvatarRenderer avatarRenderer, bool enabled)
+        {
+            avatarRenderer.lodRenderer.gameObject.SetActive(enabled);
+            avatarRenderer.SetVisibility(!enabled); // TODO: Resolve coping with AvatarModifierArea regarding this toggling
+        }
```
pbosio, line 110: This is just my obsessive disorder. Can we move private methods to the bottom of the class? pretty please? :)

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarsLODController.cs
BrianAmadori: +1. As the `CullingController` is already implemented using this approach would make total sense to have this there. We should avoid the `MonoBehaviour` as well.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarsLODController.cs
```diff
@@ -0,0 +1,147 @@
...
+                return;
+
+            avatarsList.Add(newAvatar);
+
+            // TODO: Find a way to get this behaviour but unsubscribing on RemoveAvatar()
+            // DecentralandEntity already nullifies this event on cleanup, maybe that's good enough?
+            // Worst case scenario we'll have to refactor the DCLEntity.OnTransformChange event to report which entity triggered the event
+            newAvatar.entity.OnTransformChange += (object newTransformModel) =>
+            {
+                UpdateAllLODs();
```
BrianAmadori, line 124: This shouldn't be very unperformant? If this method iterated over all avatars, this operation approaches O(n*n). It isn't better to just mark dirty and have all the lods updated a single time per frame?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Settings/SettingsControllers/SpecificControllers/MaxNonLODAvatarsControlController.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace DCL.SettingsControls
+{
+    [CreateAssetMenu(menuName = "Settings/Controllers/Controls/Max Non-LOD Avatars", fileName = "MaxNonLODAvatarsControlController")]
+    public class MaxNonLODAvatarsControlController : SliderSettingsControlController
+    {
+        public override void Initialize()
+        {
+            base.Initialize();
+
+            KernelConfig.i.EnsureConfigInitialized()
```
BrianAmadori, line 12: @pbosio we should do something about this approach.

1) KernelConfig is a singleton, and we should get rid of them as possible
2) Using it like this forces us to deal with race conditions
3) KernelConfig dirty state will be carried over from test to test, and that will force us to clear kernel config state on `IntegrationTestSuite`. At long term, this will make `IntegrationTestSuite` roll back to the `IntegrationTestSuite_Legacy` intialization clusterfuck.

pbosio: we can always move it to Environment like everything else

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarsLODController.cs
Suduck: Agree!

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarsLODController.cs
Suduck, line 49: Why we need it in the `LateUpdate`? Can't we use it in the normal `Update`?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Settings/Settings.cs
```diff
@@ -113,6 +113,8 @@ private SettingsData.GeneralSettings GetDefaultGeneralSettings()
                 sfxVolume = 1,
                 mouseSensitivity = 0.6f,
                 scenesLoadRadius = 4,
+                avatarsLODDistance = 16,
+                maxNonLODAvatars = 20,
```
Suduck, line 117: This should be in the graphics settings? This is related to the graphics rather than general information. Also, we should think a more friendly name so the users can understand what is for something like "Avatar rendering max distance" most users won't understand what LOD means

Suduck (approved):
Good job! Just a small request to move the settings!

Suduck:
I just noted, that if you stand the lods will move up and down, 
avatarlod

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarsLODController.cs
pravusjif: I reordered the methods, but left the Awake and LateUpdate on top because they are Unity hooks. Althought if I end up removing the monobehaviour functionality that will change.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarsLODController.cs
pravusjif: True that, better to update the LODs on every frame, even if it triggers unneeded updates, instead of on every character+avatars movement. Done.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarsLODController.cs
pravusjif: Done

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarsLODController.cs
pravusjif: Done, now we use those config values from the DataStore and the setting control controllers set those values in there

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarsLODController.cs
pravusjif: It's for the "Billboard" behaviour, if we use Update() instead of LateUpdate() there were issues in the past with the BillboardComponent not updating AFTER the main character moved, but in this case I think it's not needed, so I changed it to a regular `Update()` as you suggest.
If we change the controller to stop being a monobehaviour I'll remove it though.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Settings/Settings.cs
pravusjif: Moved it to the graphics settings you are right.
We can change the naming in of the many PRs that will continue modifying this feature.

pravusjif:
> I just noted, that if you stand the lods will move up and down

This is on purpose to give a little "life" to the otherwise static LODs, if we don't like it we can remove that in next PRs

unity-renderer/Assets/Rendering/Culling/Interfaces/CullingControllerSettings.cs
```diff
@@ -18,7 +18,7 @@ public class CullingControllerSettings
 
         public float enableAnimationCullingDistance = 7.5f;
 
-        public int ignoredLayersMask = LayerMask.GetMask("Tutorial", "CharacterPreview");
+        public int ignoredLayersMask = LayerMask.GetMask("Tutorial", "CharacterPreview", "AvatarLOD");
```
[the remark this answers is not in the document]
pravusjif: The new layer is needed to ignore the LODs by the Culling Controller. Do you suggest I use `Tutorial` or `CharacterPreview` existent layers for the LODs? sounds a bit counterintuitive, also I need the main camera to be able to render the LODs layer so the `CharacterPreview` wouldn't be an option. I just checked all the other existent layers and none looks intuitive to be used for the LODs, and if I reuse other existent layer I'd have to add it to this ignored layers list and that would mess up the culling on other objects using those existent layers.

I made sure the new layer is not colliding with anything.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarsLODController.cs
```diff
@@ -0,0 +1,147 @@
...
+                            {
+                                enabled = true;
+                                CommonScriptableObjects.playerUnityPosition.OnChange += OnMainPlayerReposition;
+                            }
+                            else
+                            {
+                                enabled = false;
+                            }
+                        });
+        }
```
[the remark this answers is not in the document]
pravusjif: Done, check it out, although I'm not so sure everyone will like this new "responsibility" inside the KernelConfig.

unity-renderer/Assets/Rendering/Culling/Interfaces/CullingControllerSettings.cs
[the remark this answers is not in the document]
AjimenezDCL: Oh, that makes sense, thanks!

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarsLODController.cs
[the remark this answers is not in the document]
AjimenezDCL: It's not a new responsibility, `KernelConfig` was holding this data previosuly; now only bypasses it. This allow us to avoid communicating systems when they only want to transfer data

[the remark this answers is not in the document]
pravusjif: In the end I had to revert the change because with master's latest assemblies refactor there is a magical cyclic reference when referencing the DataStore assembly in the KernelConfig assembly, even though the DataStore doesn't use the KernelConfig at all  `¯\_(ツ)_/¯`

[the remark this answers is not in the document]
AjimenezDCL: This should be impossible, `DataStore` should be a "leaf" assembly only referencing models and therefore avoiding cycles. Let's not block this PR but create a ticket to solve the issue.

unity-renderer/Assets/Rendering/Culling/Interfaces/CullingControllerSettings.cs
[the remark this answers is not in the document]
Suduck: If this layer is only used for the culling controller ( and we don't plan to filter anything by avatar lods layer), I think we should call it "CullingIgnore" because there is a maximum of 32 layers and it is more likely that there is something more that we want to ignore in the culling controller

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarsLODController.cs
pravusjif: I went to the Assemblies Nightmare Forest and finally got out after 2 days, it's done!

Suduck (approved):
LGTM! Good job!

sandrade-dcl (approved):
I like it! Good work!

unity-renderer/Assets/Rendering/Culling/Interfaces/CullingControllerSettings.cs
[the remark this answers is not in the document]
pravusjif: true that, done!

PR #687, opened 2021-06-25
fix: builder in world unduly playing sounds after the world is loaded

pbosio:


AjimenezDCL (approved):

pravusjif (approved):
Tested and the BiW sound in genesis plaza is gone, great job!

PR #688, opened 2021-06-27
feat: streaming wasm

menduz:
Testing startup time improvements and server improvements

- [x] Requires https://github.com/decentraland/explorer/pull/2450 to be merged before merging
- [x] Create specialized CDN capable of delivery of unity assets with proper mime types 
- [x] Enable automatic Content-Encoding compression of streaming assets
- [x] Adapt explorer to delegate unity files to the renderer package directly https://github.com/decentraland/explorer/pull/2450
- [x] New changes should continue working with current CDN (renderer-artifacts) https://github.com/decentraland/cdn-uploader/
- [x] Control unity initialization from this repository https://github.com/decentraland/unity-renderer/pull/689

Streaming: https://play.decentraland.zone/index.html?DEBUG_ANALYTICS=1&position=1%2C-1&realm=artemis-amber&renderer=urn%3Adecentraland%3Aoff-chain%3Arenderer-artifacts%3Atest%2Fstreaming-wasm
Non-streaming: https://play.decentraland.zone/index.html?DEBUG_ANALYTICS=1&position=1%2C-1&realm=artemis-amber&renderer=urn%3Adecentraland%3Aoff-chain%3Arenderer-artifacts%3Amaster

streaming results (brotli): `loading_time: 4700ms` 
Screen Shot 2021-07-12 at 18 53 39

streaming results (gzip): `loading_time: 5527ms 5100ms`
Screen Shot 2021-07-12 at 14 09 13

non-streaming results: `loading_time: 6710ms 6806ms`
Screen Shot 2021-07-12 at 14 10 03

CC @sicarul


kuruk-mm:
Streaming:
Screenshot from 2021-07-14 11-04-12

Non-streaming:
Screenshot from 2021-07-14 11-05-48

kuruk-mm (approved):
Looks great!

sicarul:
When opening DCL in this version, sometimes i don't see the loading screen (The loading bar with the %) and instead see this interface, at least on Chrome @ Linux with the Intel graphics card (10th gen, i9-10885H), is this intentional? 
image

I also measured with a virtual chronometer the times since page open until the user falls into the world  and the animation becomes smooth (With metamask auto-logging me in), it was (in this machine) consistently ~32 seconds with cache disabled for both versions. Is it expected that the total loading time will remain the same even if we shave off the those ~5 seconds of unity loading at the beginning? Are we just doing something like "lazy load" where we just load those same resources later in the process?

menduz:
@AjimenezDCL Do you have any idea about that? ☝️

AjimenezDCL:
I cannot reproduce it in this branch, but it surely related to the loading screen port. What's your reproduction percentage @sicarul ? do you see it often? could I borrow some of your time to do some testing?

sicarul:
@AjimenezDCL @menduz  i can't reproduce it anymore, maybe it's some kind of caching artifact?

AjimenezDCL:
If you were to use an up-to-date `kernel` with an old `unity-renderer`. The issue you posted would be exactly what you would experience.

The new loading was merged literally 4 hours ago: `52077511db367eaf0551214f6ffd7f2f16962e36` https://github.com/decentraland/unity-renderer/commits/master

So basically my theory is that you tested this PR  with `.zone` containing the `kernel` and the new loading screen but the renderer in this branch was not up to date yet (I see @menduz merged `master` around that time). I think we are good to go, but let's keep an eye on it.

sicarul:
That makes sense from my PoV, any comments about this though?:
> Is it expected that the total loading time will remain the same even if we shave off the those ~5 seconds of unity loading at the beginning? Are we just doing something like "lazy load" where we just load those same resources later in the process?

unity-renderer/ProjectSettings/ProjectSettings.asset
```diff
@@ -574,13 +574,13 @@ PlayerSettings:
...
-  webGLAnalyzeBuildSize: 0
+  webGLAnalyzeBuildSize: 1
   webGLUseEmbeddedResources: 0
-  webGLCompressionFormat: 0
+  webGLCompressionFormat: 2
   webGLWasmArithmeticExceptions: 1
   webGLLinkerTarget: 1
   webGLThreadsSupport: 0
-  webGLDecompressionFallback: 1
+  webGLDecompressionFallback: 0
```
BrianAmadori, line 583: Disabling the decompression fallback may produce problems when trying to run the build from `localhost` as the local servers may not provide the correct mime-type. @menduz did you confirm this is still working?

menduz: I will need some help with that validation for `localhost`. For the initial tests, I changed the CDN and it disables streaming with a warning but also works correctly. Do you have the means to test it quicky to validate?

menduz:
@sicarul 
> Is it expected that the total loading time will remain the same even if we shave off the those ~5 seconds of unity loading at the beginning? Are we just doing something like "lazy load" where we just load those same resources later in the process?

This PR only aims to improve the initialization time and to reduce CPU usage to decompress the brotli in JS (now the browser does it in background, _while_ downloading).

The overall time may get better but it wouldn't be related to this. It is a different piece of code.

The main difference is more noticeable in CPU constrained environments, like @kuruk-mm  https://github.com/decentraland/unity-renderer/pull/688#issuecomment-879926247 getting from 4.3sec to 2.9sec, almost 33% better timing.

sicarul:
Ok that makes sense, looking at the logs, i see that it only reduced one second in my machine using those logs:
image
image

If it'll help low end hardware then it's 1000% worth it

PR #689, opened 2021-06-28
feat: move player configuration out of explorer

menduz:
## What does this PR change?
Moves the configuration of the renderer to this repository, to continue the decoupling of kernel, so it doesn't need to know anything about the artifacts generated by this file. Only using the well known API 

## Why?
In order to be more gentle with browser's CPU, we would need to take advantage of network compression instead of the Unity JS fallback to uncompress the builds, framework and data files. That is currently no possible, since the default behavior is to create `.unityweb` files compressed with a custom algorithm that only works in JS and all of our browsers are well capable of performing such tasks in optimized threads.

Not only that is an optimization by itself, but also browsers support incremental WASM compilation with network streams. That means the WASM can be processed, optimized and compiled way faster, because those processes would start before the bundle.wasm finishes loading, like in a graphics pipeline.

To make this sort of changes plug and play, we are moving the control of the selection of project files (`.unityweb`) to this repo, isolating and encapsulating the loading behaviors in the every version of the renderer.

## How to test it?
- Make sure it works on the provided PR link. https://play.decentraland.zone/branch/feat/new-renderer-initialization-interface/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:feat/move-unity-configuration-to-renderer
- Once merged, we can update explorer to use the new method in new PRs.
- After a couple of releases, we can block the old vanilla method to prevent missusage.

## Next steps
- After this PR is merged to master, we can update the new renderer version in explorer inside this PR https://github.com/decentraland/explorer/pull/2450.


browser-interface/src/index.ts
```diff
@@ -1 +1,89 @@
...
+
+    // This function is called from the unity renderer to send messages back to the scenes
+    MessageFromEngine(type: string, jsonEncodedMessage: string) {
+      onMessageLegacy(type, jsonEncodedMessage)
+    },
+  }
+
+  const originalUnity = await createUnityInstance(canvas, config, onProgress, onSuccess, onError)
+
+  // TODO: replace originalUnity.errorHandler with a version that redirects errors to -> onError
```
kuruk-mm, line 83: Should we add an action point for this? I don't understand well enough the context.

menduz: It is a slow migration, no need to do it on this PR, the comment is there to generate awareness that `onError` should be the unique place where we will handle errors from unity-renderer, to remove the specialized code from `kernel`

pravusjif (approved):

PR #699, opened 2021-06-28
fix: fixed bug of reload scenes when you don't have permission

Suduck:
## What this PR includes?

When an user tap on enter in builder in world in a land where the user don't have permission, the scene reloads.

This fix that issue

## How to test it?
1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/biw-genesis-dissapear
2. Try to enter in a land where you don't have permissions ( like genesis plaza) and the world should stay

## Our Code Review Standards


sandrade-dcl (approved):
Reviewed and tested. It is working fine for me 👍

PR #710, opened 2021-06-29
fix: [BiW Editor] Change rotation and position of the initial camera

### LINKED ISSUE #694: BiW change rotation and position of the initial camera
The initial camera while entering the editor makes no sense. 
It needs to be changed, so it always looks at the scene from the North to South, and in an angle similar to this.!

OLD

NEW

sandrade-dcl:
### Fixes #694
The initial camera while entering the editor makes no sense.
It needs to be changed, so it always looks at the scene from the South-West to the middle. Something like this:

**OLD:**

**NEW:**
image

## How to test the changes?

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/BiW-change-rotation-and-position-of-the-initial-camera
2. Go to one of your owned lands.
3. Press K to open the BiW editor.
4. See the initial camera position (it should be the same indepent of the avatar position when you pressed K).
5. Move the camera along the scene.
6. Press `Shift + C` to reset the camera.
7. Notice the camera position is the same as initiallly.

## Our Code Review Standards


Suduck (approved):
GJ! This looks much better than before!

One more thing before merge, please apply the changes to the CameraPrefab ( if the values are different from the default ones), this way if someone quit the prefab and adds it again we still have the values!

sandrade-dcl:
> GJ! This looks much better than before!
> 
> One more thing before merge, please apply the changes to the CameraPrefab ( if the values are different from the default ones), this way if someone quit the prefab and adds it again we still have the values!

It is not needed to update the prefab because it already got the values that I put as default in code, and these values are the same as we are currently using in the scene 👍 Thanks for the feedback!

image

PR #733, opened 2021-07-01
fix: fix exit view camera, undo/redo from entity panel, point value, land check, hide ui, correct url

### LINKED ISSUE #725: BIW: bug fixing
- [x] Transform made by the entity info panel are not added to the undo/redo list

- [x] In the entity info panel the '.'(coma) and the '.'(point) are not working equally in the scale,rotation and position

- [x] If you move the camera outside your scene, you are not able to exit builder in world since it says that you don't have permission in this land

- [x] If you hide the UI the explorer bar appear

- [x] Now it have into account if you are connected to ZONE or ORG

Suduck:
## What does this PR change?
Fixes #725 

Fixes a couple of biw bugs

- [x] Transform made by the entity info panel are not added to the undo/redo list

- [x] In the entity info panel the '.'(coma) and the '.'(point) are not working equally in the scale,rotation and position

- [x] If you move the camera outside your scene, you are not able to exit builder in world since it says that you don't have permission in this land

- [X] If you hide the UI the explorer bar appear

- [X] Now it have into account if you are connected to ZONE or ORG

## How to test the changes?

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:feat/biw-bug-fixing-5
2. Enter in builder in world
3. Test the bugs mentioned in this PR

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldController.cs
```diff
@@ -341,28 +341,27 @@ public void ChangeEditModeStatusByShortcut()
...
+
         FindSceneToEdit();
 
         if (!UserHasPermissionOnParcelScene(sceneToEdit))
         {
             ShowGenericNotification(BuilderInWorldSettings.LAND_EDITION_NOT_ALLOWED_BY_PERMISSIONS_MESSAGE);
             return;
         }
-        else if (IsParcelSceneDeployedFromSDK(sceneToEdit))
+        if (IsParcelSceneDeployedFromSDK(sceneToEdit))
```
sandrade-dcl, line 357: Why did you remove the `else` here? In case the user don't have pemissions to edit a land and also this scene is created from the SDF, do we want to show both notifications at the same time? Would it not a bit weird?

sandrade-dcl (approved):

pravusjif (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldController.cs
Suduck: The else here is redundant, if the user doesn't have permission on the parcel scene we will return and we will never reach that part of the code.  So the else is not needed since the user doesn't have permission on the parcel scene it will never check if it is from SDK. And if the user has permission it will always check if it is from the SDK

I really don't mind putting it back

sandrade-dcl: Oh I didn't see this `return;`, perfect then! Thanks!

PR #746, opened 2021-07-02
fix: gltf loading after entity is destroyed

pbosio:
we got a race condition where gltf was still loaded even if the entity that the gltf was attached to was already destroyed before gltf finished loading.


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/Tests/LoadWrapperShould.cs
```diff
@@ -0,0 +1,59 @@
...
+        entity.OnCleanupEvent?.Invoke(entity);
+
+        yield return new WaitUntil(() => loaded || failed || unloaded);
+
+        Object.Destroy(meshRootGameObject);
+
+        Assert.IsTrue(unloaded, "Unload should be called if entity is cleaned up while loading mesh");
+    }
+
+    class LoadWrapper_GLFT_Overload : LoadWrapper_GLTF
```
BrianAmadori, line 49:
```suggestion
    class LoadWrapper_GLTF_Overload : LoadWrapper_GLTF
```

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/Tests/LoadWrapperShould.cs
```diff
@@ -0,0 +1,59 @@
...
+public class LoadWrapperShould : IntegrationTestSuite
+{
+    protected override PlatformContext CreatePlatformContext() { return DCL.Tests.PlatformContextFactory.CreateWithGenericMocks(WebRequestController.Create()); }
+
+    [UnityTest]
+    public IEnumerator UnloadWhenEntityDestroyedBeforeFinishing()
+    {
+        GameObject meshRootGameObject = new GameObject();
+
+        string url = TestAssetsUtils.GetPath() + "/GLB/Lantern/Lantern.glb";
```
BrianAmadori, line 23:
```suggestion
        string url = TestAssetsUtils.GetPath() + "/GLB/Lantern/Lantern.glb";
```

The Lantern glb is pretty heavy, maybe we can make do with another model that takes less time to load?

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/Tests/LoadWrapperShould.cs
BrianAmadori: I think we can use `NSubstitute.ForPartsOf()` and `Unload().Received()` instead of inheriting like this.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/GLTF/Tests/AnyAssetPromiseShould.cs
```diff
@@ -137,5 +140,55 @@ public IEnumerator ForceNewInstanceIsOnMultipleTimes()
...
+        [UnityTest]
+        public IEnumerator NotTryToLoadAfterForget()
+        {
+            var keeper = new AssetPromiseKeeper_GLTF();
+
+            var promises = new List<AssetPromise_GLTF>();
+            var forgottenPromises = new Dictionary<int, bool>();
+            bool waitMasterPromise = true;
+
+            string url = TestAssetsUtils.GetPath() + "/GLB/Lantern/Lantern.glb";
```
BrianAmadori, line 153: Same as before, this glb is quite heavy.

BrianAmadori (commented):
❤️

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/Tests/LoadWrapperShould.cs
pbosio: Will `NSubstitute` use the original `Unload` method or it will mock it?
I'm not entirely sure about using `NSubstitute` for the class we are actually testing.
For example, if `NSubstitute` mocks the `Unload` we will never know if there is an Exception happening inside that method under this conditions.
I will check that out, but if the original method is not called I think we should keep that as it is :)

AjimenezDCL: Yep, `NSubstitute.ForPartsOf<>` will use the implementation unless you override it. Brian's approach should be correct here.

AjimenezDCL (approved):
Besides the stuff mentioned by Brian, the rest looks good. Great job!

sandrade-dcl (approved):

PR #747, opened 2021-07-02
fix: gltf importer texture scale hiccup

BrianAmadori:
## What does this PR change?

This PR reimplements our texture resizing code to happen entirely GPU side. This code is used by the GLTF importer (and will potentially be used for more optimizations). In the previous implementation, memory had to be copied between GPU and CPU, causing big noticeable hiccups.

## How to test the changes?

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/gltf-importer-texture-scale-hiccup&ENV=org&NO_ASSET_BUNDLES
2. No new texture artifacts should be observed for models that use textures that go up of 1024x1024

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/UUIDComponent/InteractionHoverCanvas/InteractionHoverCanvasController.cs
```diff
@@ -41,7 +41,8 @@ public void Setup(string button, string feedbackText, IDCLEntity entity)
 
     void ConfigureIcon(string button)
     {
-        hoverIcon?.SetActive(false);
+        if ( hoverIcon != null )
+            hoverIcon.SetActive(false);
```
sandrade-dcl, line 45: Just as a curiosity: Why did you have to make this change? To use the operator (`?`) is the same as check the null in the conditional, right?

sandrade-dcl (approved):
LGTM!

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/UUIDComponent/InteractionHoverCanvas/InteractionHoverCanvasController.cs
BrianAmadori: On Unity objects, the equality and inequality operators are overloaded. Using the null-conditional operator `?` bypass this checks. So in theory using them on those objects can cause problems. (even Rider gives out a warning about this).

This until now caused no problems, but started to crash when I enabled the "Enter Play Mode Options" checkbox that for some strange reason speeds up the editor by a LOT.
image

sandrade-dcl: Good to know! Thanks!

pravusjif (approved):
also tested the build with no problems

Suduck (approved):

PR #756, opened 2021-07-05
fix: biw smart item compatibility

### LINKED ISSUE #744: BIW: smart item compatibility with builder

Suduck:
Fixes #744 
## What does this PR change?

This PR adds the compatibility with builder around smart items so you can load the builder scenes even if they have smart items inside it 

It also include a quick fix

- [x] If you try to enter in builder in world while the scene is still not loaded (from the shorcut) it won't start the builder in world flow 

## How to test the changes?

1. Go to: https://play.decentraland.zone/branch/feat/biw-smart-item-handle/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/biw-fix-bug
2. Open builder in world land with an scene that you has previously created in builder with smart items 
3. Made a change in the scene and open again the scene in the builder
4. All the connections in the smart items should remain

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLName.cs
```diff
@@ -17,6 +17,8 @@ public class DCLName : BaseDisposable
     public class Model : BaseModel
     {
         public string value;
+        //Note: This value is used for builder to manage the smart items, when the builder is no longer active we should remove it
```
sandrade-dcl, line 20: Only as suggestion: I would make this note as a `TO DO` in order to have it more localized in the future.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldController.cs
```diff
@@ -485,6 +485,10 @@ public void TryStartEnterEditMode(bool activateCamera, IParcelScene targetScene
             return;
         }
 
+        //If the scene is still not loaded, we return as we still can't enter in builder in world 
+        if (sceneToEditId != null)
+            return;
```
sandrade-dcl, line 490: Only by curiossity: Why it is necessary to do this check by second time? As I can see, the `FindSceneToEdit()` function is not any kind of coroutine nor an async method... in what case could we get the `sceneToEditId` as null in this point?

sandrade-dcl (approved):
Good job! Approved! I only left some doubts.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldController.cs
Suduck: In the case that we try to enter in the land but the scene is still not loaded, so when we try to find the scene we can't find it since it is not loaded yet, To do this you need to enter with the 'K' key very quickly after we exit.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/BuilderInWorld/DCLName.cs
Suduck: Will do in the current branch that I am working!

PR #759, opened 2021-07-05
fix: disable unused webgl data caching to fix firefox incognito mode …

### LINKED ISSUE #591: Landing page: Forever loading with FF in incognito mode
This happens because FF in incognito has disabled IDB but somehow the IDB object is not null. 

In this case we should handle the issue and crash the application instead of having loading forever.

kuruk-mm:
…infinity loading

Fix #591 

## What does this PR change?

Seems that Unity 2020.3 introduced `internal preprocessor` and enabled the feature of WebGL caching that we don't use, and that made Firefox in Incognito Mode loads forever (Unity bug).

Unity issue: https://forum.unity.com/threads/unity-2020-2-2021-1-unitycache-indexeddb-database-could-not-be-opened-in-firefox-private-mode.1047761/

## How to test the changes?

1. Open Firefox Incognito Mode
2. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/ff-incognito&ENV=org
3. It works?

## Our Code Review Standards

### DIFF
--- a/unity-renderer/ProjectSettings/ProjectSettings.asset
+++ b/unity-renderer/ProjectSettings/ProjectSettings.asset
@@ -569,7 +569,7 @@ PlayerSettings:
   webGLMemorySize: 512
   webGLExceptionSupport: 1
   webGLNameFilesAsHashes: 0
-  webGLDataCaching: 1
+  webGLDataCaching: 0
   webGLDebugSymbols: 1
   webGLEmscriptenArgs: 
   webGLModulesDirectory: 


Suduck (approved):
This change won't affect the cache of others browsers?

AjimenezDCL (approved):

BrianAmadori (approved):

kuruk-mm:
> This change won't affect the cache of others browsers?

Yes, but we don't use this kind of cache. It was accidentally activated because Unity 2020.3 introduced `internal preprocessor`.

PR #760, opened 2021-07-05
fix: improve gltf external texture caching

pbosio:
Before, loading different GLTFs that pointed to the same external texture hash did not used the cache correctly, and two identical images were generated. This PR fixes this.


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/GLTF/AssetPromise_GLTF.cs
```diff
@@ -46,19 +46,24 @@ protected override void OnLoad(System.Action OnSuccess, System.Action OnFail)
...
         }
 
-        void ParseGLTFWebRequestedFile(ref string requestedFileName) { provider.TryGetContentsUrl(assetDirectoryPath + requestedFileName, out requestedFileName); }
+        bool FileToHash(string fileName, out string hash)
+        {
+            if (provider.TryGetContentHash(assetDirectoryPath + fileName, out hash))
+            {
+                return true;
+            }
+            return false;
```
AjimenezDCL, line 65: you could use
`return provider.TryGetContentHash(assetDirectoryPath + fileName, out hash);`

unity-renderer/Assets/UnityGLTF/Scripts/Cache/AssetCache.cs
```diff
@@ -37,8 +36,9 @@ public static RefCountedTextureData AddImage(string uri, string idSuffix, Textur
         /// <param name="texture">The texture to cached image</param>
         public static RefCountedTextureData AddImage(string fullId, Texture2D texture)
         {
-            ImageCacheByUri[fullId] = new RefCountedTextureData(fullId, texture);
-            return ImageCacheByUri[fullId];
+            var refCountedTextureData = new RefCountedTextureData(fullId, texture);
+            ImageCacheByUri[fullId] = refCountedTextureData;
+            return refCountedTextureData;
```
AjimenezDCL, line 41: This should be equivalent, not requesting a change, just curious if this looks more readable to you

AjimenezDCL (approved):
Good job, havent test it but codewise looks good.

We should avoid autoformatting plugin files (such as : `UnityGLTF/Scripts/Cache/AssetCache.cs `) to ease conflict issues when upgrading it.

In any case this should be a different PR to add the excepction to thepre-commit hook.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/GLTF/AssetPromise_GLTF.cs
pbosio: Ups! I'll change it. 
I forget it like this while debugging because it easier to add breakpoints this way :P

unity-renderer/Assets/UnityGLTF/Scripts/Cache/AssetCache.cs
pbosio: the alternative would be 
```
ImageCacheByUri[fullId] = new RefCountedTextureData(fullId, texture);
return ImageCacheByUri[fullId];
```
but I dislike very much the idea of accessing a dictionary element by key twice when it can be avoided :)

AjimenezDCL: Well, accessing a dictionary is O(1) but yeah no big deal

pravusjif (approved):
Also tested with `NO_ASSET_BUNDLES` and saw no issues.
https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/better-gltf-texture-caching&ENV=org&NO_ASSET_BUNDLES

BrianAmadori (changes requested):
Can we add some testing on this? I'm not comfortable with adding a big change like this and not having some kind of coverage.

The test can go like this:
- In `unity-renderer\TestResources\GLTF\SimpleCube` there's a simple cube I made in blender that points to a external texture, so we can load this cube and make `GLTFComponent` the cube has id 0.
- We can load the same cube, and fool `GLTFComponent` to believe the cube has id 1.
- Ensure that the external texture is reused and not cloned.

pbosio:
> Can we add some testing on this? I'm not comfortable with adding a big change like this and not having some kind of coverage.
> 
> The test can go like this:
> 
>     * In `unity-renderer\TestResources\GLTF\SimpleCube` there's a simple cube I made in blender that points to a external texture, so we can load this cube and make `GLTFComponent` the cube has id 0.
> 
>     * We can load the same cube, and fool `GLTFComponent` to believe the cube has id 1.
> 
>     * Ensure that the external texture is reused and not cloned.

I've ended up adding 2 more GLTF for this test, since we needed different GLTF sharing the same texture

PR #775, opened 2021-07-07
refactor: biw refactor

### LINKED ISSUE #706: BIW: Refactor to remove the prefab from initial scene
- [x] Refactor  BIW to remove the monobehaviour dependency using the SceneReference prefab

- [x] Remove the monobehaviour part of the controllers removing the whole prefab

- [x] Remove the dependency with builder

- [x] Remove the monobehaviour part of the other parts like entities

    The last past includes:    

- [x] Remove monobehaviour entities

- [x] Remove Builder MeshLoadingIndicatorController

- [x] Remove Builder Input prefab
 
- [ ] Remove audio prefab <- will do later as a separate refactor

Suduck:
## What does this PR change?
Fixes #706 

This is a part of the refactor of builder in world where we get out the prefab from the initial scene and we only instantiate it if the feature is active

## How to test the changes?

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:refactor/biw
2. Enter builder in world 
3. Check that everything is worked as expected

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Controllers/BIWPublishController.cs
```diff
@@ -23,6 +24,9 @@ public override void Init()
             HUDController.i.builderInWorldMainHud.OnPublishAction += StartPublishFlow;
             HUDController.i.builderInWorldMainHud.OnConfirmPublishAction += StartPublishScene;
         }
+
+        builderInWorldBridge = InitialSceneReferences.i.builderInWorldBridge;
```
sandrade-dcl, line 28: Suggestion: maybe it would be fine to pass the `InitialSceneReferences` as injected reference in the `BIWController.Init()` function and not have to always depend of the singleton instance. In this way we would be following the dependency injection philosofy and would ease the testing using mocks (in this case, for example we could have a mocked `InitialSceneReferences` with a mocked `builderInWorldBridge` and we wouldn't have to depend of a "real" instance of `InitialSceneReferences`).

The same for the rest of `Init()` functions of each `BIWController`. I think it could be a good practice to apply.

Anyway let me know waht do you think? I may be missing something that I'm not taking into account 👍

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Tests/BIWSaveControllerShould.cs
```diff
@@ -15,8 +16,8 @@ protected override IEnumerator SetUp()
         yield return base.SetUp();
         gameObject = new GameObject();
 
+        builderInWorldBridge = InitialSceneReferences.i.builderInWorldBridge;
```
sandrade-dcl, line 19: I guess this `InitialSceneReferences.i.builderInWorldBridge` should be loaded with the correct references in the test scene, right?

Following with the previous suggestion about to inject the `InitialSceneReferences` as reference in each BIW controller, I think doing this we could make these kind of tests independent of any instanced game object in the scene and we could use mocked specific instances only for testing.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/FeatureController/FeatureController.cs
```diff
@@ -0,0 +1,59 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using DCL;
+using UnityEngine;
+
+public class FeatureController
```
sandrade-dcl, line 7: Is this class something for only the BuilderInWorld feature or for other ones too? Can we add some kind of generic documentation in the class header describing its main responsibility?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/FeatureController/FeatureTemplate/Feature.cs
```diff
@@ -0,0 +1,8 @@
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class Feature
```
sandrade-dcl, line 5: The same as before... it would be fine to add some kind of description about what this class represents.

sandrade-dcl (changes requested):
I left some suggestions/doubts that I would like to confirm with you before give the approve. Thanks!

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Controllers/BIWPublishController.cs
Suduck: The problem with that is that `InitialSceneReference ` can destroy his current instance (imagine that we reload the initial scene, or load an additional scene that changes the scenes references).  Thats been said, it is probably that this won't happen. 

`InitialSceneReference `  is a special class that only references the monobehaviours of the initial scene, I think this is the only case where the Singleton pattern is OK. 

I think that create a mocked `InitialSceneReference ` instance won't be a good choice since it will need to instantiate gameobjects (This is the current solution for reference monobehaviour in scene) and it will create more overheat to the test increasing their duration.

I can do that without problem but that depends on the future implementation of `InitialSceneReference`

@BrianAmadori  what do you think?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Tests/BIWSaveControllerShould.cs
Suduck: Yes, it have the correct reference in the test scene. 

Anyway, you still need to instanciate the gameObjects since this class reference to monobehaviours. If brian says that we can go for the dependency injection, I will change it!

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/FeatureController/FeatureController.cs
Suduck: This is something more generic. Sure! I will add some documentation to both parts

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Controllers/BIWPublishController.cs
sandrade-dcl: Yeah, I think it is a good momento to open this discussion and decide what approach to follow in these cases. For me, the less dependencies a unit test has on a Unity scene, the better. This would force us to create the needed game objects in runtime but I think the non-dependency of anything continues having more benefits.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Tests/BIWSaveControllerShould.cs
BrianAmadori: Remember that right now we are trying to get rid of `IntegrationTestSuite_Legacy` and this means we are going to get rid of the test scene eventually. This said, if we go for the dependency injection we eventually could load the references without needing the scene, so this would help to get to that point. What do you think?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Controllers/BIWPublishController.cs
BrianAmadori: I think we can keep the singleton but in this case inject the dependency. This would help us to get rid of the scene references in the future as I stated in the other comment. I think that using a singleton is a last resort tool, we should try to inject dependencies if it makes sense.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Tests/BIWSaveControllerShould.cs
Suduck: Sure! Lets change it then as a dependency injection!

PR #788, opened 2021-07-12
feat: biw save each 5 seconds

Suduck:
## What does this PR change?

This PR ensures that we are saving biw scenes each 5 seconds instead of each change, this will make the servers more stables

## How to test the changes?

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/biw-save-time
2. Enter builder in world
3. Check that the scene is not saved each time that you made a change

## Our Code Review Standards


sandrade-dcl (approved):
The code looks fine!

My only concern about to left 5 seconds between saves is that it could happens that if we exit from the editor in less than 5secs from the last save, we can be breaking the scene screenshot flow. Please take a good look to the use of `numberOfSaves` variable that we set to `true` in the `BIWSaveController.ForceSave()` function and is checked in `StartExitMode()` and `ExitEditMode()` in the `BuilderInWorldController`. If you can confirm this flow is working fine, great!

PR #789, opened 2021-07-12
refactor: remove biw controller monobehaviour

Suduck:
## What does this PR change?

This PR removes completely the monobehaviour from the builder in world controllers. 

The controllers have been refactored to be interfaces so we can have an easy unit testing development

Now it will use the FeatureController to control all the monobehaviour callbacks

## How to test the changes?

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:refactor/biw-2
2. Enter builder in world
3. Test that everything works as expected

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BIWContext.cs
```diff
@@ -0,0 +1,82 @@
...
+    private IBIWOutlinerController outlinerControllerValue;
+    private IBIWInputHandler inputHandlerValue;
+    private IBIWInputWrapper inputWrapperValue;
+    private IBIWPublishController publishControllerValue;
+    private IBIWCreatorController creatorControllerValue;
+    private IBIWModeController modeControllerValue;
+    private IBIWFloorHandler floorHandlerValue;
+    private IBIWEntityHandler entityHandlerValue;
+    private IBIWActionController actionControllerValue;
+    private IBIWSaveController saveControllerValue;
```
BrianAmadori, line 44: For each of those, we can do:
```
public IInterfaceClass property { get; private set; }
```
This way we save the `Value` fields boilerplate.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BIWContext.cs
```diff
@@ -0,0 +1,82 @@
...
+        IBIWModeController biwModeController,
+        IBIWFloorHandler biwFloorHandler,
+        IBIWEntityHandler biwEntityHandler,
+        IBIWActionController ibiwActionController,
+        IBIWSaveController biwSaveController)
+    {
+        projectReferencesValue = Resources.Load<BIWProjectReferences>(PROJECT_REFERENCES_PATH);
+        godModeDynamicVariablesValue = Resources.Load<BIWGodModeDynamicVariables>(GOD_MODE_DYNAMIC_VARIABLE_PATH);
+        firstPersonDynamicVariablesValue = Resources.Load<BIWFirstPersonDynamicVariables>(FIRST_PERSON_DYNAMIC_VARIABLE_PATH);
+        inputsReferencesValue = Resources.Load<BIWInputsReferences>(INPUTS_PATH);
```
BrianAmadori, line 60: Maybe we can add the `asset` suffix. So it's more clear that they must be freed.

i.e.: `projectReferencesAsset`

Also, we shouldn't be calling `Instantiate` on these Assets? Because if we don't do so, modifying them in runtime will modify the assets in the project and pollute the git changelist.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Controllers/BIWSaveController.cs
```diff
@@ -2,26 +2,31 @@
...
     {
-        base.Init();
+        base.Init(biwContext);
 
-        builderInWorldBridge = InitialSceneReferences.i.builderInWorldBridge;
-        if (builderInWorldBridge != null)
-            builderInWorldBridge.OnKernelUpdated += TryToSave;
+        bridge = InitialSceneReferences.i.builderInWorldBridge;
+        if (bridge != null)
+            bridge.OnKernelUpdated += TryToSave;
```
BrianAmadori, line 29: I think this relationship is backwards. Bridges call features, not the other way around. Anyway, I think this is not on the scope of this PR.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/States/BuilderInWorldMode.cs
```diff
@@ -35,10 +33,12 @@ public class BuilderInWorldMode : MonoBehaviour
...
-    public virtual void Init()
+    public virtual void Init(BIWContext biwContext)
     {
-        gameObject.SetActive(false);
-        builderInWorldEntityHandler.OnEntityDeleted += OnDeleteEntity;
+        biwEntityHandler = biwContext.entityHandler;
+        biwSaveController = biwContext.saveController;
+        ibiwActionController = biwContext.actionController;
+        biwEntityHandler.OnEntityDeleted += OnDeleteEntity;
     }
```
BrianAmadori, line 42: We should remove all the `ibiw` and `biw` field prefixes here so they are called `entityHandler`, `saveController` and `actionController`

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/FeatureController/FeatureController.cs
```diff
@@ -23,7 +24,14 @@ public void Start()
         KernelConfig.i.OnChange += OnKernelConfigChanged;
     }
 
-    // Update is called once per frame
+    public void OnGUI()
```
BrianAmadori, line 27: If we are using `OnGUI` only for the selection box, maybe we can extract the selection object to a specific `MonoBehaviour` to avoid exposing `OnGUI` to all the features? I think this can help to mitigate further usages, as this event is pretty controversial.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/FeatureController/FeatureController.cs
```diff
@@ -32,27 +40,60 @@ public void Update()
...
         {
-            if (builderInWorld != null)
+            if (activeFeatures[i].GetType() == typeof(T))
                 return;
-            builderInWorld = GameObject.Instantiate(builderInWorldFeaturePrefab);
-
         }
-        else
+
+        Feature feature = (Feature) Activator.CreateInstance(typeof (T));
```
BrianAmadori, line 83: Be aware that `Activator.CreateInstance` is excruciatingly slow. We can avoid it by using the `new()` generic constraint.

```
private void InitializeFeature<T>() where T : Feature, new ()
{
    ...
    
    Feature feature = (Feature) new T();
}
```

In any case we should revise this code on the future, because we'd want to use custom constructors to DI. So here we most likely need to use an abstract factory, or just pass the new `Feature` object by argument.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/FeatureController/FeatureController.cs
```diff
@@ -32,27 +40,60 @@ public void Update()
...
+        activeFeatures.Add(feature);
+    }
+
+    private void RemoveFeature<T>() where T : Feature
+    {
+        for (int i = 0; i < activeFeatures.Count; i++)
         {
-            if (builderInWorld != null)
-                GameObject.Destroy(builderInWorld);
+            if (activeFeatures[i].GetType() == typeof(T))
```
BrianAmadori, line 92: `activeFeatures` already restricts the type to `Feature` so this check is redundant.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BuildModeHUD/Scripts/BuilderInWorldLoadingController.cs
```diff
@@ -17,9 +18,33 @@ public class BuilderInWorldLoadingController : IBuilderInWorldLoadingController
...
 
-    public void Initialize(IBuilderInWorldLoadingView initialLoadingView) { this.initialLoadingView = initialLoadingView; }
+    private const string VIEW_PATH = "BuilderInWorldLoadingView";
 
-    public void Dispose() { initialLoadingView.StopTipsCarousel(); }
+    public void Initialize() { AsignMainView(CreateView()); }
+
+    public void Initialize(IBuilderInWorldLoadingView view) { AsignMainView(view); }
+
+    private void AsignMainView(IBuilderInWorldLoadingView view)
```
BrianAmadori, line 27:
```suggestion
    private void AssignMainView(IBuilderInWorldLoadingView view)
```

unity-renderer/Assets/Tutorial/Scripts/Steps/BuilderInWorld/TutorialStep_PutSceneObject.cs
```diff
@@ -13,7 +13,7 @@ public class TutorialStep_PutSceneObject : TutorialStep
     public override void OnStepStart()
     {
         base.OnStepStart();
-        builderInWorldController = FindObjectOfType<BIWCreatorController>();
+        builderInWorldController = new BIWCreatorController();
```
BrianAmadori, line 16: <3

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Main/Main.cs
```diff
@@ -96,13 +93,19 @@ private void Update()
             featureController?.Update();
         }
 
-        private void LateUpdate() { Environment.i.world.sceneController.LateUpdate(); }
+        private void LateUpdate()
+        {
+            Environment.i.world.sceneController.LateUpdate();
+            featureController?.LateUpdate();
```
BrianAmadori, line 99: `featureController` will never be null. We really want to check for nullity every frame here?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Controllers/BIWSaveController.cs
Suduck: They way it was designed is that the bridge is only a way to communicate between kernel <> Unity, it only has that responsibility to act as a bridge, It is weird to me that a bridge needs to know when it should say to the `SaveController ` "Save now" since that responsibility is on the `SaveController `

Let me know what do you think

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BIWContext.cs
Suduck: Changed the name.

For the Instantiate, it is not necessary, if you made a change in runtime to the asset, it will be reverted after you exit from play mode

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/FeatureController/FeatureController.cs
Suduck: We just have removed all the monobehaviours from builder in world to try to remove full dependency on them.

Making this change sounds like a hack in the current implementation to me. I totally understand your desire to remove That part but it is more obscure to encapsulate that call in a monobehaviour.

I will rather add it as a technical debt and change it in the future rather than that. Anyway, your call!

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/FeatureController/FeatureController.cs
Suduck: Changed and added a note in the code to have this into account. Should I create a debt tech tiquet?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/FeatureController/FeatureController.cs
Suduck: If I am not mistaken ( I could be), this is getting the type of the Subclass type, (The subclass should inherit from Feature) so this way only the child is removed from the feature

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/FeatureController/FeatureController.cs
```diff
@@ -4,17 +4,18 @@
 using DCL;
 using UnityEngine;
 
+/// <summary>
+/// This class is used to handle a feature that needs the monobehaviour callbacks.
+///
+/// You need to add it as a feature toggle in gitlab so kernel will understand when to activate it and deactivate it.
+/// After that, you feature manager should herith from 'Feature' so it can start receveing monobehaviour callbacks whene it is activated
```
sandrade-dcl, line 11:
```suggestion
/// After that, your feature manager should inherit from 'Feature' so it can start receveing monobehaviour callbacks whene it is activated
```

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/FeatureController/FeatureTemplate/Feature.cs
```diff
@@ -2,7 +2,18 @@
 using System.Collections.Generic;
 using UnityEngine;
 
+/// <summary>
+/// This is the main template for the features handled by FeatureController
+///
+/// This will handle all the monobehaviour callbacks when your feature is activated
+/// </summary>
 public class Feature
```
sandrade-dcl, line 10: If this class will not have any implementation, why don't we convert it into an interface? or into an abstract class al least?

sandrade-dcl (approved):
Due to the big size of this PR, I have tried to review the most important things. Good work, I think it has been improved a lot. I left some comments... tell me what do you think!

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/Controllers/BIWController.cs
```diff
@@ -3,13 +3,13 @@
 using System.Collections.Generic;
 using UnityEngine;
 
-public class BIWController : MonoBehaviour
+public class BIWController
```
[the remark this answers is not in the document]
Suduck: The class can be abstract and I have changed it! however, their methods shouldn't be abstract since it is not mandatory to implement them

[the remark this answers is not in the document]
sandrade-dcl: Oh ok, it's true that not all methods are mandatory. Then let left the class as it is, no problem! 👍

PR #810, opened 2021-07-15
chore: avatars lod impostor textures

pravusjif:
issue: https://app.zenhub.com/workspaces/unity-6047bad476c3c0001942ee91/issues/decentraland/unity-renderer/712

Changed avatars LOD Impostor textures to **content-server-fetched body snapshots** or **pre-generated generic snapshots from a texture atlas** when the user snapshot cannot be fetched.

**Generic snapshots atlas used:**
avatar-impostors-atlas

### Testing with bots (generic body snapshot lod impostors from atlas)
1. enter https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:chore/AvatarLODImpostorTextures&ENV=zone&position=91,151&ENABLE_AVATAR_LODS
2. open browser console

https://user-images.githubusercontent.com/1031741/126836869-329bd619-4903-4071-b759-5e87b9020b76.mov

### Testing with real users (content-server fetched body snapshot lod impostor)
1. enter https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:chore/AvatarLODImpostorTextures&ENV=org&position=-27,60&ENABLE_AVATAR_LODS
2. Try getting close and far away from the other real users to see their LOD Impostor with their snapshot fetched from the content server. It may be better to test on the realm that has more users in that scene (Explore & Play panel may help for this)

https://user-images.githubusercontent.com/1031741/126839336-c4895121-26b0-4843-bfa2-62bf992e471a.mov

### Known Issues
Currently we have 2 known bugs already tracked in specific issues:
* ~[Weird rotation and Y position of impostors](https://app.zenhub.com/workspaces/unity-6047bad476c3c0001942ee91/issues/decentraland/unity-renderer/840)~ [Fixed]
* ~[Incomplete body parts when transitioning from lod](https://app.zenhub.com/workspaces/unity-6047bad476c3c0001942ee91/issues/decentraland/unity-renderer/847)~ [Fixed]


sandrade-dcl:
I have just found several cases where the avatars are keeping as flat sprite even if I'm near to them. I don't know how to reproduce it, I found it testing in Wonderzone.

image
image
image

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarLOD/AvatarsLODController.cs
```diff
@@ -63,11 +63,11 @@ private void UpdateLODsVerticalMovementAndBillboard()
             GameObject lodGO;
             for (int i = 0; i < listCount; i++)
             {
-                lodGO = avatarsList[i].GetLODRenderer().gameObject;
+                lodGO = avatarsList[i].meshRenderer.gameObject;
                 if (!lodGO.activeSelf)
                     continue;
 
-                if (applyVerticalMovement)
+                if (applyVerticalMovement && Vector3.Distance(lodGO.transform.position, CommonScriptableObjects.playerUnityPosition.Get()) > 2 * DataStore.i.avatarsLOD.LODDistance.Get())
```
sandrade-dcl, line 70: Maybe we can move this hardcoded value `2` to some constant with a descriptive name? Or if not, at least we should put some comment above this line explaining why are we multiplying by 2 the `DataStore.i.avatarsLOD.LODDistance.Get()` value.

sandrade-dcl (changes requested):
I found an issue with the LOD sprites. I explain it in the previous comment on this PR.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarLOD/AvatarsLODController.cs
Suduck: Agree!

Suduck (approved):
I think we should remove the movement of the sprites in the Y-axis, since it is a bit weird right now, the avatars near you don't move up and down. Other than that it looks good to me! GB!

pravusjif:
> I have just found several cases where the avatars are keeping as flat sprite even if I'm near to them. I don't know how to reproduce it, I found it testing in Wonderzone.

I was able to fix this issue in this same PR

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarLOD/AvatarsLODController.cs
pravusjif: Removed the "vertical movement" on far away impostors as nobody likes it XD, so this code doesn't exist anymore.

pravusjif:
Managed to fix the "invisible wearables" bug when unlodding an avatar bug in this same PR

sandrade-dcl (approved):
I no longer reproduce the bug of some avatars that stayed with the LOD sprite even if we were near of them.

I think we have to improve the transition between LOD sprites and the 3D models of the avatars, we should review the correct distance with the other avatars to activate/deactivate their LODs, and we should resolve the problem of seeing farest avatars moving with the LOD sprite without any animation. But I know itill be workd that we will be contemplating on future steps, so good work, this feature is taking shape! :)

PR #849, opened 2021-07-27
fix: attached entities follow instant the player

### LINKED ISSUE #449: Entities attached to player lag a little
When attaching an entity to the player, if the player moves or rotates quickly there may be some lag.
A lot less than if manually setting the entity position every frame, but in scenes that aren't so light it can be noticed.

For the record, we refer to doing this:

```ts
openDialogSound.setParent(Attachable.AVATAR)
```

In Discord, devs complain that an entity set in the same position as the player, centered, will not be perfectly in stereo if the player moves. Becasue of this lag, there may be shifts in sound direction.

> Although if you rotate quickly or move around, the 3d object thats attached to the user will lag and make the effect only come out of one speaker for a little
> at the moment its clunky. If the user moves too far left than the sound only goes to one channel. It's not ideal.
> What we would want is that if something is attached to the avatar that it is permanent. No lag.
> This goes for holding items as much as it goes for sounds

You can also sometimes notice this visually, especially if your machine is running with difficulty.
Check the portal scene, where the portal gun does sometimes have a little bit of lag when rotating:
https://play.decentraland.org/?position=-133%2C-40&realm=baldr-amber

Suduck:
## What does this PR change?

This PR fixes #449.

The attached entities follow the player in first person with a delay, now the entities follow the player instantly without a delay

## How to test the changes?

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/attached-entities&position=-133%2C-40&ENV=org
2. Pick up the portal gun
3. See if it follows you correctly
4. Change between 1º and 3º person camera to check that everything work as expected

## Our Code Review Standards


pravusjif (approved):
Is it necessary to change the InitialScene file?

Suduck:
> Is it necessary to change the InitialScene file?

Yes, to add the CameraController as part of the FirstEntityReference, another option will be to use InitialSceneReferences singleton if you want!

sandrade-dcl (approved):

PR #850, opened 2021-07-27
fix: Noticeable hiccups when receiving a chat message and the chat history is long

### LINKED ISSUE #123: Noticeable hiccups when receiving a chat message and the chat history is long
Seems to be that when the chat history gets quite long, every new chat message received takes some milliseconds in processing, and that causes noticeable hiccups

sandrade-dcl:
## What does this PR change?

Fix #123 

Seems to be that when the chat history gets quite long, every new chat message received takes some milliseconds in processing, and that causes noticeable hiccups

### Main cause of the hiccups
- Each time a message entered the chat, we was forcing to update **ALL** the existing entries layouts with `Utils.ForceUpdateLayout(transform as RectTransform, delayed: false)`, that had having a huge impact on the performance when the chat had accumulated a lot of messages.
- After a message entered the chat, we was setting the vertical position of the HUD scroll to 0 with `scrollRect.verticalNormalizedPosition = 0`, and it was affecting negatively to the performance.

### Implemented solutions
- [x] Now, instead of always update all the entries each time a new message entered the chat, we only update the new added entry and, if it is necessary, the entries whose position was changed during the re-ordering by timestamp.
- [x] Now, before setting the vertical position of the HUD scroll to 0, we check if the current position is grater that 0. In this way we will only apply this action when it is needed.
- [x] We have changed the maximum number of entries in the chat (`MAX_CHAT_ENTRIES`) from `100` to `30`. It will make the HUD lighter.

### Metrics obtained from the `AddEntry(...)` action hiccups
- **BEFORE**: Hiccups up to `~287.50ms` per each entry added in the chat.
- **NOW**: Hiccups up to `~3.45ms` per each entry added in the chat.

## How to test the changes?

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/Noticeable-hiccups-when-receiving-a-chat-message-and-the-chat-history-is-long
2. Test the world and private chats.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDController.cs
@@ -5,7 +5,7 @@ using UnityEngine.Events;
 
 public class ChatHUDController : IDisposable
 {
-    public static int MAX_CHAT_ENTRIES { internal set; get; } = 100;
+    public static int MAX_CHAT_ENTRIES { internal set; get; } = 30;
 
     public ChatHUDView view;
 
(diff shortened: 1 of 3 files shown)


pravusjif (approved):
awsome job!

Suduck (approved):

PR #857, opened 2021-07-27
refactor: AssetBundle and GLTF AssetLibrary

pbosio:
Refactor to avoid the strange situation when an Asset Bundles and GLTF collide with each other in the pool while having the same id.


BrianAmadori (approved):
LGTM. Tested it for a while and found no issues.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Common/AssetLibrary_Poolable.cs
```diff
@@ -6,18 +6,26 @@ namespace DCL
     public class AssetLibrary_Poolable<AssetType> : AssetLibrary<AssetType>
         where AssetType : Asset_WithPoolableContainer
     {
+        private readonly string salt = null;
```
sandrade-dcl, line 9: What does "salt" mean? Maybe we could find a more descriptive name.

sandrade-dcl (approved):

PR #862, opened 2021-07-28
feat: expanded audio settings

Landeplage:
### What does this PR change?
The audio section allows players to control all the audio channels for separate instead of toggling on/off all the platform sounds.

### How to test the changes?
1- Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:origin/feat/expanded_audio_settings&ENV=org
2- Open the settings panel
3- Try different sound configurations.

**_Disclaimer: sound toggle in the general section will be removed._**


Landeplage:
We are still discussing whether or not the "Music" slider should control music inside scenes. This would require a feature in the builder, where authors can tag certain audio sources as either music or sfx.

I have managed to test everything except audio from video streams. This still needs to be tested.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
```diff
@@ -45,7 +45,17 @@ public class DCLVideoTexture : DCLTexture
...
+        {
+            model = new Model();
+
+            DataStore.i.virtualAudioMixer.sceneSFXVolume.OnChange += OnVirtualAudioMixerChangedValue;
+        }
+
+        ~DCLVideoTexture()
+        {
+            DataStore.i.virtualAudioMixer.sceneSFXVolume.OnChange -= OnVirtualAudioMixerChangedValue;
+        }
```
BrianAmadori, line 58: Please don't use finalizers, they are not recommended. Use the `Dispose()` pattern instead.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
```diff
@@ -228,15 +238,21 @@ private void CalculateVideoVolumeAndPlayStatus()
...
                 return;
 
             float targetVolume = 0f;
 
-            if (CommonScriptableObjects.rendererState.Get() && IsPlayerInSameSceneAsComponent((CommonScriptableObjects.sceneID.Get())))
-                targetVolume = baseVolume * distanceVolumeModifier * Settings.i.generalSettings.sfxVolume;
+            if (CommonScriptableObjects.rendererState.Get() && IsPlayerInSameSceneAsComponent((CommonScriptableObjects.sceneID.Get()))) {
+                targetVolume = baseVolume * distanceVolumeModifier;
+                targetVolume *= Utils.ToVolumeCurve(DataStore.i.virtualAudioMixer.sceneSFXVolume.Get() * Settings.i.currentAudioSettings.sceneSFXVolume * Settings.i.currentAudioSettings.masterVolume);
+            }
```
BrianAmadori, line 255: Can you spread this code a bit more using temporary variables if necessary? It's getting very hard to read as it is right now.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Settings/Settings.cs
```diff
@@ -46,6 +53,14 @@ public Settings()
 
             LoadQualitySettings();
             LoadGeneralSettings();
+            LoadAudioSettings();
+
+            SubscribeToVirtualAudioMixerEvents();
+            audioMixer = Resources.Load<AudioMixer>("AudioMixer");
+        }
+
+        ~Settings() {
```
BrianAmadori, line 62:
```suggestion
        public void Dispose() {
```
Please use the `Dispose()` pattern instead of finalizers.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Settings/Settings.cs
```diff
@@ -89,14 +104,34 @@ private void LoadGeneralSettings()
             }
         }
 
+        private void LoadAudioSettings() {
```
BrianAmadori, line 107:
```suggestion
        private void LoadAudioSettings() 
        {
```
Please put line breaks before the braces to conform to our general style in the codebase.

BrianAmadori (changes requested):
Hey Mordi, I'm glad to review your contributions once more. Please drop me a line on slack if you have any doubts on the requests.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Settings/Settings.cs
```diff
@@ -89,14 +104,34 @@ private void LoadGeneralSettings()
...
+            {
+                try
+                {
+                    object currentSetting = currentAudioSettings;
+                    JsonUtility.FromJsonOverwrite(PlayerPrefsUtils.GetString(AUDIO_SETTINGS_KEY), currentSetting);
+                    currentAudioSettings = (SettingsData.AudioSettings)currentSetting;
+                }
+                catch (Exception e)
+                {
+                    Debug.Log(e.Message);
```
sandrade-dcl, line 120:
```suggestion
                    Debug.LogError(e.Message);
```

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Settings/SettingsControllers/Tests/SettingsControlsShould.cs
```diff
@@ -263,22 +263,6 @@ public void ChangeMouseSensivityCorrectly()
...
-
-            // Act
-            bool newValue = true;
-            settingController.UpdateSetting(newValue);
-
-            // Assert
-            Assert.AreEqual(newValue, settingController.GetStoredValue(), "muteSound stored value mismatch");
-            Assert.AreEqual(newValue ? 1f : 0f, AudioListener.volume, "sfxVolume value mismatch");
-        }
-
```
sandrade-dcl, line 281: Could you create the corresponding test for each new audio setting control you have added? (`ChangeAvatarSFXVolumeCorrectly`, `ChangeChatSFXToggleCorrectly`, `ChangeMasterVolumeCorrectly`, `ChangeMusicVolumeCorrectly`, `ChangeSceneSFXVolumeCorrectly`, `ChangeUISFXVolumeCorrectly`).

sandrade-dcl (changes requested):
The code looks fine in general, good work! In addition to Brian's requests, I have added a few new ones. Let me know if you have any doubt, thanks!

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
Landeplage: Split it into temporary variables.
image

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Settings/Settings.cs
Landeplage: Nice catch. I also changed the existing exceptions for QualitySettings and GeneralSettings, which follow the same pattern.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Settings/SettingsControllers/Tests/SettingsControlsShould.cs
Landeplage: Added tests for all the controls. Currently, they only test their internal SettingsController value. Should they also be testing the final volume output? For example, some of them change the volume of AudioMixer groups - so should we test the value against these?

sandrade-dcl: It's ok just like you did it, thanks! :)

sandrade-dcl (approved):

BrianAmadori (approved):
I'm leaving a technical approve, but I am not entirely convinced by some of the settings.

- Music volume slider doesn't change the genesis plaza music volume. I think we should remove this setting or change the wording--sorry for coming up again with this, but it actually irked me that this didn't work on the very first starting position of the decentraland experience.
- Why we have "Chat sounds" as a toggle, but "UI volume" as another slider? Why Chat sounds don't have their own volume or aren't considered as part of the UI?

olavra:
@BrianAmadori we were considering extending the SDK so audio sources or clips can be defined as "music" or "sfx" so it would be included in the right channel for the mixer. Right now, all scene sounds are part of the world sound slider. Genesis plaza music is not part of the core game (like the tutorial, the builder or the backpack) and need to be considered as any other scene. Do you have any other approach?
Related to chat sounds, it was a direct request of many people to be able to disable those sounds specifically as it was very annoying during live events, so it's independent of the UI volume.

BrianAmadori:
@olavra Maybe we can use heuristics, and define as "background music" just any looping sound? As I don't think our users really will grasp the concept of "music that's part of decentraland UI". Conceptually, genesis plaza is our scene, so the mental model is "the music of our main scene isn't part of decentraland UI music, and its our music, but because it is not bundled with the Unity package it's on another slider"... pretty complex.

olavra:
Im ok about considering other options for this. We can merge this PR and work on that in a separate issue. There is a lot of corner cases that we need to consider. ie: genesis plaza has 2 types of music, one on top of the clouds and another streming audio in the bar from the jukebox. Assuming loop audiosorces or streaming audio are music feels a bit odd, I think we should go with the option on the SDK and let the creators decide.

PR #872, opened 2021-07-29
feat: disable SSAO for far-away avatars

### LINKED ISSUE #859: Disable SSAO for far-away avatars
Also we are applying the SSAO effect on avatars that are too far away to appreciate the effect, similar to the performance issue we had with shadows on very far away objects.

- [ ] Toggle each avatar SSAO effect based on distance (off when they are far away)

**Testing with lots of avatar-bots** -> https://www.notion.so/decentraland/Performance-Sync-36542a31bd714de19eaeb0c53510e97b?p=d800d1361f1d41a0844bb385068d78d7&showMoveTo=true

**Measuring performance score to compare changes** -> https://www.notion.so/decentraland/Performance-Sync-36542a31bd714de19eaeb0c53510e97b?p=550b7d14369a425ba1d8a40b5785716a&showMoveTo=true

AjimenezDCL:
Fixes #859 

Testing this feature is easier in UnityEditor. Fix the editor view to another player (you can open an incognito tab to sign-in as guest) and change the distance between yourself and the other player to check that the SSAO is being disabled by the distance.

Test coverage is not taking down here due to the fact that our current AvatarShape test suite is disabled and on the way of being refactored.


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
```diff
@@ -47,12 +47,12 @@ protected override void OnEnable()
...
-            if (disableFacialFeatureRoutine != null)
+            if (checkDistanceRoutine != null)
             {
-                StopCoroutine(disableFacialFeatureRoutine);
-                disableFacialFeatureRoutine = null;
+                StopCoroutine(checkDistanceRoutine);
+                checkDistanceRoutine = null;
             }
-            disableFacialFeatureRoutine = StartCoroutine(SetFacialFeaturesVisibleRoutine());
+            checkDistanceRoutine = StartCoroutine(CheckDistanceToPlayerRoutine());
```
BrianAmadori, line 55: Please let's put this logic inside a specialized class, and try to contain it in `AvatarRenderer`. `AvatarShape` shouldn't be the place for putting performance related plugins. Yes, `AvatarRenderer` is in emergency mode, but even so its a better place to put this into.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
```diff
@@ -229,14 +229,16 @@ public override void Cleanup()
...
-        private IEnumerator SetFacialFeaturesVisibleRoutine()
+        private IEnumerator CheckDistanceToPlayerRoutine()
         {
             while (true)
             {
                 yield return WaitForSecondsCache.Get(DISABLE_FACIAL_FEATURES_DISTANCE_DELAY);
                 Vector3 position = lastAvatarPosition ?? (entity.gameObject.transform.position + CommonScriptableObjects.worldOffset);
                 float distanceToPlayer = Vector3.Distance(CommonScriptableObjects.playerWorldPosition, position);
-                avatarRenderer.SetFacialFeaturesVisible(distanceToPlayer <= DISABLE_FACIAL_FEATURES_DISTANCE);
+                bool isNear = distanceToPlayer <= DISABLE_FACIAL_FEATURES_DISTANCE;
```
BrianAmadori, line 239: Here we should check if the avatar is loaded to avoid weirdness.

unity-renderer/Assets/Rendering/Shaders/Toon/Compiled/DCLSimpleLighting.hlsl
```diff
@@ -32,11 +32,13 @@ half4 DCL_SimpleFragmentPBR(InputData inputData, SurfaceData surfaceData)
 #endif
 
     #if defined(_SCREEN_SPACE_OCCLUSION)
+    #if !defined(_SSAO_OFF)
```
BrianAmadori, line 35:
```suggestion
    #if !defined(_SSAO_ON)
```
Can please invert the if to avoid double negatives? It makes code hard to maintain.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
```diff
@@ -44,6 +44,7 @@ public enum VisualCue
         private Coroutine loadCoroutine;
         private List<string> wearablesInUse = new List<string>();
         private bool facialFeaturesVisible = true;
+        private bool ssaoEnabled = true;
```
BrianAmadori, line 47: Be wary that sometimes `AvatarRenderer` is inside a pooled object. This means we have to ensure the values are true if this object is pooled. This may be a good opportunity to track those booleans in a specialized class that manages those optimizations.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
```diff
@@ -520,6 +523,18 @@ public void SetFacialFeaturesVisible(bool visible)
...
+        public void SetSSAOEnabled(bool newEnabled)
+        {
+            if (newEnabled == ssaoEnabled )
+                return;
+            ssaoEnabled = newEnabled;
+            if (bodyShapeController == null || !bodyShapeController.isReady)
+                return;
+            bodyShapeController.SetSSAOEnabled(ssaoEnabled);
+            foreach (WearableController wearableController in wearableControllers.Values)
+                wearableController.SetSSAOEnabled(ssaoEnabled);
```
BrianAmadori, line 535:
```suggestion
            foreach (WearableController wearableController in wearableControllers.Values)
            {
                wearableController.SetSSAOEnabled(ssaoEnabled);
            }
```
Small style suggestion

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
```diff
@@ -520,6 +523,18 @@ public void SetFacialFeaturesVisible(bool visible)
             bodyShapeController.SetFacialFeaturesVisible(visible, true);
         }
 
+        public void SetSSAOEnabled(bool newEnabled)
+        {
+            if (newEnabled == ssaoEnabled )
```
BrianAmadori, line 528:
```suggestion
            if (newEnabled == ssaoEnabled)
```

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/WearableController.cs
```diff
@@ -202,4 +202,18 @@ public bool IsLoadedForBodyShape(string bodyShapeId)
 
         return wearable.data.representations.FirstOrDefault(x => x.bodyShapes.Contains(bodyShapeId))?.mainFile == lastMainFileLoaded;
     }
+
+    public void SetSSAOEnabled(bool ssaoEnabled)
```
BrianAmadori, line 206: Please remember to reset this value to `true` when the controller is cleaned up or we may get pooled wearables with this badly toggled.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
BrianAmadori: Not sure if related to this, but I just confirmed that facial features can appear as hidden in the backpack, I think this may be related to the pooling issue.

BrianAmadori (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
AjimenezDCL: This shouldn't go in `AvatarRenderer`. `AvatarRenderer` is used by the AvatarEditor and the own player for the 3rd person Camera.

This feature (and the disabling of the facial features) is just related to how we render other players. IMO the capability of enabling/disabling SSAO and Facial Features should be in the AvatarRenderer (as it is in the implementation), but the decision should be external (in the AvatarShape in this case).

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
AjimenezDCL: I dont agree with that. The AvatarShape just rely the actual enabling/disabling to the `AvatarRenderer`. 
If the avatar is not loaded yet we still have to call the AvatarRenderer so it has into account the state of the SSAO/FacialFeatures for the moment is ready.

unity-renderer/Assets/Rendering/Shaders/Toon/Compiled/DCLSimpleLighting.hlsl
AjimenezDCL: Sure, I followed the other directives such as  `_RECEIVE_SHADOWS_OFF`. I also think it's way less problematic as it is so by default the SSAO is enabled (because the materials does not contains the `_SSAO_OFF` keyword).

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
AjimenezDCL: I think it's 100% related to that. The status of FacialFeatures is set by the `AvatarShape` so it shouldn't affect the AvatarEditor or the own player representation. So it's for sure an issue with the pooling. I thought we solved that by defaulting the flags in the `Cleanup` but it might not be enough, gonna take a look.

Thanks! nice catch!

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/WearableController.cs
```diff
@@ -156,8 +156,9 @@ public void SetAnimatorBones(SkinnedMeshRenderer skinnedMeshRenderer)
         boneRetargetingDirty = false;
     }
 
-    public void CleanUp()
+    public virtual void CleanUp()
```
sandrade-dcl, line 159: Why you needed to make this function as virtual?

sandrade-dcl (approved):
LGTM!

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/WearableController.cs
AjimenezDCL: Override in `BodyShapeController`

BrianAmadori (approved):
Approving as we are going to tackle the concerns in a new PR.

PR #874, opened 2021-07-30
fix: scene boundaries unit test

### LINKED ISSUE #854: Test coverage: Scene Boundarie Checker
Remove the explicit tag for the Scene Boundaries Checker test

Suduck:
## What does this PR change?

Fixes #854

This PR removes the explicit test from the Scene Boundaries Checker 

## How to test the changes?

1. Open the project in Unity
2. Run the test of the SceneBoundaries Checker

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneBoundariesController/Tests/SBC_Asserts.cs
```diff
@@ -70,7 +70,7 @@ public static IEnumerator NFTShapeIsInvalidatedWhenStartingOutOfBounds(ParcelSce
 
             var componentModel = new NFTShape.Model()
             {
-                src = "ethereum://0x06012c8cf97BEaD5deAe237070F9587f8E7A266d/558536"
+                src = "local://0x06012c8cf97BEaD5deAe237070F9587f8E7A266d/" + TestAssetsUtils.GetAbsolutePath() + "/NFT/nftInfo.json"
```
sandrade-dcl, line 73: I noticed you added a new piece of logic in `NFTShapeLoaderController` -> `FetchNFTImage()` only to make tests using local NFTs files and here you are changing the current tests from real NFTs urls to local ones, right?. But then, won't we have tests that contemplate real cases of valid/invalids urls? I mean for not "local" nfts?

sandrade-dcl (approved):
I approve it but with a doubt that I would like to know. Thanks!

pravusjif (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneBoundariesController/Tests/SBC_Asserts.cs
AjimenezDCL: We should never test against external resources but I agree with you. To best way to solve this is to use mocking in the `NFTShapeController`.

I would go for something like a `INFTHelper` interface with a `FetchInfo` method. Since `NFTShapeController` is a `Monobehaviour` some trickery would be needed to inject the `INFTHelper`, the easiest way would be to just have it public and make the test override it.

Once we are able to mock the fetching of the NFT, we can just do whatever we want in the test: Force a failure, retrieve local asset as textures, etc..

We would also avoid having test code in production (the "local" protocol).

AjimenezDCL (changes requested):
Let's try to tackle down the concern raised by @sandrade-dcl with the `NFTShapeController` and the use of the "local" protocol.

If it gets too complicated we can discuss creating a card for this.

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneBoundariesController/Tests/SBC_Asserts.cs
Suduck: Okey, I will change it.

I just copied the way we test the GLTF, since we have a code in the GLTF loader that test against local assets

BrianAmadori: I agree with the general request. On addition, I'd like to add that `FileLoader` us an utility class that should be used exclusively by the GLTF importer, if we need something similar we should add that functionality to our web request controller. This way we keep the GLTF code decoupled from the rest of our repo as best as possible.

Suduck: I will rename the `FileLoader ` to `GLTFileLoader `so in the future we don't use this utility class for other use then

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTInfoFetcher.cs
```diff
@@ -0,0 +1,45 @@
...
+}
+
+public class NFTInfoFetcher : INFTInfoFetcher
+{
+    internal Coroutine fetchCoroutine;
+
+    public void Dispose()
+    {
+        if (fetchCoroutine != null)
+            ;
```
BrianAmadori, line 25: You created a style monster here 💥 
```suggestion
            
```

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/Tests/NFTInfoFectcherShould.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+
+    [Test]
+    public void NFTInfoFetcherFail()
+    {
+        //Arrange
+        NFTInfoFetcher fetcher = new NFTInfoFetcher();
+
+        //Act - Assert
+        fetcher.FetchNFTImage("testAddress", "testId", info =>  Assert.Fail(), Assert.Pass );
```
BrianAmadori, line 18: Don't forget to dispose the fetcher!

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTInfoFetcher.cs
```diff
@@ -0,0 +1,45 @@
...
+        CoroutineStarter.Stop(fetchCoroutine);
+        fetchCoroutine = null;
+    }
+
+    public void FetchNFTImage(string address, string id, Action<NFTInfo> OnSuccess, Action OnFail) { fetchCoroutine = CoroutineStarter.Start(FetchNFTImageCoroutin(address, id, OnSuccess, OnFail)); }
+
+    private IEnumerator FetchNFTImageCoroutin(string address, string id, Action<NFTInfo> OnSuccess, Action OnFail)
```
BrianAmadori, line 32:
```suggestion
    private IEnumerator FetchNFTImageCoroutine(string address, string id, Action<NFTInfo> OnSuccess, Action OnFail)
```
Please be more careful with the naming...

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTInfoFetcher.cs
```diff
@@ -0,0 +1,45 @@
...
+
+    public void Dispose()
+    {
+        if (fetchCoroutine != null)
+            ;
+        CoroutineStarter.Stop(fetchCoroutine);
+        fetchCoroutine = null;
+    }
+
+    public void FetchNFTImage(string address, string id, Action<NFTInfo> OnSuccess, Action OnFail) { fetchCoroutine = CoroutineStarter.Start(FetchNFTImageCoroutin(address, id, OnSuccess, OnFail)); }
```
BrianAmadori, line 30: Maybe we should stop the previous coroutine if it was not null?

Suduck: It shouldn't fetch the NFTImage twice since there is a controller for each NFT but it won't harm, so I will add it

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShapeLoaderController.cs
```diff
@@ -146,41 +150,55 @@ public void UpdateBackgroundColor(Color newColor)
...
-    IEnumerator FetchNFTImage()
+    private void FetchNFTImage()
     {
         if (spinner != null)
             spinner.SetActive(true);
+        nftInfoFetcher.FetchNFTImage(darURLProtocol, darURLRegistry, NFTInfoFeteched, NFTInfoFetchedFail);
+    }
 
-        NFTInfo nftInfo = new NFTInfo();
+    private void NFTInfoFeteched(NFTInfo nftInfo) { fetchNftImage = StartCoroutine(FetchNFTImageCoroutine(nftInfo)); }
```
AjimenezDCL, line 160: `NFTInfoFetched`

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShapeLoaderController.cs
```diff
@@ -55,12 +57,13 @@ public enum NoiseType
     int BASEMAP_SHADER_PROPERTY = Shader.PropertyToID("_BaseMap");
     int COLOR_SHADER_PROPERTY = Shader.PropertyToID("_BaseColor");
 
-    IPromiseLike_TextureAsset assetPromise = null;
     GifPlayer gifPlayer = null;
     NFTShapeHQImageHandler hqTextureHandler = null;
 
     bool isDestroyed = false;
 
+    private Coroutine fetchNftImage;
```
AjimenezDCL, line 65: Let's rename this to `fetchNftImageRoutine` or something clarifying this holds a coroutine.

AjimenezDCL (approved):
Left a couple of naming requestes, it looks good with the new mocking approach!

Issue #876, opened 2021-07-30
Memory leak in green blockers from loading scenes


olavra:
Profiling possible memory leaks happened while the player is moving along the world, I discovered something interesting related to the green blockers. It seems each time we change our position between scenes, the old green blockers that were cleaned are provoking important memory leaks keeping several materials in memory that are ccumulating indefinitely.

### How to reproduce
1. Modify the `SceneController` to avoid to process any scene messages related to visual objects. It is recommended in order to isolate the problem. You can put these 2 breaks temporally:
![image.png](https://images.zenhubusercontent.com/5eb2b8ca65c88f6e391ef16e/e0cedbd1-1bd3-4a8b-8116-47e3d3e6e4cb)

2. Load the explorer in the position (100,100) and open the Unity Profiler.

3. See the numer of material loaded.
![image.png](https://images.zenhubusercontent.com/5eb2b8ca65c88f6e391ef16e/e1fd57de-ac04-4741-9b0f-f8895f79c92f)

4. Teleport to position (130,100) and see the number of materials.
![image.png](https://images.zenhubusercontent.com/5eb2b8ca65c88f6e391ef16e/10667150-c872-4710-a2d9-f67f16e000bf)

5. Teleport to position (105,120) and see the number of materials.
![image.png](https://images.zenhubusercontent.com/5eb2b8ca65c88f6e391ef16e/b8ac729b-8666-49cf-82b2-79272fb56670)

6. Teleport back to the initial position (100,100) and see the number of materials.
![image.png](https://images.zenhubusercontent.com/5eb2b8ca65c88f6e391ef16e/b3c54b4a-542f-4e65-8d32-9897360a32d8)

7. Notice that the number of material is always increasing.

8. Go to the class `WorldBlockersController` and comment this line in the `InitializeWithDefaultDependencies()` function:
`Initialize(sceneHandler,blockerInstanceHandler);`. With that we will deactivate the world blockers feature.

9. Repeat the process and notice the number of material is not increasing between jumps.

PR #879, opened 2021-07-30
fix: TextMeshPro materials used by TextShape are not properly destroyed

### LINKED ISSUE #772: TextMeshPro materials used by TextShape are not properly destroyed
There were lots of mesh count related to TextMeshPro on unity's profiler.
We should analyze if the `TextShape` component is destroyed correctly

sandrade-dcl:
## What does this PR change?

Fix #772 

There were lots of mesh count related to TextMeshPro on unity's profiler.
We should analyze if the TextShape component is destroyed correctly

## Implemented solution
1. Each time a `TextShape` object is created, we store a copy of the shared material in a private variable.
2. Instead of using `text.fontMaterial...`, we use `text.fontSharedMaterial...` in order to not to provoke new copies of the material.
3. Once the `TextShape` object is destroyed, we destroy our stored material.

**NOTE:** Making different tests with the `TextShape` components loading isolated, I have not seen a huge memory saving (the most important issue that I've detected with the number of materials in the profiler has been telated to this other [ISSUE](https://github.com/decentraland/unity-renderer/issues/876)). Anyway, in order to avoid problems with the material handling, I think it is the best way to manage the cleaning of font materials.

## Test link
https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/TextShape-memory-leaks&ENV=org

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
@@ -62,8 +62,14 @@ namespace DCL.Components
         public TextMeshPro text;
         public RectTransform rectTransform;
         private Model cachedModel;
+        private Material cachedFontMaterial;
 
-        private void Awake() { model = new Model(); }
+        private void Awake()
+        {
+            model = new Model();
+            cachedFontMaterial = new Material(text.fontSharedMaterial);
+            text.fontSharedMaterial = cachedFontMaterial;
+        }
 
         public void Update()
         {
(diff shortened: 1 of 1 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
```diff
@@ -62,8 +62,14 @@ public class Model : BaseModel
         public TextMeshPro text;
         public RectTransform rectTransform;
         private Model cachedModel;
+        private Material cachedFontMaterial;
```
pbosio, line 65: I think probably a better name would be `fontMaterial` removing de `cached` part

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
```diff
@@ -62,8 +62,14 @@ public class Model : BaseModel
...
 
-        private void Awake() { model = new Model(); }
+        private void Awake()
+        {
+            model = new Model();
+            cachedFontMaterial = new Material(text.fontSharedMaterial);
+            text.fontSharedMaterial = cachedFontMaterial;
```
pbosio, line 71: Probably is cleaner this way:
```suggestion
            cachedFontMaterial = text.fontMaterial;
```
since `text.fontMaterial` is already creating a new material and assigning it to the `TextMeshPro`

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
```diff
@@ -130,24 +136,24 @@ public static IEnumerator ApplyModelChanges(IParcelScene scene, TMP_Text text, M
...
+                text.fontSharedMaterial.EnableKeyword("OUTLINE_ON");
                 text.outlineWidth = model.outlineWidth;
                 text.outlineColor = model.outlineColor;
             }
-            else if (text.fontMaterial.IsKeywordEnabled("OUTLINE_ON"))
+            else if (text.fontSharedMaterial.IsKeywordEnabled("OUTLINE_ON"))
             {
-                text.fontMaterial.DisableKeyword("OUTLINE_ON");
+                text.fontSharedMaterial.DisableKeyword("OUTLINE_ON");
             }
```
pbosio, line 157: If you are already holding an instance of the material it would be better to use that directly
i.e: 
```
cachedFontMaterial.EnableKeyword("UNDERLAY_ON");
cachedFontMaterial.SetColor("_UnderlayColor", model.shadowColor);
cachedFontMaterial.SetFloat("_UnderlaySoftness", model.shadowBlur);
```

pbosio (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
BrianAmadori: I'd prefer not relying on the `fontMaterial` awful copying mechanism. It's easier to understand as it is now.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
BrianAmadori: This can conflate with `text.fontMaterial` making the code harder to read.

BrianAmadori (approved):
LGTM

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
sandrade-dcl: We can't do it directly because the `ApplyModelChanges` function is static but not our `cachedFontMaterial` variable. We could pass the `cachedFontMaterial` variable through the function signature but, in order not to make the function more complex, I prefeer to continue using `text.fontSharedMaterial`.

PR #884, opened 2021-08-02
feat: fix issue with lands where you have "UpdateManager" permission

Suduck:
## What does this PR change?

This PR fixes an issue where if you have "UpdateManager" permissions on lands, it will fail.

Also, it solves a problem where if you have more than 5000 lands, it won't load

## How to test the changes?

In order to test the changes, you should have a UpdateManager permissions from any land. 

If you don't have that, you can just test that everything work as normally

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/release-bug
2. Enter builder in world by the project panel

## Our Code Review Standards


sandrade-dcl (approved):
LGTM, good work!

PR #892, opened 2021-08-02
chore: avatar lod impostor snapshots

pravusjif:
### Testing with bots 
1. enter https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:chore/AvatarLODImpostorSnapshots&ENV=zone&position=91,151&ENABLE_AVATAR_LODS
2. open browser console

### Testing with real users
1. enter https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:chore/AvatarLODImpostorSnapshots&ENV=org&position=-27,60&ENABLE_AVATAR_LODS

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarLOD/AvatarLODController/AvatarLODController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarLOD/AvatarLODController/AvatarLODController.cs
@@ -6,6 +6,9 @@ namespace DCL
     {
         private const bool ONLY_GENERIC_IMPOSTORS = false;
         private const string LOD_TEXTURE_SHADER_VAR = "_BaseMap";
+        private readonly int LOD_IMPOSTOR_LAYER = LayerMask.NameToLayer("CharacterPreview");
+
+        private static Camera snapshotCamera;
 
         // 2048x2048 atlas with 8 512x1024 snapshot-sprites
         private const int GENERIC_IMPOSTORS_ATLAS_COLUMNS = 4;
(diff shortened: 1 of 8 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarLOD/AvatarLODController/AvatarLODController.cs
```diff
@@ -49,17 +83,57 @@ private void ResetMeshUVs()
...
+            snapshotCamera.gameObject.SetActive(true);
+            snapshotCamera.transform.SetParent(Camera.main.transform);
+            snapshotCamera.transform.localPosition = Vector3.zero;
+            snapshotCamera.transform.forward = (impostorMeshRenderer.transform.position - snapshotCamera.transform.position).normalized;
+            snapshotCamera.transform.position = impostorMeshRenderer.transform.position + -snapshotCamera.transform.forward * 2f;
+
+            snapshotCamera.Render();
+            RenderTexture.active = snapshotRenderTexture;
+            snapshotTex.ReadPixels(new Rect(0, 0, snapshotRenderTexture.width, snapshotRenderTexture.height), 0, 0);
+            snapshotTex.Apply();
```
BrianAmadori, line 132: Oh, of course, this is what is killing the performance. We should find a way to directly render to the RT, and assign the RT to the LOD billboards materials directly. This optimization only has chances to work if we never create `Texture2D` objects dynamically, and of don't use the crazy expensive `ReadPixels` + `Apply` combo.

pravusjif: Right, I didn't know we could use the `RenderTexture` directly in the material and avoid the `Texture2D` update with `ReadPixels()` + `Apply()`.
I just updated the code to use 1 `RenderTexture` per avatar and no `Texture2D`, instead of 1 `Texture2D` per avatar and a shared `RenderTexture`.

Next I'll add the `Animation` component toggle, the shared `RenderTexture` atlas (without actual atlassing logic, just sharing the same renderTex) and after that we can join all our optimizations into an experimental branch to measure performance.

pravusjif:
Added animation toggling and shared renderTex (no atlas logic) that breaks the experience but will be used this way to test an integration of all the optimizations.

PR #898, opened 2021-08-03
chore: exclude tests, ABConverter and WSSController from code coverage

### LINKED ISSUE #887: Discard test assemblies from coverage calculation
Everything that ends with `*Test`, `*Should` and everything related with asset bundles

kuruk-mm:
Resolves #887 

## What does this PR change?

Remove tests, ABConverter, and WSSController from code coverage

## How to test the changes?

1. Look at the `github-action` comment bot below this PR, and open the code coverage.
2. Compare with https://renderer-artifacts.decentraland.org/branch-coverage/master if the tests, ABConverter and WSSController coverage has gone.

## Our Code Review Standards


BrianAmadori (commented):
When opening the coverage I still see several assemblies ending with `Tests`. I shouldn't see them there right?

kuruk-mm:
> When opening the coverage I still see several assemblies ending with `Tests`. I shouldn't see them there right?

You're right, so I changed it. Now every path that has `Test` or `Should` will be ignored in the Code Coverage.

We must take care of avoiding using `Test` or `Should` in code that we want to be coverable. I reviewed all cases, and we don't use `Test` or `Should` in other places.

Do you think it's the right approach?

BrianAmadori (approved):

PR #903, opened 2021-08-04
fix: load empty lands on the builder panel 

### LINKED ISSUE #897: BIW: An account with lots of lands doesn't load the Builder In World Panel
If you use an account that has lots of lands, the builder in world panel won't load and will say that you don't have lands or scenes even if you do

Suduck:
## What does this PR change?

This PR fixes the problem where if you try to load the scene entity from an empty parcel where a deploy never occurs, the builder project panel never load anything.

Fixes #897 

## How to test the changes?

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/lots-of-lands
2. Enter with an account that has an empty land where a deploy has never occurs
3. Check that the builder panel loads correctly

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/ServiceProviders/TheGraph/TheGraphQueries.cs
```diff
@@ -1,25 +1,28 @@
...
+    public static string getLandQuery => landQuery.Replace("[AMOUNT]", MAX_LAND_QUERY_AMOUNT.ToString());
+
+    private static readonly string landQuery = @"
   query Land($address: Bytes) {
-    ownerParcels: parcels(first: 500, where: { estate: null, owner: $address }) {
+    ownerParcels: parcels(first: [AMOUNT], where: { estate: null, owner: $address }) {
       ...parcelFields
     }
-    ownerEstates: estates(first: 500, where: { owner: $address }) {
+    ownerEstates: estates(first: [AMOUNT], where: { owner: $address }) {
```
pbosio, line 11: Could we use a variable here like `$address` for amount instead of using a string replace?

sandrade-dcl: +1

sandrade-dcl (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/ServiceProviders/TheGraph/TheGraphQueries.cs
Suduck: Sure, but the problem is that $address is replaced in the TheGraph part of the query ( not on our side) so if in the future someone adds a variable $amount for the query in the TheGraph part it will collide!

pbosio: We are actually sending the query and it's variables. It's not something happening in black box on TheGraph side.
Here we are setting `$address` variable for example
https://github.com/decentraland/unity-renderer/blob/master/unity-renderer/Assets/Scripts/MainScripts/DCL/ServiceProviders/TheGraph/TheGraph.cs#L87

Suduck: I think that what you mean is to replace the value in thegraph, I have change it!

Suduck: Didn't see the comment but I realize after a while. Yep, I have changed it!

pbosio: Great :)
Don't forget you must also change the query args
`query Land($address: Bytes)`  to something like `query Land($address: Bytes, $amount: Int)`
and also maybe create a new variable class since `AddressVariable` is used for other stuff that doesn't require `$amount`

Suduck: I just reverted the changes,  to let them as they were before my changes, their API is no accepting an array of variables and the query needs to be changed since the fragment also needs the variable. I think that invest more hours into this to change that is not worth it

PR #909, opened 2021-08-04
fix: Improve feedback for NFTs not loaded

### LINKED ISSUE #657: Manage feedback for NFTs not loaded
## OLD ISSUE (discarded at the moment due to incompatibilities on the OpenSea API)
According to someone from the community, there are some NFTs that used to load in picture frames that no longer do
for example with: `ethereum://0x272f20cb476ef250e6e8478740dc62416040bdf0/17`

>Hi everybody, I had a couple of nfts in matic showing as NFTShape objects in my scene, they were displaying correctly until a couple of days ago. Now instead of the images I see a circle progress that never loads. Has the way of displaying nfts in matic changed? This address was working: `ethereum://0xbccaa7acb552a2c7eb27c7eb77c2cc99580735b9/5465` It has changed to another address, type `matic://0xbccaa7acb552a2c7eb27c7eb77c2cc99580735b9/5465` or something similar? Or is it no longer supported?
>

Apparently Opensea changed the path to include `/matic` in the middle, maybe they did the same with their API?

If we can’t know if a token is on matic or not, maybe we could try requesting again the asset (if the firsrt request fails) and use the new `/matic` in the request?

Another example: https://opensea.io/assets/matic/0xbccaa7acb552a2c7eb27c7eb77c2cc99580735b9/5465

## NEW ISSUE
Due to the incompatibility on the OpenSea API side, while the API is updated to the V2 where we will be able to support the loading of Polygon assets, we are going to manage the feedback that the user/developer will receive when an asset can't be loaded in the NFT frame and details dialog.

For both components we will show an image that represents when an assets couldn't be loaded (it has to be designed by Romi), instead of let the loading spinner remaining visible forever. This will give more visual information to the users about what happened.

We will also update the messages that we receive in the browser console when an asset can't be loaded, in order to give more information to the scenes developers.

### LINKED ISSUE #676: Update picture frame NFT loading animation

sandrade-dcl:
## What does this PR change?
Fix #657 
Fix #676 

### Show visual feedback to the users when a NFT can't be loaded
For both components (frame and dialog) we have modified its flow to show an image that represents when an assets couldn't be loaded instead of let the loading spinner remaining visible forever. This will give more visual information to the users about what happened.

- When a NFT frame can't load the asset:

image

- When a NFT dialog can't load the asset from the api:

image

- When a NFT dialog can load the asset from the api but can't load the image:

image

### Send detailed error messages to the browser console when a NFT can't be loaded
We will also update the messages that we receive in the browser console when an asset can't be loaded, in order to give more information to the scenes developers and users. In addition, while the OpenSea api becomes compatible with Polygon assets fetching, we have added the next warning when a NFT can't be loaded: `Warning: OpenSea API does not support fetching Polygon assets`.

### Update loading animations
We have replaced the spinner animations with new ones.

- For NFT frame:

c2dac99b-339a-4a55-b659-4206e007895f

**NOTE**: In order to avoid to have an animator running for each NFT frame loaded in the scene, this animation has been made through a shader that is using a shared material. This shader reads a sprite sheet with the full animation and render it in the proper way. In this way all the NFT frames will share the same animation during its loading stage.

- For NFT dialog HUD:

3bd45b6e-af1c-438c-aebf-536ea693c293

**NOTE**: In order to be more performant, we have created a Sprite Atlas with all the sprites that are part of this animation.

## How to test the changes?

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:feat/spinnermulticolor&ENV=org
2. Find a place where there are NFTs frames and NFT dialogs, or create a test scene with NFT frames and dialogs and test it in preview mode.

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShapeLoaderController.cs
```diff
@@ -159,13 +167,16 @@ IEnumerator FetchNFTImage()
             },
             (error) =>
             {
-                Debug.LogError($"Couldn't fetch NFT: '{darURLRegistry}/{darURLAsset}' {error}");
+                Debug.LogError($"{COULD_NOT_FETCH_NFT_FROM_API.Replace("{darURLRegistry}", darURLRegistry).Replace("{darURLAsset}", darURLAsset)} {error}. {DOES_NOT_SUPPORT_POLYGON}");
```
pbosio, line 170: Could we change al those `Replace` for the errors? It doesn't make much sense creating a constant like `COULD_NOT_FETCH_NFT_FROM_API` while at the same time using a hardcoded string in the line like `"{darURLRegistry}"`. Using `Replace` here is like too much for something so little 🤣
Is probably better just to use `string.Format`
i.e:
```
COULD_NOT_FETCH_NFT_FROM_API = "Couldn't fetch NFT: '{0}/{1}'.";
 Debug.LogError(string.Format(COULD_NOT_FETCH_NFT_FROM_API, darURLRegistry, darURLAsset));
```

pbosio (commented):
95% of the time the loading animation is not visible due to the culling.
Can that be added the ignore layer?

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShapeLoaderController.cs
sandrade-dcl: You're totally right, I have changed all these `.Replace()` by `string.Format()`. Commited in https://github.com/decentraland/unity-renderer/pull/909/commits/662d0ab4ad8be0b33b488369659f9d643a6e651b. Thanks for the feedback, Pato!

sandrade-dcl:
> 95% of the time the loading animation is not visible due to the culling.
> Can that be added the ignore layer?

@pbosio "95% of the time" is a lot of times xD Could you please tell me in what specific situations you are not seeing the loading? I'm testing the build and I don't see anything weird:

image
image

The old loading was also not added as ignored layer in the culling and I don't know why we should add this one.

Thanks!

pbosio:
> > 95% of the time the loading animation is not visible due to the culling.
> > Can that be added the ignore layer?
> 
> @pbosio "95% of the time" is a lot of times xD Could you please tell me in what specific situations you are not seeing the loading? I'm testing the build and I don't see anything weird:
> 
> image
> image
> 
> The old loading was also not added as ignored layer in the culling and I don't know why we should add this one.
> 
> Thanks!

Do you have culling enabled, right? :P
I've tested it at SuperRare museum (55,94)  and I couldn't see almost any loading animation :(

pbosio (approved):

sandrade-dcl:
> Do you have culling enabled, right? :P
> I've tested it at SuperRare museum (55,94) and I couldn't see almost any loading animation :(

You were right, I couldn't reproduce the bug in other scenes but I've been able to do it in `55,94` 👍 Fixed in https://github.com/decentraland/unity-renderer/pull/909/commits/e706465a840a1eb39e940a81506a12579640d98b. Thanks!

Suduck (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/Shaders/FlibBook.shader
```diff
@@ -0,0 +1,94 @@
+// https://unitycoder.com/blog/2018/11/30/sprite-sheet-flip-book-shader/
+
+Shader "UnityLibrary/Sprites/FlipBook (Cutout)"
```
AjimenezDCL, line 3: The file is named `FlibBook` we should rename to `FlipBook`

AjimenezDCL (approved):
Small request, approving regardless

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/Shaders/FlibBook.shader
sandrade-dcl: Ups, good catch! Changed in https://github.com/decentraland/unity-renderer/pull/909/commits/e156b2009b323aad335c2c628712dd6757913e2d. Thanks!

PR #914, opened 2021-08-04
chore: remove test requirement to publish build

BrianAmadori:
We need to be able to test builds regardless of the `playmode-tests` job.


BrianAmadori:
Closing because the `publish-renderer` step is not the step that publish the PR testing artifacts. That step is `build-js-interface` and it already runs regardless of test run outcome.

PR #915, opened 2021-08-04
feat: increased draw distance and updated quality presets, #880

### LINKED ISSUE #880: Increment settings max draw distance and adjust profiles
Right now, the max draw distance is capped to 100. This is too short for most of the popular big scenes. (WonderMine, Genesis Plaza). This tightness is unjustified now that we have the object detail culling, as small objects will not get rendered anyways.

We should increment the max cap to 500 and adjust the presets to be:

* Basic = 70 or 100
* Medium = 100 or 150
* High = 250
* Ultra = 500

image

This should fix #666 and could be done along #803

Kinerius:
Closes #880 

Also increased scene load max radius from 4 to 6

### DIFF
--- a/unity-renderer/Assets/Resources/ScriptableObjects/QualitySettingsData.asset
+++ b/unity-renderer/Assets/Resources/ScriptableObjects/QualitySettingsData.asset
@@ -21,7 +21,7 @@ MonoBehaviour:
     shadows: 0
     softShadows: 0
     shadowResolution: 256
-    cameraDrawDistance: 40
+    cameraDrawDistance: 100
     bloom: 0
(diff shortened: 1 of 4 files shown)


BrianAmadori:
Test link: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:feat/increased-draw-distance&ENV=org

BrianAmadori (approved):
Tested it and the WonderMine zone looks awesome. We have to be wary with the loading radius, but as only the max value has been changed I think we can live with it, and in fact, high end users will appreciate it.

PR #924, opened 2021-08-05
fix: Moving platforms now rotate character avatar and first person camera

### LINKED ISSUE #477: Fix moving platforms rotation affecting players
The moving platforms don't rotate the character, they should.
* First person: should rotate the camera
* Third person: should rotate the character

This can be probably fixed by adding the moving platform rotation to the character similar as we add the moving platform position (In `DCLCharacterController` we use `TransformPoint()` and `InverseTRansformPoint()` utility methods to "simulate the parented movement" without parenting the character to the platforms to avoid deformations based on the moving platforms scale. Maybe there is something similar to simulate parented rotations)

**Bug testing A**
This can be easily tested on the moving platforms at the test scene at ZONE https://play.decentraland.zone/?position=59%2C-69

Sometimes the rotating platform in there doesn't work (that deployed scene is super old and may have issues)

**Bug Testing B**
Using the genesis plaza flying tour cart:

https://images.zenhubusercontent.com/337227404/b6077412-1772-4127-9ecc-3c64e728d962/roating_bug_b.mov

https://images.zenhubusercontent.com/337227404/3a8434ea-b4f0-4517-b586-e7aab272db5b/rotating_bug_a.mov

Kinerius:
Fixes #477 

How to test: 

- Go to Genesis Plaza flying tour carts or any moving platform that rotates
- On first person, the camera should rotate as the platform rotates
- On third person, the camera should stay still and the character should rotate as the platform rotates

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
@@ -2,9 +2,7 @@ using DCL;
 using DCL.Configuration;
 using DCL.Helpers;
 using UnityEngine;
-using System.Collections;
 using Cinemachine;
-using UnityEngine.SceneManagement;
 
 public class DCLCharacterController : MonoBehaviour
 {
@@ -44,6 +42,10 @@ public class DCLCharacterController : MonoBehaviour
     float lastMovementReportTime;
     float originalGravity;
     Vector3 lastLocalGroundPosition;
+    
+    Vector3 lastCharacterRotation;
+    Vector3 lastGlobalCharacterRotation;
+    
     Vector3 velocity = Vector3.zero;
 
     public bool isWalking { get; private set; } = false;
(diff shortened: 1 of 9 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
```diff
@@ -346,11 +348,21 @@ internal void LateUpdate()
 
         if (isOnMovingPlatform)
         {
-            lastLocalGroundPosition = groundTransform.InverseTransformPoint(transform.position);
+            SaveLateUpdateGroundTransforms();
         }
 
         OnUpdateFinish?.Invoke(deltaTime);
     }
+    private void SaveLateUpdateGroundTransforms()
```
BrianAmadori, line 356: I think we can add a test to ensure the character is rotated correctly? We already have tests in place to check moving platforms. Look for `CharacterSupportsRotatingPlatforms`. I think the test was made explicit but not sure why (maybe because of this bug).

Kinerius: Added an assert to verify the rotation of the character after being rotated by the platform on the last commit, alas, this was working even without my changes, so an external system was causing the bug.

sandrade-dcl (approved):
LGTM! Reviewed and tested, good work!

PR #925, opened 2021-08-05
fix: master promise cleaned or unloaded while waiting promises still exist

### LINKED ISSUE #904: When gltf loading is stressed, there's the possibility of massive redundant loading of assets
This bug was reproduced in a GLTF stress loading scenario, this means that a lot of objects were being loaded and unloaded concurrently.

In this scenario, a race condition can occur, this race condition is pretty complex and can be described as:

- A shape X starts loading for id 1
- Shapes A-B-C start loading for id 1
- As id 1 wasn't finished loading, `AssetPromiseKeeper` will wait until shape X finishes loading until A, B and C can finish.
- Shape X finishes loading. All is right. 
- Now, a deferred process will start and resolve the A, B and C promises. This process is time-sliced, this means that can skip frames in-between.
- In some of those frame skips, **shape X can be unloaded**.
- When this happens, the remaining shapes to be resolved (let's say B and C) will be loaded **completely from scratch**.
- In populated zones of the world, there's not only 3 depedencies, there can be hundreds of them.

pbosio:
fix #904 
wait before all waiting promises for the same asset are resolved before cleaning up the master promise.

fixed issue where an asset could be loaded twice if:
* master promise is forgetted while processing promises that were waiting for it
* a promise for the same asset is kept just after the waiting promises started to be processed


BrianAmadori (approved):
The fix and tests look very clean, LGTM!

sandrade-dcl (approved):

pravusjif (approved):

PR #929, opened 2021-08-06
feat: correct biw flow with the shortcut

### LINKED ISSUE #896: BIW: If you try to use the shorcut in a land that you own and you didn't open
If you didn't open the builder in world panel and try to use the shortcut in a land that you own, it will say that you don't have permission the first time because we still didn't query the lands to the graph

Suduck:
## What does this PR change?

Fixes #896

Entering in builder in world by the shortcut is not working correctly. This PR changes that adding a message and trying to edit the land after the permissions of the lands has been checked if the shortcut has been pressed

## How to test the changes?

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/biw-shorcut-flow
2. Try to enter in your land with the shortcut "K" without opening the builder in world panel
3. Restart the session and try to enter in a land where you don't have permissions

## Our Code Review Standards


sandrade-dcl:
Testing this change in ORG, just after the notifications, the app is crashing:
image
image

Other thing: maybe we should configure the first notification of "Checking..." as a warning without button? I don't know, it would be fine to discuss it with product.

sandrade-dcl (approved):

PR #936, opened 2021-08-06
fix: Avoid asking for owned wearables when the explorer loads

### LINKED ISSUE #836: Avoid asking for owned wearables when the explorer loads
Right now, Unity is asking for the user's owned wearables when the profile loads for the first time. This happens during the initial load, so it's every time, for every user.

This call is the most expensive call there is, since it has to go to query different subgraphs to learn all owned wearables (in both L1 and L2).

We should try to make this call only when necessary, and that is when the Avatar editor is opened. Users that don't open the avatar editor don't need to learn all the wearables they own.

sandrade-dcl:
## What does this PR change?

Fix #836 

Right now, Unity is asking for the user's owned wearables when the profile loads for the first time. This happens during the initial load, so it's every time, for every user.

This call is the most expensive call there is, since it has to go to query different subgraphs to learn all owned wearables (in both L1 and L2).

We should try to make this call only when necessary, and that is when the Avatar editor is opened. Users that don't open the avatar editor don't need to learn all the wearables they own.

## How to test the changes?

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/avoid-asking-for-owned-wearables-when-the-explorer-loads&ENV=org
2. Open the Backpack and check that all your owned wearables are loaded.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/AvatarEditorHUDController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/AvatarEditorHUDController.cs
@@ -86,7 +86,6 @@ public class AvatarEditorHUDController : IHUD
 
     private void LoadUserProfile(UserProfile userProfile)
     {
-        LoadOwnedWereables(userProfile);
         LoadUserProfile(userProfile, false);
         QueryNftCollections(userProfile.userId);
     }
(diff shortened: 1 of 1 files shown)


pravusjif (changes requested):
There's a problem when entering as guest and doing the onboarding creating an avatar(I checked that this doesn't happen with master branch build)

I think that "load retry" item shouldn't appear there.

Screen Shot 2021-08-09 at 14 34 23

sandrade-dcl:
> There's a problem when entering as guest and doing the onboarding creating an avatar(I checked that this doesn't happen with master branch build)
> 
> I think that "load retry" item shouldn't appear there.
> 

@pravusjif I forgot to left deactivated a game object in the `ItemSelection` prefab. This was provoking the issue that you have detected. Already fixed in https://github.com/decentraland/unity-renderer/pull/936/commits/1b699da85e509cc4a12cb848ef6d9cecfe0c5e5d. Good catch! Thanks!

Suduck (approved):
Tested and worked!

PR #946, opened 2021-08-09
Fix: pools instantiating additional objects when renderer was turned off

### LINKED ISSUE #876: Memory leak in green blockers from loading scenes

Kinerius:
Fixes #876 
When teleporting, the renderer is being turned off and every pool instantiates new objects instead of using the unused ones.

I'm not sure if this is the best fix for this issue

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/PoolManager/Pool.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/PoolManager/Pool.cs
@@ -39,6 +39,7 @@ namespace DCL
         private readonly LinkedList<PoolableObject> usedObjects = new LinkedList<PoolableObject>();
 
         private int maxPrewarmCount = 0;
+        private bool isInitialized;
 
         public float lastGetTime { get; private set; }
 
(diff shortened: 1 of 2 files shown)


sandrade-dcl (approved):
I have tested your change and confirmed that the memory leaks related to the blockers no longer happen, great! I'm not completely sure if yur change in the `Pool` -> `Get()` could have any collateral effect on other thing of our app, but for now I left my approve and let wait for any others reviews to confirm 👍 Good work!

PR #949, opened 2021-08-10
fix: SendWebRequest retries system is blocking the async operation

### LINKED ISSUE #910: WebRequestController.SendWebRequest<T> retry & return value
https://github.com/decentraland/unity-renderer/blob/master/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestController.cs#L102-L170

`WebRequestController.SendWebRequest<T> `is returning `WebRequestAsyncOperation resultOp` when called
but `resultOp` is assigned a new instance for every retry.

In the following scenario:

```
var op = WebRequestController.SendWebRequest<T>(...)
yield return op
```
if there was a retry, `op` is no longer valid for the operation and will keep waiting forever

sandrade-dcl:
## What does this PR change?

Fixes #910

`WebRequestController.SendWebRequest<T>` is returning `WebRequestAsyncOperation resultOp` when called
but `resultOp` is assigned with a new instance for every retry.

In the following scenario:

```
var op = WebRequestController.SendWebRequest<T>(...)
yield return op
```
If there was a retry, `op` is no longer valid for the operation and will keep waiting forever.

## Test link
https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:fix/SendWebRequest-retries-system-is-blocking-the-async-operation&ENV=org

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/Interfaces/WebRequestAsyncOperation.cs
```diff
@@ -49,6 +49,8 @@ public WebRequestAsyncOperation(UnityWebRequest webRequest)
             disposeOnCompleted = true;
         }
 
+        public void SetNewWebRequest(UnityWebRequest newRequest) { webRequest = newRequest; }
```
pbosio, line 52: could we set this as `internal`?
If we keep it `public` it must handle internally the scenario of replacing the `webRequest` like `WebRequestController` does when calling `resultOp.Dispose()` before retrying

pbosio (approved):
LGTM! just a small observation:

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/Interfaces/WebRequestAsyncOperation.cs
sandrade-dcl: You're right! done in https://github.com/decentraland/unity-renderer/pull/949/commits/632c6957d2cbfa7703abe3fba3c0b150911bce68. Thanks!

AjimenezDCL (approved):
We should have a test covering this, but good job

pravusjif (approved):

PR #969, opened 2021-08-11
fix: NFTImage material had wrong material

### LINKED ISSUE #782: NFTShape/PictureFrame has problems with overlapped transparencies
https://play.decentraland.org/?position=55%2C94

Kinerius:
## What does this PR change?

Fix #782 

## How to test the changes?

See issue #782 

## Our Code Review Standards


BrianAmadori (commented):
QQ about the approach. We can't make the material used by NFTs opaque or alpha-test? Because as this change could fix this specific case, in general terms transparency-transparency order issues can always arise, regardless of the shader used.

Kinerius:
> QQ about the approach. We can't make the material used by NFTs opaque or alpha-test? Because as this change could fix this specific case, in general terms transparency-transparency order issues can always arise, regardless of the shader used.

I was also wondering why that material isn't opaque, if we don't really need transparency on NFT images I don't see why we shouldn't change it.

pravusjif (approved):

PR #971, opened 2021-08-11
fix: avatar lod transition performance

pravusjif:
* We are constantly telling every avatar which state they should be in
* We were already checking if the new state values correspond to the "current transition" values, the problem here is that we also checked if the coroutine was not null to escape, removing this check (and leaving the `Mathf.Approximately()` checks) **already fixes the performance issue**:
Screen Shot 2021-08-11 at 15 53 12
(`AvatarsLODController.UpdateAllLODs()` average CPU usage with 200 bots: ~1.55ms)

* However, the 2 `Mathf.Approximately()` checks have a bit of performance cost as well, even though it's not much, the `AvatarsLODController.UpdateAllLODs()` ends up consuming 50% more CPU. By introducing a "state" `enum` and using that  to escape before calling the `StartTransition()` method we can lower the `AvatarsLODController.UpdateAllLODs()` average CPU usage with 200 bots to ~1ms
Screen Shot 2021-08-11 at 15 57 23

### Testing with bots 
1. enter https://play.decentraland.zone/index.html?renderer-branch=fix/AvatarLODTransitionPerformance&ENV=zone&position=91,151&ENABLE_AVATAR_LODS
2. open browser console

### Testing with real users
1. enter https://play.decentraland.zone/index.html?renderer-branch=fix/AvatarLODTransitionPerformance&ENV=org&position=-27,60&ENABLE_AVATAR_LODS

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarLOD/AvatarLODController/AvatarLODController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarLOD/AvatarLODController/AvatarLODController.cs
@@ -6,6 +6,13 @@ namespace DCL
 {
     public class AvatarLODController : IDisposable
     {
+        private enum LODState
+        {
+            LOD0,
+            LOD1,
+            LOD2
+        }
+
         private const float TRANSITION_DURATION = 0.25f;
         private Player player;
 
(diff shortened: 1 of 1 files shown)


AjimenezDCL:
The bug with the routine was fixed already here: https://github.com/decentraland/unity-renderer/pull/955/files

Do we need the rest of this PR?

Suduck (approved):
LGTM! For event better performance, using integers instead of enums are even more performance  although I don't know if in this case it will make any difference

pravusjif:
I'm closing the PR as this issue was already fixed by @AjimenezDCL in [this pr](https://github.com/decentraland/unity-renderer/pull/955/files)

PR #1000, opened 2021-08-17
feat: enable avatar merge cull optimization heuristic

BrianAmadori:
## What does this PR change?
This PR forces all the opaque wearables that use double-sided materials to use back-face instead. This greatly enhances the performance of avatar rendering given by the avatar mesh merging system. 

The performance improvement works because using back-face on all opaques will reduce the need to create new materials for render state change. So, avatars with far less submeshes will be produced, in most cases only taking a single submesh (and this means a single draw call/no gl pipeline flush).

## How to test the changes?
Wearables that use double-sided opaques should be tested. A clear example are the wearables of the `urn:decentraland:ethereum:collections-v1:tech_tribal_marc0matic` collection.

Those wearables are currently in process of being fixed. When the wearables are fixed, then this PR could be merged.

## Our Code Review Standards


pravusjif (approved):
Tested in wonder mine and saw no problems!

PR #1002, opened 2021-08-17
fix: gltf download order

### LINKED ISSUE #906: Gltf loading prioritization can become inconsistent in some cases, leaving near assets to be enqueued last

pbosio:
Fix #906 
Replaced distance test function for a handler class to decide which GLTF in the queue should be downloaded.
Also now distance is only calculated horizontally.


Suduck (approved):

unity-renderer/Assets/UnityGLTF/Scripts/DownloadQueueHandler.cs
```diff
@@ -0,0 +1,117 @@
...
+            else if (gltfComponent.isInQueue)
+            {
+                CheckCompetitionFinished();
+            }
+        }
+
+        public bool IsNextInQueue()
+        {
+            //NOTE: start all downloads while renderer is disabled
+            if (mainCamera == null)
```
pravusjif, line 46: I think these "renderer state" checks would be better handled with the `RendererState` CommonScriptableObject, as we do in other places. We could have both checks in that conditional (I'd check the bool first).

https://github.com/decentraland/unity-renderer/blob/102fb7113e4b7acf8b8ade68112e81c25daf6edc/unity-renderer/Assets/Scripts/MainScripts/DCL/ScriptableObject/CommonScriptableObjects.cs#L153

pravusjif (approved):

unity-renderer/Assets/UnityGLTF/Scripts/DownloadQueueHandler.cs
BrianAmadori: A mild issue with this is that this falls into the GLTF assembly, and we should try to not reference our assemblies from there (I know we do it anyways for some cases, but we should try to avoid it as much as possible).

unity-renderer/Assets/UnityGLTF/Scripts/DownloadQueueHandler.cs
```diff
@@ -0,0 +1,117 @@
...
+
+namespace UnityGLTF
+{
+    internal class DownloadQueueHandler : IDisposable
+    {
+        private static event Action OnCompetitionReset;
+        private static GLTFComponent nearestComponent;
+        private static float nearestDistance = float.MaxValue;
+        private static int competingCount;
+        private static bool isCompetitionFinished = false;
```
BrianAmadori, line 12: Not sold at all with the design choice of making instances of `DownloadQueueHandler` with the queue competition being tracked by static variables. I know the original was worst though, but extracting the code to another class opens up opportunities to do test coverage of this logic. 

I think that even a singleton can be better than this because we can just instance the class to test it.
We can just make a `DownloadQueueHandler` composite as a singleton, in a way that each `GLTFComponent` creates its own `DownloadQueueHandler` like it does now, but reports it to the composite.

This means that we should add coverage to it too, so I'll add this as a request.

BrianAmadori (changes requested):
I read the code and saw some design issues. BUT how does this PR fix the queue issue? We are flattening the distances but we are still sending the `OnSilentForget` promises to Mordor (and not below the floor like the discussed workaround?).

unity-renderer/Assets/UnityGLTF/Scripts/GLTFComponent.cs
```diff
@@ -349,42 +358,22 @@ public IEnumerator LoadAssetCoroutine()
...
-            if (result)
-            {
-                //NOTE(Brian): Reset values so the other GLTFComponents running this coroutine compete again
-                //             for distance.
-                nearestGLTFComponent = null;
-                nearestDistance = float.MaxValue;
-            }
-
-            return result;
+            throw new NotImplementedException();
```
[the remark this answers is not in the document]
pbosio: We cannot cause it's a method  from `ILoadable` interface

unity-renderer/Assets/UnityGLTF/Scripts/DownloadQueueHandler.cs
pbosio: I have changed the approach for the handler to make it independent from static variables.
I've added the code for master promises that I had in other PR related to the download queue.
Sending forgotten master promise to the floor will not solve the issue cause we don't know where the pending promises are in world position, so for that cases I can only prioritize the download

unity-renderer/Assets/UnityGLTF/Scripts/DownloadQueueHandler.cs
pravusjif: If we don't want to reference our assemblies in the UnityGLTF library then I guess it's fine, ignore my original comment!

PR #1003, opened 2021-08-17
fix: prod rollout percentage

menduz:
## What does this PR change?

10% has proven to generate too much signal-noise ratio. Increasing to 50% to reduce falloff to prevent older versions to be selected in rollouts.

Screen Shot 2021-08-17 at 19 28 48

## How to test the changes?

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:{branch_name}&{desired_url_params}
2. ...

## Our Code Review Standards


kuruk-mm (approved):
Seems OK. We should implement a 100% percentage after some time because there is always a possibility of using very old versions.

PR #1012, opened 2021-08-18
feat: video component send events to kernel

### LINKED ISSUE #941: Get an event in the ECS when the video start/stop/pause
For static videos:
When a VideoTexture starts playing, there should be an event from Unity to Kernel signaling that the video started, the event should include the current position of the video in milliseconds.
When the video is paused or stopped, there should be also events informing the new state to the Kernel.
Once per second, if the video is playing, there should be a VideoProgressEvent(VideoTextureComponentId, offset, max) including the video component id and the milliseconds of the current playback. This event exists as a best effor to prevent synchronization/buffering issues between the Kernel and the Renderer.
Event definition
type VideoProgressEvent = {
  videoTextureId: string // component ID for the video texture
  status: number // 0 = paused, 1 = playing, 2 = buffering (paused)
  currentOffset: number // current absolute position of the video in milliseconds
  videoLength?: number // total video length in milliseconds, may be absent in streaming
}
For streaming:
When a VideoTexture starts streaming, there should be an event to know when the transmission started arriving. This event should include data about how long ago the transmission began for players that arrive late to the event (this information is typically included in any streaming data)

This is crucial for doing live events.

We often need to synchronize things that happen in the scene with a transmission, like light animations matching the music. The streaming always has a lag, which may often be up to 10 seconds, and depends on geography. If the instructions to run these animations are not properly synced to the actual arriving of the streaming, they will never be properly in sync with the video.

This task should do the changes relative to Unity and Kernel, but not the ecs
This task is relative to https://app.zenhub.com/workspaces/sdk-60febfe80dcd78000f01aca0/issues/decentraland/sdk/11

Kinerius:
## What does this PR change?

Fixes #941 

This PR containers a refactor to the WebVideoPlayer class that now receives a plugin interface so it could be mocked and fully tested.

## How to test the changes?

1. Go to: https://play.decentraland.zone/index.html?renderer=urn:decentraland:off-chain:renderer-artifacts:{branch_name}&{desired_url_params}
2. ...

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
@@ -95,7 +95,7 @@ namespace DCL.Components
 
                 string videoId = (!string.IsNullOrEmpty(scene.sceneData.id)) ? scene.sceneData.id + id : scene.GetHashCode().ToString() + id;
                 texturePlayer = new WebVideoPlayer(videoId, dclVideoClip.GetUrl(), dclVideoClip.isStream, new WebVideoPlayerNative());
-                texturePlayerUpdateRoutine = CoroutineStarter.Start(VideoTextureUpdate());
+                texturePlayerUpdateRoutine = CoroutineStarter.Start(OnUpdate());
                 CommonScriptableObjects.playerCoords.OnChange += OnPlayerCoordsChanged;
                 CommonScriptableObjects.sceneID.OnChange += OnSceneIDChanged;
                 scene.OnEntityRemoved += OnEntityRemoved;
@@ -178,33 +178,42 @@ namespace DCL.Components
             texture.Apply(unitySamplingMode != FilterMode.Point, true);
         }
 
-        private IEnumerator VideoTextureUpdate()
+        private IEnumerator OnUpdate()
         {
             while (true)
             {
-                if (isPlayStateDirty)
-                {
-                    CalculateVideoVolumeAndPlayStatus();
-                    isPlayStateDirty = false;
-                }
-
-                if (!isPlayerInScene && currUpdateIntervalTime < OUTOFSCENE_TEX_UPDATE_INTERVAL_IN_SECONDS)
-                {
-                    currUpdateIntervalTime += Time.unscaledDeltaTime;
-                }
-                else if (texturePlayer != null && !isTest)
-                {
-                    currUpdateIntervalTime = 0;
-                    texturePlayer.UpdateWebVideoTexture();
-                }
-
-                if (texturePlayer.playing && IsTimeToReportVideoProgress())
-                {
-                    lastVideoProgressReportTime = Time.unscaledTime;
-                    ReportVideoProgress();
-                }
-
+                UpdateDirtyState();
+                UpdateVideoTexture();
+                UpdateProgressReport();
                 yield return null;
+            }
+        }
+        private void UpdateDirtyState()
+        {
+            if (isPlayStateDirty)
+            {
+                CalculateVideoVolumeAndPlayStatus();
+                isPlayStateDirty = false;
+            }
+        }
+        private void UpdateVideoTexture()
+        {
+            if (!isPlayerInScene && currUpdateIntervalTime < OUTOFSCENE_TEX_UPDATE_INTERVAL_IN_SECONDS)
+            {
+                currUpdateIntervalTime += Time.unscaledDeltaTime;
+            }
+            else if (texturePlayer != null && !isTest)
+            {
+                currUpdateIntervalTime = 0;
+                texturePlayer.UpdateWebVideoTexture();
+            }
+        }
+        private void UpdateProgressReport()
+        {
+            if (texturePlayer.playing && IsTimeToReportVideoProgress())
+            {
+                lastVideoProgressReportTime = Time.unscaledTime;
+                ReportVideoProgress();
             }
         }
         private void ReportVideoProgress()
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
@@ -148,12 +148,12 @@ namespace DCL.Components
                 if (model.playing)
                 {
                     texturePlayer.Play();
-                    WebInterface.ReportVideoStartedEvent(lstSeekTime);
+                    ReportVideoProgress();
                 }
                 else
                 {
                     texturePlayer.Pause();
-                    WebInterface.ReportVideoPausedEvent(texturePlayer.GetTime());
+                    ReportVideoProgress();
                 }
 
                 if (baseVolume != model.volume)
@@ -199,29 +199,29 @@ namespace DCL.Components
                     texturePlayer.UpdateWebVideoTexture();
                 }
 
-                ReportVideoProgress();
+                if (texturePlayer.playing && IsTimeToReportVideoProgress())
+                {
+                    lastVideoProgressReportTime = Time.unscaledTime;
+                    ReportVideoProgress();
+                }
 
                 yield return null;
             }
         }
         private void ReportVideoProgress()
         {
-            if (texturePlayer.playing && IsTimeToReportVideoProgress())
-            {
-                lastVideoProgressReportTime = Time.unscaledTime;
-                var videoStatus = GetVideoStatus();
-                var currentOffset = texturePlayer.GetTime();
-                var length = texturePlayer.GetDuration();
-                WebInterface.ReportVideoProgressEvent(lastVideoClipID, videoStatus, currentOffset, length );
-            }
+            var videoStatus = GetVideoStatus();
+            var currentOffset = texturePlayer.GetTime();
+            var length = texturePlayer.GetDuration();
+            WebInterface.ReportVideoProgressEvent(id, scene.sceneData.id, lastVideoClipID, videoStatus, currentOffset, length );
         }
         private int GetVideoStatus()
         {
             //TODO: Handle buffering state that returns 2
-            
+
             if (texturePlayer.playing)
                 return 1;
-            
+
             return 0;
         }
         private bool IsTimeToReportVideoProgress() { return Time.unscaledTime - lastVideoProgressReportTime > VIDEO_PROGRESS_UPDATE_INTERVAL; }
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
@@ -17,8 +17,8 @@ namespace DCL.Components
         internal static bool isTest = false;
 #endif
 
-        private const float OUTOFSCENE_TEX_UPDATE_INTERVAL = 1.5f;
-        private const float VIDEO_PROGRESS_UPDATE_INTERVAL = 1f;
+        private const float OUTOFSCENE_TEX_UPDATE_INTERVAL_IN_SECONDS = 1.5f;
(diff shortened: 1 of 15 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/WebInterface/Interface.cs
```diff
@@ -1240,5 +1261,37 @@ public static void NotifyStatusThroughChat(string message)
...
+        }
+        public static void ReportVideoPausedEvent(float videoPosition)
+        {
+            SendVideoPausedEvent sendVideoStartedEvent = new SendVideoPausedEvent()
+            {
+                currentPosition = videoPosition
+            };
+            
+            SendMessage("VideoPausedEvent", sendVideoStartedEvent);
+        }
```
menduz, line 1281: Can this be replaced by the `ReportVideoProgressEvent`? Embedding all the data of the video in a single atomic message is better for the data-oriented approach of the ECS, it would add a component to the video that has a consistent full state all the times. What do you think?

The package is also missing the SceneId and the componentId of the video texture, that is necessary to redirect the event directly to the scene

menduz (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/Tests/VideoTests.cs
```diff
@@ -43,17 +43,23 @@ public IEnumerator MessageIsSentWhenVideoPlays()
...
-            var expectedEvent = new WebInterface.SendVideoStartedEvent()
+            var expectedEvent = new WebInterface.SendVideoProgressEvent()
             {
-                currentPosition = 10
+                sceneId = scene.sceneData.id,
+                componentId = component.id,
+                videoLength = 0,
+                videoTextureId = id,
+                currentOffset = 0,
+                status = 1
```
menduz, line 55: Should the "status" value be an enum to prevent accidents?

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
```diff
@@ -140,9 +146,15 @@ public override IEnumerator ApplyChanges(BaseModel newModel)
...
                 if (model.playing)
+                {
                     texturePlayer.Play();
+                    ReportVideoProgress();
+                }
                 else
+                {
                     texturePlayer.Pause();
+                    ReportVideoProgress();
+                }
```
menduz, line 157: I would pay attention to this method, probably calling `ReportVideoProgress` at the end at every change may be a good idea.

```suggestion
                if (model.playing)
                {
                    texturePlayer.Play();
                }
                else
                {
                    texturePlayer.Pause();
                }
                ReportVideoProgress();
```

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
```diff
@@ -187,9 +199,32 @@ private IEnumerator VideoTextureUpdate()
...
+            var currentOffset = texturePlayer.GetTime();
+            var length = texturePlayer.GetDuration();
+            WebInterface.ReportVideoProgressEvent(id, scene.sceneData.id, lastVideoClipID, videoStatus, currentOffset, length );
+        }
+        private int GetVideoStatus()
+        {
+            //TODO: Handle buffering state that returns 2
+
+            if (texturePlayer.playing)
+                return 1;
```
menduz, line 223: Same here, this number may better be a constant or enum somewhere to prevent accidents and to make the code easier to read and maintain

menduz (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
Kinerius: Yes you are right, it made sense last time when they were different

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
Kinerius: Hmm this message structure is a DTO, I'm not sure if it's ok to create an enum for this because we already have a VideoState enum that has its own status
```c#
    public enum VideoState
    {
        NONE = 0,
        ERROR = 1,
        LOADING = 2,
        READY = 3,
        PLAYING = 4,
        BUFFERING = 5
    };
```
Maybe we can use the same status numbers at sdk.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
```diff
@@ -17,6 +18,7 @@ public class DCLVideoTexture : DCLTexture
 #endif
 
         private const float OUTOFSCENE_TEX_UPDATE_INTERVAL = 1.5f;
+        private const float VIDEO_PROGRESS_UPDATE_INTERVAL = 1f;
```
Suduck, line 21: Can we specify the measure of time for this? I guess is in seconds, so something like SECONDS_VIDEO_PROGRESS_UPDATE_INTERVAL

Suduck (approved):
LGTM! Small request!

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
Kinerius: I like this, made the change on last commit

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/Plugins/IWebVideoPlayerPlugin.cs
```diff
@@ -0,0 +1,23 @@
...
+        void SetVolume(string id, float volume) ;
+        int GetHeight(string id) ;
+        int GetWidth(string id) ;
+        float GetTime(string id) ;
+        float GetDuration(string id) ;
+        int GetState(string id) ;
+        string GetError(string id) ;
+        void SetTime(string id, float second) ;
+        void SetPlaybackRate(string id, float playbackRate) ;
+        void SetLoop(string id, bool loop) ;
```
BrianAmadori, line 21:
```suggestion
        void Create(string id, string url, bool useHls);
        void Remove(string id);
        void TextureUpdate(string id, IntPtr texturePtr, bool isWebGL1);
        void Play(string id, float startTime);
        void Pause(string id);
        void SetVolume(string id, float volume);
        int GetHeight(string id);
        int GetWidth(string id);
        float GetTime(string id);
        float GetDuration(string id);
        int GetState(string id);
        string GetError(string id);
        void SetTime(string id, float second);
        void SetPlaybackRate(string id, float playbackRate);
        void SetLoop(string id, bool loop);
```
We should remove spaces before the semicolon.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/Plugins/WebVideoPlayerNative.cs
```diff
@@ -0,0 +1,73 @@
...
+        private static int WebVideoPlayerGetHeight(string id) { return 0; }
+        private static int WebVideoPlayerGetWidth(string id) { return 0; }
+        private static float WebVideoPlayerGetTime(string id) { return 0; }
+        private static float WebVideoPlayerGetDuration(string id) { return 0; }
+        private static int WebVideoPlayerGetState(string id) { return (int)VideoState.ERROR; }
+        private static string WebVideoPlayerGetError(string id) { return "WebVideoPlayer: Platform not supported"; }
+        private static void WebVideoPlayerSetTime(string id, float second) { }
+        private static void WebVideoPlayerSetPlaybackRate(string id, float playbackRate) { }
+        private static void WebVideoPlayerSetLoop(string id, bool loop) { }
+#endif
```
BrianAmadori, line 55: Cool refactor. We can't move the static methods to a `Utils` class or something similar? This code would be far easier to follow with the change.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/Tests/WebVideoPlayerTests.cs
```diff
@@ -0,0 +1,213 @@
+using DCL.Components.Video.Plugin;
+using NSubstitute;
+using NUnit.Framework;
+using UnityEngine;
+
+namespace Tests
+{
+    public class WebVideoPlayerTests
```
BrianAmadori, line 8: We are moving to the `Should...When` naming convention for tests. New test suites should abide to that convention if possible.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
```diff
@@ -187,9 +198,23 @@ private IEnumerator VideoTextureUpdate()
                     texturePlayer.UpdateWebVideoTexture();
                 }
 
+                if (texturePlayer.playing && IsTimeToReportVideoProgress())
```
BrianAmadori, line 201: I think that this component logic is already getting pretty convoluted. We can't extract all the logic related to reporting the video progress to another class? (I don't mind if we have to create another coroutine for handling this).

BrianAmadori (commented):
Pretty cool refactor, ty! Left some minor remarks to address before we merge it.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/Tests/VideoTests.cs
```diff
@@ -34,6 +34,95 @@ public IEnumerator VideoTextureIsCreatedCorrectly()
...
+                videoLength = 0,
+                videoTextureId = id,
+                currentOffset = 0,
+                status = (int)VideoState.ERROR
+            };
+            var json = JsonUtility.ToJson(expectedEvent);
+            var wasEventSent = false;
+            yield return TestHelpers.WaitForMessageFromEngine("VideoProgressEvent", json,
+                () => { },
+                () =>wasEventSent = true);
```
kuruk-mm, line 121:
```suggestion
                () => wasEventSent = true);
```

kuruk-mm (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/Plugins/WebVideoPlayerNative.cs
Kinerius: Wouldn't the Utils class be the same as this then? you need to handle the existence of this DLL on multiple platforms and this class just does that

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/Tests/WebVideoPlayerTests.cs
Kinerius: Got it, will rename the class for now.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
Kinerius: What about single level of abstraction? I prefer having less coroutines.
```c#
private IEnumerator OnUpdate()
{
    while (true)
    {
        UpdateDirtyState();
        UpdateVideoTexture();
        UpdateProgressReport();
        yield return null;
    }
}
```
Moving the report logic to another class could be another refactor or just leave it as technical debt

BrianAmadori: Sounds good!

BrianAmadori (approved):
Approving, will really appreciate if the feedback is addressed before the merge. TY!.

PR #1037, opened 2021-08-20
fix: make unity editor work again with race condition fixes for .zone [MERGE ASAP TO UNBLOCK TEAM]

menduz:
## What does this PR change?
@BrianAmadori @kuruk-mm there is something strange happening in that file. I left you a comment to apply your knowledge in the future and debug what is happening.

Here is a sequence diagram about how things should work between renderer and kernel.

Screen Shot 2021-08-20 at 20 18 19
 

## How to test the changes?

Go to the editor and connect it to https://play.decentraland.zone, using the WSSController it should work now.

## Our Code Review Standards


kuruk-mm (approved):
I'm going to refactor all this code in the next week. So hacky/temporal solutions are ok. I did some debugging, and idk why is not sending the message with `WebInterface...`

BrianAmadori (approved):

BrianAmadori:
I think I rushed to merge this. `WSSController` debugging still doesn't work with this change. @menduz

kuruk-mm:
I'm doing the `WSSController` refactor, and I saw that some messages are sent the `MessageFromEngine` callback (in WebInterface) is assigned (this is why `SendMessageToWeb` works and `WebInterface.SendSystemInfoReport` don't) or before the socket is open.

So I think we should `queue` the messages sent before the WebSocket connection starts and flush those when the callback is assigned or connected.

PR #1101, opened 2021-08-30
style: quest mini polish

RominaMarchetti:
- Read more pop up doesn't overlap the main panel anymore.
- "To do" title is now "Current".
- Fixed Heavy Black Stroke. White texts were hard to read.
- Fonts sizes homogenized. 

After
image

Now
Screen Shot 2021-08-30 at 13 59 00


BrianAmadori (approved):
Looking way better now, ty!

pravusjif (approved):

PR #1103, opened 2021-08-30
fix: race condition null reference issue in BodyShapeController

BrianAmadori:
## What does this PR change?

Many null references were detected on crowded places related to facial features controller trying to set colors before the body shape controller was ready. This PR fixes this issue.

## How to test the changes?

1. Go to WonderMine in editor mode
2. Note that in `master` we get a bunch of null references coming from `FacialFeatureController.PrepareWearable`.
3. This shouldn't happen anymore on this branch.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/BodyShapeController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/BodyShapeController.cs
@@ -5,6 +5,7 @@ using UnityEngine;
 
 public interface IBodyShapeController
 {
+    bool isReady { get; }
     string bodyShapeId { get; }
     void SetupEyes(Material material, Texture texture, Texture mask, Color color);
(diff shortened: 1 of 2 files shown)


AjimenezDCL (commented):
I think we have an underlying issue worth investigating.

This issue didn't happen before and if you take a look at `AvatarRenderer`: https://github.com/decentraland/unity-renderer/blob/6afc068a16e6f83fc5a503136362f282fb679758/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs#L407
The facial features are already waiting for the bodyshape to be ready.

I think this fix doesn't actually addresses the issue but in a couple of minutes of testing I have not seen the bug yet.

Maybe `BodyShapeController.isReady` is not reliable anymore due to changes in the `RendereableAssetLoadHelper`?

AjimenezDCL (approved):
Nice catch, it was probably the issue with the CleanUp not disposing the routine!

PR #1121, opened 2021-09-02
chore: optimize avatar names

### LINKED ISSUE #1086: Limit avatar visible names when too many avatars are on the screen
Right now, if we instantiate 200 avatars, 200 names will be rendered. This has a performance overhead we should mitigate.

We should only show a limited number of the closest names (i.e. 50).

AjimenezDCL:
Fixes #1086 
Now avatar names are tied to `AvatarsLODController`. Current implementation just enables the avatar name for visible avatars within LOD0/LOD1 distance.

If we are still experiencing performances issues we could even lower the amount.

If `ENABLE_AVATAR_LODS` is disabled the algorithm performs as before with a hard maximum of 100 avatar names.

### Testing with bots 
1. enter https://play.decentraland.zone/?renderer-branch=chore/opimize-avatar-names&NETWORK=ropsten&position=91,151
2. open browser console
3. run `clientDebug.InstantiateBotsAtCoords({ amount: 100, xCoord: 91.2, yCoord: 151.2, areaWidth: 36, areaDepth: 50 })` to instantiate 100 bot in the stadium. It may take a little while to load as the bots collections and wearables are being randomized and loaded.

### Testing with real users
1. enter https://play.decentraland.zone/?renderer-branch=chore/opimize-avatar-names&NETWORK=mainnet&position=-27,60

Be sure to test in `.zone` and `.org` (with and without LODS)


pravusjif (approved):
tested using `MAX_VISIBLE_PEERS=200`, also with and without `ENABLE_AVATAR_LODS` and saw no issues

sandrade-dcl (approved):
LGTM!

PR #1126, opened 2021-09-02
fix: add defines to only use native bridge on webgl, and only dump pe…

kuruk-mm:
…rformance metric on unity editor to use it on desktop

## What does this PR change?

On the WebSocketCommunication we're trying to use (Web)NativeCommunication instead of WebSocket.

And we're dumping some performance data when we are out of the UNITY_EDITOR that are not needed

## Our Code Review Standards


pravusjif (approved):
Had no problems testing the build

PR #1128, opened 2021-09-03
fix: special rotating platforms going full helicopter

### LINKED ISSUE #1122: Dragon Rush avatar rotation is broken
When I get on the dragon in Dragon rush and try to fly, I start spinning like crazy

This is likely related to the rotation of avatars on rotating platforms

https://play.decentraland.org/?island=I9d82&position=-42%2C54&realm=odin

Kinerius:
## What does this PR change?

Disabled camera rotation when platforms rotate on First Person camera.
Player rotation report is now based on camera rotation only.

Fixes #1122 

## How to test the changes?

Go to Dragon Rush, get on the dragon and smash F, neither FPS or TPS cameras should go helicopter mode.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Camera/CameraStateFPS.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Camera/CameraStateFPS.cs
@@ -17,44 +17,7 @@ namespace DCL.Camera
             if (defaultVirtualCamera is CinemachineVirtualCamera vcamera)
                 pov = vcamera.GetCinemachineComponent<CinemachinePOV>();
         }
-
-        public override void OnSelect()
-        {
-            base.OnSelect();
-            CommonScriptableObjects.playerIsOnMovingPlatform.OnChange += OnMovingPlatformStart;
-            if (CommonScriptableObjects.playerIsOnMovingPlatform.Get())
-                OnMovingPlatformStart(true, true);
-        }
-
-        public override void OnUnselect()
-        {
-            base.OnUnselect();
-            CommonScriptableObjects.playerIsOnMovingPlatform.OnChange -= OnMovingPlatformStart;
-            CommonScriptableObjects.movingPlatformRotationDelta.OnChange -= OnMovingPlatformRotate;
-        }
-        private void OnMovingPlatformStart(bool current, bool previous)
-        {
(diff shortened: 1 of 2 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Camera/CameraStateFPS.cs
```diff
@@ -45,6 +45,8 @@ private void OnMovingPlatformStart(bool current, bool previous)
         }
         private void OnMovingPlatformRotate(Quaternion current, Quaternion previous)
         {
+            // Disabled camera rotation on moving platforms because this was causing issues for some scenes
```
BrianAmadori, line 48: If this method is causing issues we should just remove it entirely.

BrianAmadori (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Camera/CameraStateFPS.cs
menduz: Please track an issue to re-enable the feature, removing a behavior may be a good temporary fix for the helicopter situation but we should restore the original behavior

pravusjif:
I don't like that we are reverting [this fix](https://github.com/decentraland/unity-renderer/pull/864) so that a scene (Dragon Rush) keeps working as it was when they implemented their code according to the our bugged reported rotation.

I believe the best solution would be to change the scene code to use the camera rotation instead of the character rotation (I don't know if we already are reporting the _character rotation_ and the _camera rotation_ separately, so that the creators can use any of them separately, otherwise that would be needed for the proper solution)

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Camera/CameraStateFPS.cs
pravusjif: I would either remove it (but we will need to re-implement it again once we have a new `playerRoation` or similar exposed for the SDK) or comment it out with a clear comment that it has to be uncommented when we have that new property to use.

pravusjif (approved):

BrianAmadori (approved):

Issue #1125, opened 2021-09-02
Too aggressive shading introduced by GPU skinning


BrianAmadori:
This issue is happening because the normals "rotate" with the original mesh rotation. The only rotation that should be taken into account is the root bone.

Before:
![Old](https://user-images.githubusercontent.com/6875814/132073587-bdbb8825-a6be-4fe1-b99c-33bcc99f80da.png)

After GPU skinning:
![New](https://user-images.githubusercontent.com/6875814/132073600-39fa67b8-16aa-425f-9111-1d264108015f.png)

PR #1130, opened 2021-09-03
fix: gpu skinning normal issues

BrianAmadori:
## What does this PR change?
Now, the avatar rendering should be identical as before the GPU skinning was applied. Address #1125.

* Matrix calculations for GPU skinning were rewritten to be compatible with normal transformations. Also the formula looks nicer now.
* Normal transformations don't depend anymore on the object transform (before, rotating the mesh renderer rotated the normals when they shouldn't).
* Changed fragment normal from tangent space to object space.
* Nullified world normal recalculations caused by Variants.hlsl on vertex shader because they conflicted with our calculated normals, as the output of the normals of the skinning operation are already in world space.

...

## How to test the changes?

2. ...

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Rendering/Shaders/Toon/Compiled/DepthNormalsOnlyPass.hlsl
+++ b/unity-renderer/Assets/Rendering/Shaders/Toon/Compiled/DepthNormalsOnlyPass.hlsl
@@ -9,6 +9,7 @@ PackedVaryings vert(Attributes input)
     ApplyGPUSkinning(input, input.tangentOS, input.uv1);
     #endif
     output = BuildVaryings(input);
+    output.normalWS = input.normalOS;
     PackedVaryings packedOutput = (PackedVaryings)0;
     packedOutput = PackVaryings(output);
     return packedOutput;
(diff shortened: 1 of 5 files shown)


unity-renderer/Assets/Rendering/Shaders/Toon/Compiled/DepthNormalsOnlyPass.hlsl
```diff
@@ -9,6 +9,7 @@ PackedVaryings vert(Attributes input)
     ApplyGPUSkinning(input, input.tangentOS, input.uv1);
     #endif
     output = BuildVaryings(input);
+    output.normalWS = input.normalOS;
```
AjimenezDCL, line 12: If we cannot (easily) set the shadergraph to expect a OS normal I would add a comment there for the future. It will be confusing to get OS normals in a variable called `normalWS`.

AjimenezDCL (approved):
Great job, approved with a small comment.

unity-renderer/Assets/Rendering/Shaders/Toon/Compiled/DepthNormalsOnlyPass.hlsl
BrianAmadori: The thing is they are not OS normals, they are actually WS normals. Going to see what I can do about this.

pravusjif (approved):
Tested in wondermine and it looks fixed
https://play.decentraland.zone/?renderer-branch=fix/gpu-skinning-normal-issues&MAX_VISIBLE_PEERS=200&position=-28%2C55

unity-renderer/Assets/Rendering/Shaders/Toon/Compiled/DepthNormalsOnlyPass.hlsl
BrianAmadori: Done, added a TODO comment but I should refactor this in the future.

PR #1131, opened 2021-09-06
fix: AvatarModiferArea is only hiding own avatar (should hide all)

### LINKED ISSUE #1089: AvatarModiferArea is only hiding own avatar (should hide all)
The AvatarModiferArea should hide every avatar that enters the area, and display every avatar that gets out of it.

Lastraum reported that currently this feature is bugged and only works for the own player and not the other avatars.

Let's create some tests to cover the correct functionality of the feature.

sandrade-dcl:
## What does this PR change?

Fixes #1089 

The AvatarModiferArea should hide every avatar that enters the area, and display every avatar that gets out of it.

Lastraum reported that currently this feature is bugged and only works for the own player and not the other avatars.

Let's create some tests to cover the correct functionality of the feature.

## Steps to reproduce (from preview mode)
1. Create a new scene with `dcl init` and paste this code:
```
const modArea = new Entity()
modArea.addComponent(
  new AvatarModifierArea({
    area: { box: new Vector3(16, 4, 16) },
    modifiers: [AvatarModifiers.HIDE_AVATARS],
  })
)
modArea.addComponent(
  new Transform({
    position: new Vector3(8, 0, 8),
  })
)
engine.addEntity(modArea)
```
2. Play the scene in preview mode (`dcl start`) and debug it from Unity (using this branch).
3. Open other browser in incognito mode with the same url in order to have a second user in the scene.
4. Enter and exit into the avatar modifier area with both avatars and notice they are dissapearing correctly.

## Test Link
https://play.decentraland.zone/?renderer-branch=fix/AvatarModiferArea-is-only-hiding-own-avatar

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarVisibility.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarVisibility.cs
@@ -40,10 +40,13 @@ namespace DCL
             OnSetVisibility?.Invoke(visibility);
         }
 
-        private void SetVisibilityForGameObjects(bool value)
+        internal void SetVisibilityForGameObjects(bool value)
         {
             foreach (GameObject gameObject in gameObjectsToToggle)
             {
+                if (gameObject == null)
+                    continue;
+
                 gameObject.SetActive(value);
             }
         }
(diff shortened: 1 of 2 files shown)


pravusjif (approved):
Tested in preview mode and the issue is now fixed, great job!

PR #1139, opened 2021-09-07
fix: camera rotation on teleport or forced set

### LINKED ISSUE #999: 3rd person camera starts on the floor when spawning
Camera angles seem to be messed up when spawing in 3rd person into scenes that don't have a spawn point defined.

If I spawn into genesis plaza, for example, the camera is fine in 1st or 3rd person.
But if I spawn into any scene that doesn't have an explicitly defined spawn point, like an empty parcel, the camera looks fine in 1st person, but not in 3rd. The initial 3rd person camera is pointing from the floor, up.

https://play.decentraland.org/?position=-55%2C-6&realm=artemis-pink

this currently also happens in zone

I suspect the issue is with how the default camera angle of the default spawn point in 0,0,0 is interpreted.

Kinerius:
## What does this PR change?

Fix #999 
...

## How to test the changes?

1 - `/goto 10,10` camera should point forward
2 - `/goto 0,0` camera should point forward and should not move through the floor

## Our Code Review Standards


BrianAmadori (approved):
LGTM

sandrade-dcl (approved):
Reviewed and tested. It is working fine, good work!

Issue #1084, opened 2021-08-27
Add throttling to animation and gpu skinning components


BrianAmadori:
When the scenes are too crowded, we should mitigate the overhead of uploading the bone matrices each frame for each avatar.

* When more than a given number of avatars are on the screen, the far away avatars that are not impostors should only be updated with given frame interval.

* This should be done with a smart algorithm, so it's ensured that no more than X avatars have GPU skinning and Animation component updated each frame.

PR #1140, opened 2021-09-07
feat: gpu skinning throttling

AjimenezDCL:
#1084 

Take a look at the different implementations (two commits), Hardcoded values in an ENUM vs Animation Curve


pravusjif (approved):
Tested with bots and users and it works great, good job!

PR #1157, opened 2021-09-09
fix: aggressive thumbnail loading on avatar load

BrianAmadori:
## What does this PR change?
Addresses #1158 

This PR introduces a new way of loading the avatar snapshots through the new `LazyTextureObserver` helper. This helper makes sure a texture is only loaded when a listener marks that the texture is needed. 

Also, if the texture should change because the profile is updated, `LazyTextureObserver` notifies all its listeners of the change.

A few changes on the UI panels that use snapshots were introduced to only load textures when the UI is visible (i.e. enabled). The outcome is that when you open a UI that use them, they will start showing as blank and lazily load over time.

## How to test the changes?
The changes should affect the following HUDs:
* The neighboring users panel that's on the minimap
* Everything related with friends (private conversation, speech bubbles, etc)
* Passport
* Your profile card

The snapshots should unload/load when the HUDs visibility is changed. A good test is to close and open the HUDs and ensure the snapshots load as intended.

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/LazyTextureObserver/LazyTextureObserver.cs
```diff
@@ -0,0 +1,234 @@
...
+    /// All existing listeners will be keep informed of <b>RefreshWithX</b> calls to update the texture if it changes.
+    /// </summary>
+    public class LazyTextureObserver : ILazyTextureObserver
+    {
+        enum State
+        {
+            NONE,
+            IN_PROGRESS,
+            LOADED
+        }
```
sandrade-dcl, line 72: ❤️

sandrade-dcl (approved):

sandrade-dcl (approved):
LGTM! Please take a llok to the tests in CI.

pbosio (approved):
LGTM.
My only concern is that, for example, every time I close and open the nearby user list the thumbnail for those profiles are fetch again

AjimenezDCL (approved):

pravusjif (approved):

PR #1159, opened 2021-09-09
chore: set AvatarShape layer to ViewportCullingIgnored

AjimenezDCL:
Signed-off-by: Alejandro Jimenez <ajimenez@decentraland.org>

Culling controller were messing with our avatars optimizations, transitions and impostors. We ignore them from now on.

To test this PR just go to a crowded scene (or put some bots in) and check if small avatars in the distance get culled off if the setting is enabled. It shouldnt.

## Our Code Review Standards


sandrade-dcl (changes requested):
I'm doing some tests in Wondermine (with the culling setting deactivated) and I'm seeing how the very far avatars are hidden.

sandrade-dcl (approved):

menduz:
@AjimenezDCL for future times, we should add a test to prevent this from breaking again. It should be three lines only and I _strongly_ suggest to make every PR with tests or otherwise do not merge.

CC @BrianAmadori @olavra

PS: Why is my face showing up in the commit??

PR #1176, opened 2021-09-10
chore: reduce body snapshot resolution

### LINKED ISSUE #1177: Body snapshots are too big
Currently the body snapshot texture is 512x1024, to reduce the memory needed to load them we cut the resolution in half

AjimenezDCL:
Fixes #1177 

Reduce the resolution of the body snapshot taken.

The only way to test it would be to save your avatar in the backpack and then retrieve your profile from https://peer.decentraland.org/lambdas/profile/{wallet}

## Our Code Review Standards


Kinerius (approved):
LGTM

PR #1188, opened 2021-09-13
fix: change fonts used by the sdk

### LINKED ISSUE #1109: Override San Francisco font usage from the SDK
We are no longer supporting the use of San Francisco font i, and its been replaced by Inter. We still need to keep retro compatibility with scenes that are currently deployed and using that font. If any entity or component uses this font, it need to be replaced internally by the "Inter" alternative in all its variants (semi bold, heavy, regular). 

To prevent breaking compatibility again, the SDK will also start sending `sans-serif` `sans-serif semibold` and `sans-serif heavy` to refer to the new font. The actual font will be an implementation detail of the renderer.

pbosio:
fixes #1109

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Font/DCLFont.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Font/DCLFont.cs
@@ -10,8 +10,24 @@ namespace DCL.Components
 {
     public class DCLFont : BaseDisposable
     {
-        const string RESOURCE_FONT_PREFIX = "builtin:";
         const string RESOURCE_FONT_FOLDER = "Fonts & Materials";
+
+        private const string DEFAULT_SANS_SERIF_HEAVY = "Inter-Heavy SDF";
+        private const string DEFAULT_SANS_SERIF_BOLD = "Inter-Bold SDF";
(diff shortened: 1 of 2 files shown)


RominaMarchetti (approved):

sandrade-dcl (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Font/DCLFont.cs
```diff
@@ -101,7 +117,7 @@ public override IEnumerator ApplyChanges(BaseModel newModel)
                 }
                 else
                 {
-                    Debug.Log($"couldn't fetch font from resources {resourceName}");
+                    Debug.Log($"couldn't fetch font from resources {fontResourceName}");
```
menduz, line 120: Should we load the default font as a fallback?

pbosio: The "default" font is already set in the component prefab...
since TextShape has LiberationSans and UIText had San Francisco (now replaced for Inter) as default font I opted for a "couldn't load font then don't modify the already existing font for that component" approach.
but I could load a default font in that scenario if you prefer that

PR #1194, opened 2021-09-14
fix: race condition leaving orphan gltf

pbosio:
The following scenario was leaving some GLTF orphan:
* keep `promiseA` for gltf asset `X`
* forget `promiseA`
* keep `promiseB` for gltf asset `X`
* 
if `promiseB` finish loading gltf `X` before `promiseA` (since we are waiting for gltf to finish loading) we will end up having an orphan gltf `X`

can be tested in [WonderMine `-26,59`](https://play.decentraland.zone/?renderer-branch=fix/orphan-gltfs&position=-26,59) when not using asset bundles
image

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/GLTF/AssetPromise_GLTF.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/GLTF/AssetPromise_GLTF.cs
@@ -194,10 +194,11 @@ namespace DCL
         {
             asset = loadedAsset;
             waitingAssetLoad = false;
+            bool alreadyInLibrary = false;
 
             if (success && asset != null)
             {
-                bool alreadyInLibrary = library.Contains(asset);
+                alreadyInLibrary = library.Contains(asset);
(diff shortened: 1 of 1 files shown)


menduz (commented):
How difficult is to add tests to this to prevent the future collapse of civilization?

pbosio:
> How difficult is to add tests to this to prevent the future collapse of civilization?

I will try to code a test for that, but since I don't know how much time it will take (not very optimistic about that) I would prefer to send this fix to prod ASAP since it's probably breaking a lot of other stuff not just the visual issue in wonder mine but it's probably also leaking memory all over the place :S

sandrade-dcl (changes requested):
I have just tested the build and it is still happening the issue:

image

pravusjif (approved):
Tested the build and it fixes the issue for me. It also fixes the other problem in which going from Wonder Mine to -41, 10 would then show me the pink meteor in those other coordinates in the middle of the streat. Now it's fixed

BrianAmadori (approved):
The approach looks fine. However let's prioritize implementing a test after this is merged if possible.

sandrade-dcl (approved):

PR #1207, opened 2021-09-15
fix: teleport memory handling

BrianAmadori:
## What does this PR change?

This PR add small tweaks to how we free memory when teleporting.

* When clearing entities in "immediate mode" now the pool is cleaned up.
* When the rendering is deactivated to start a teleport, `ParcelScenesCleaner` pending entities are cleaned up and `Resources.UnloadUnusedAssets` is called. (before this happened when the rendering was **activated**, leading to potential unwanted memory spikes)
* Added a few refactors to improve `MemoryManager` and `ParcelScenesCleaner` encapsulation. `MemoryManager` now just exposes a `OnCriticalMemory` event. `MemoryManager` responsibility was simplified, and now it doesn't uses `ParcelScenesCleaner` anymore. The `MemoryManager` responsibility of cleaning pools is still in place, but the pool cleaning code was moved to `PoolManager`.

These changes shouldn't fix the recurring memory out of bounds situation. The memory issue is a combination of many issues that are going to be addressed in subsequent PRs. Those PR are going to take advantage of the changes presented on this one.

## How to test the changes?

- Try to teleport a few times around the world.
- As this PR introduced a more aggresive unloading on teleports, we should ensure loading times don't increase very much.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ParcelScenesCleaner.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ParcelScenesCleaner.cs
@@ -41,12 +41,24 @@ namespace DCL
 
         Coroutine removeEntitiesCoroutine;
 
-        public void Start() { removeEntitiesCoroutine = CoroutineStarter.Start(CleanupEntitiesCoroutine()); }
-
-        public void Stop()
+        public ParcelScenesCleaner ()
         {
-            if (removeEntitiesCoroutine != null)
-                CoroutineStarter.Stop(removeEntitiesCoroutine);
+            removeEntitiesCoroutine = CoroutineStarter.Start(CleanupEntitiesCoroutine());
+            CommonScriptableObjects.rendererState.OnChange += OnRendererStateChange;
+        }
+
+        public void Initialize()
+        {
+            Environment.i.platform.memoryManager.OnCriticalMemory += ForceCleanup;
+        }
+
+        private void OnRendererStateChange(bool isEnable, bool prevState)
+        {
+            if (!isEnable)
+            {
+                ForceCleanup();
+                Resources.UnloadUnusedAssets();
+            }
         }
 
         public void MarkForCleanup(IDCLEntity entity)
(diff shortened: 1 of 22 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ParcelScenesCleaner.cs
```diff
@@ -92,94 +117,58 @@ public void ForceCleanup()
...
             while (entitiesMarkedForCleanup.Count > 0)
             {
                 IDCLEntity entity = entitiesMarkedForCleanup.Dequeue();
                 entity.SetParent(null);
                 entity.Cleanup();
+
+                if (!immediate && DCLTime.realtimeSinceStartup - lastTime >= MAX_TIME_BUDGET)
+                {
+                    yield return null;
+                    lastTime = Time.unscaledTime;
```
Suduck, line 137: Shouldn't this be also DCLTime.realtimeSinceStartup?

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ParcelScenesCleaner.cs
```diff
@@ -92,94 +117,58 @@ public void ForceCleanup()
...
 
             foreach (var scene in scenesToRemove)
             {
                 if (scene != null && !Environment.i.world.state.loadedScenes.ContainsKey(scene.sceneData.id))
                     Object.Destroy(scene.gameObject);
+
+                if (!immediate && DCLTime.realtimeSinceStartup - lastTime >= MAX_TIME_BUDGET)
+                {
+                    yield return null;
+                    lastTime = Time.unscaledTime;
```
Suduck, line 149: same here

Suduck (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ParcelScenesCleaner.cs
BrianAmadori: `DCLTime.realtimeSinceStartup` is less performant than `Time.unscaledTime`, but it works for cases where we need to measure time between instructions without skipping a frame. This is because `Time.time` and `Time.unscaledTime` is just updated a single time when the frame starts. Here, as we just skipped a frame, `Time.unscaledTime` is enough.

sandrade-dcl (approved):

PR #1213, opened 2021-09-17
fix: skin and hair materials are unable to collide now.

AjimenezDCL:
Hair and skin rely on the material name to be colored properly and our algorithm to reuse materials was getting in the way. If a material shares the same CRC with one already used, we reuse it.

The problem is that the material's name is not part of the CRC calculation so we had the following issue:
1) `WearableMaterial_Skin` is processed and cached with `CRC = 12341234`.
2) `WearableMaterial_Hair` is processed and his CRC is also `12341234` so we reuse `WearableMaterial_Skin` and it's assigned to the mesh.
3) We end up tinting as Skin a mesh that was initialy tagged as Hair due to the materials CRC colliding.

The solution is simply, if our material contains `Hair` or `Skin` we add it to the CRC for caching, this way they won't collide anymore.

Test:
You can repro e issue using `WITH_COLLECTIONS:urn:decentraland:ethereum:collections-v1:wonderzone_steampunk:`  in `master`. Just go to the avatar Editor, equip the steampunk hat first and then the jacket. You will the the hands colored as the hair. You can also refresh and equip first the jacket and then the hat, you will see the hair colored like the skin.

This shouldn't happen using this branch.


pravusjif (approved):
Fantastic!

unity-renderer/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -2320,6 +2320,10 @@ protected virtual IEnumerator ConstructMaterial(GLTFMaterial def, int materialIn
             for (int i = 0; i < 2; i++)
             {
                 string materialCRC = material[i].ComputeCRC().ToString();
+                if (def.Name.Contains("Hair"))
+                    materialCRC += "Hair";
+                else if (def.Name.Contains("Skin"))
+                    materialCRC += "Skin";
```
BrianAmadori, line 2326: Maybe we can disable the material cache altogether for avatars instead of doing this. It should be more clean.

In fact I forgot that I modified this line:
https://github.com/decentraland/unity-renderer/blob/203e0dfb8867af122688dae11dbee18f4988b061/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/WearableController.cs#L67
Formerly it was set as `CACHE_SHADERS`.

BrianAmadori: I modified the line because I thought that we copy the material before changing its color. Now I look at the code and it seems that i was mistaken. As `MapSharedMaterialsRecursively` is used directly on the mats.

AjimenezDCL: I didnt even try that because I didnt want to degrade the performance. I can profile both solutions to find degradation in performance.

Suduck: I don't have enough context but I agree with Brian. At least we should add a comment explaining this functionality. Using the gltf scene importer for something so specific to the avatars is an overkill imo

Suduck (approved):
GJ! I will approve and let Brian handle the thing in the GLTFimporter

unity-renderer/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
BrianAmadori: I think that replacing this fix for `CACHE_SHADERS` in `WearableController` shouldn't degrade performance significantly. With both approaches the outcome is to create a new Material. The only difference is that we will clone a few more materials, but I think is worth it as the solution would be more maintainable, and the gltf importer shouldn't know about the specifics of the world.

AjimenezDCL: Okay! gonna do it!

BrianAmadori (approved):

PR #1225, opened 2021-09-20
fix: notify kernel that DebugBridge is ready

pbosio:
`DebugBridge` is created at runtime. This means that kernel should wait for `DebugBridge` to be created before sending debug messages like those used when `PREVIEW` is initialized and/or using url params:
```
SetDebug()
SetSceneDebugPanel()
ShowFPSPanel()
SetEngineDebugPanel()
```

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/DebugController/DebugBridge.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/DebugController/DebugBridge.cs
@@ -15,6 +15,7 @@ namespace DCL
         public void Setup(IDebugController debugController)
         {
             this.debugController = debugController;
+            WebInterface.SetDebugBridgeReady();
         }
         
         // Beware this SetDebug() may be called before Awake() somehow...
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/WebInterface/Interface.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/WebInterface/Interface.cs
@@ -1298,5 +1298,10 @@ namespace DCL.Interface
 
             SendMessage("VideoProgressEvent", progressEvent);
         }
+
+        public static void SetDebugBridgeReady()
+        {
+            SendMessage("SetDebugBridgeReady");
+        }
     }
 }


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/DebugController/DebugBridge.cs
```diff
@@ -15,6 +15,7 @@ public class DebugBridge : MonoBehaviour
         public void Setup(IDebugController debugController)
         {
             this.debugController = debugController;
+            WebInterface.SetDebugBridgeReady();
```
BrianAmadori, line 18: The `DebugBridge` initialization is tied to the `KernelConfig` notification when the features are ready:

https://github.com/decentraland/unity-renderer/blob/940c2395583c4dc88e028549a8ab94c31f3f062a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/FeatureController/PluginSystem.cs#L24

This means that you can just wait for the feature flags to be ready from kernel side, and the bridge should be there. Doing is this way is better because:
- You can just close this PR and not make any changes Unity side
- You can avoid adding yet another kernel message for a very specific thing
- Not doing this avoids the precedent of having to create a new kernel message for every bridge. Take into account that we plan to create **all bridges** using this approach in the future to avoid tying the bridges to the Unity scene.

BrianAmadori (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/DebugController/DebugBridge.cs
pbosio: so what do you suggest for this scenario: https://github.com/decentraland/unity-renderer/blob/940c2395583c4dc88e028549a8ab94c31f3f062a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/FeatureController/PluginSystem.cs#L66
since creating the "feature" for scenario and enabling it using a url param for example will cripple a lot the client debugging. for example we won't be able to use any of those debug messages if we ran into an issue while walking around the world and we didn't have the "feature" enabled in the url

pbosio: on the other hand: `DebugPluginFeature` is always enabled but it is still not available when needed, so this kind of expose the situation that we will probably still need some kind of message, maybe a generic one with a feature id or something, to tell kernel that the bridge is ready in some scenarios

BrianAmadori: I don't understand your point.

1. `KernelConfig.i.EnsureConfigInitialized()` should be always triggered at Kernel initialization
2. This means that after Kernel sends the `KernelConfig` update message, the `DebugBridge` should be already in place.

So in this scenario, the `DebugBridge` is always present after the KernelConfig is updated (because its forced to true in this case).

In any other cases it will not matter, because we will set the feature as true in `KernelConfig.Features` when the URL param is present, send the message, and the Bridge will be there.

pbosio: On kernel side it something like this:
1- initialize unity
2- send kernel configuration
3- send messages
so there is nothing in between 2 and 3 where we can say "wait for this... ok! you can now start to send those messages"
what I mean is that the reaction on unity after kernel send the kernel config is not immediate, so we need something to signal when it can start sending the debug messages

PR #1230, opened 2021-09-21
fix: combined avatar mesh leak

BrianAmadori:
## What does this PR change?

* Combined avatar meshes were leaking because the renderer used to retrieve the mesh references for unload was being destroyed before the unload was called.
* Fix unhandled error when the face wearables can't be fetched from wearables endpoint.

## How to test the changes?

* Instance many avatars by using `clientDebug.InstantiateBotsAtCoords({ amount: 200 })`
* Clear the bots by using `clientDebug.ClearBots()`.

Notice that when doing this in `master` a lot of dangling meshes can be seen on Unity's profiler, incurring a leak of around 2.4 mb per avatar.

On the version of this PR, the meshes should be destroyed and the memory freed.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
@@ -424,10 +424,6 @@ namespace DCL
             // TODO(Brian): Evaluate using UniTask<T> instead of this way.
             yield return new WaitUntil(() => bodyShapeController.isReady && wearableControllers.Values.All(x => x.isReady));
 
-            if ( eyesController == null || eyebrowsController == null || mouthController == null )
-            {
-            }
-
             eyesController.Load(bodyShapeController, model.eyeColor);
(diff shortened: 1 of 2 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
```diff
@@ -415,6 +424,10 @@ private IEnumerator LoadAvatar()
             // TODO(Brian): Evaluate using UniTask<T> instead of this way.
             yield return new WaitUntil(() => bodyShapeController.isReady && wearableControllers.Values.All(x => x.isReady));
 
+            if ( eyesController == null || eyebrowsController == null || mouthController == null )
+            {
+            }
```
Kinerius, line 429: You might want to delete this

Kinerius (approved):
Looking good, there is an unintended if with nothing in it.

pravusjif (approved):
Tested with no problems! let's just address Kinerius comment

PR #1232, opened 2021-09-21
fix: invisible UI container blocking raycast

pbosio:
related to https://github.com/decentraland/sdk/issues/75

It's not uncommon for scenes developers to leave UIContainers on screen since by default it alpha is 0 so they don't see it.
This generate issues like hover feedback not being display or not being able to lock the mouse.
So I've added an alpha tolerance to decide if a UIContainer should be a target for raycast or not.

issue can be reproduced here: https://play.decentraland.zone/?position=-29,55 trying to lock the mouse on the bottom-right corner of the screen is impossible cause an invisible UIContainer is blocking it
test issue is fixed here: https://play.decentraland.zone/?renderer-branch=fix/invisible-ui-container-blocking-raycast&position=-29,55

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/UI/UIShape.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/UI/UIShape.cs
@@ -54,7 +54,7 @@ namespace DCL.Components
         where ReferencesContainerType : UIReferencesContainer
         where ModelType : UIShape.Model
     {
-        public const float RAYCAST_ALPHA_THRESHOLD = 0.02f;
+        public const float RAYCAST_ALPHA_THRESHOLD = 0.01f;
         
         public UIShape() { }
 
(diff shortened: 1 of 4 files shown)


unity-renderer/Assets/Scripts/Tests/CursorControllerTests/CursorControllerTests.cs
```diff
@@ -286,5 +286,66 @@ public IEnumerator OnPointerHoverFeedbackIsBlockedByUI()
 
             DCLCharacterController.i.ResumeGravity();
         }
+        
+        [UnityTest]
+        public IEnumerator OnPointerHoverFeedbackIsNotBlockedByFullyAplhaUIContainer()
```
BrianAmadori, line 291:
```suggestion
        public IEnumerator OnPointerHoverFeedbackIsNotBlockedByFullyAlphaUIContainer()
```

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/UI/UIShape.cs
```diff
@@ -54,6 +54,8 @@ public class UIShape<ReferencesContainerType, ModelType> : UIShape
         where ReferencesContainerType : UIReferencesContainer
         where ModelType : UIShape.Model
     {
+        public const float RAYCAST_ALPHA_THRESHOLD = 0.02f;
```
BrianAmadori, line 57:
```suggestion
        public const float RAYCAST_ALPHA_THRESHOLD = 0.01f;
```
Why not 0.01f?

BrianAmadori (approved):

PR #1233, opened 2021-09-21
fix: mesh and material leaks on avatar lods

BrianAmadori:
## What does this PR change?

Material and meshes were being accessed using the `material` and `mesh` property instead of `sharedMaterial` and `sharedMesh`. This caused small material/mesh leaks on avatars.

## How to test the changes?

The easiest way to test is to add/remove 100 bots from Unity and check the memory details on the profiler. The material/mesh count as shown by the profiler should go back to the original values.

## Our Code Review Standards

-----
Wondermine testing: https://play.decentraland.zone/?renderer-branch=fix/mesh-and-material-leaks-on-avatar-lods&ENABLE_AVATAR_LODS&MAX_VISIBLE_PEERS=200&position=-28%2C55

Stadium testing: https://play.decentraland.zone/?renderer-branch=fix/mesh-and-material-leaks-on-avatar-lods&ENABLE_AVATAR_LODS&MAX_VISIBLE_PEERS=200&position=91,151&NETWORK=ropsten


pravusjif (changes requested):
Using the non-shared properties was actually on purpose, because if we use them, we can't individually tint the materials nor update the meshes:

https://user-images.githubusercontent.com/1031741/134284171-d4f2d19a-b80c-4719-a9f1-275a2d126ed1.mov

https://user-images.githubusercontent.com/1031741/134284231-f302a636-6392-466c-ba93-04ff46db3f89.mov

I guess we could move this functionality to the shader and then use everything with the shared properties?

BrianAmadori:
@pravusjif We shouldn't use `material` (even if intended) as the materials created by this way **are never destroyed** unless this is done explicitly. The property can be used and then the material destroyed, but this results in very opaque code difficult to understand. I think we can fix this by cloning/destroying the material at the start/end of the `AvatarRenderer`. Going to aim for that.

pravusjif:
> @pravusjif We shouldn't use `material` (even if intended) as the materials created by this way **are never destroyed** unless this is done explicitly. The property can be used and then the material destroyed, but this results in very opaque code difficult to understand. I think we can fix this by cloning/destroying the material at the start/end of the `AvatarRenderer`. Going to aim for that.

Yes totally, I must've forgotten to destroy those materials in my implementation, I agree that adding that missing destruction will fix the leak

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
```diff
@@ -132,23 +135,38 @@ void onFailWrapper(bool isFatalError)
...
-                AvatarRendererHelpers.RandomizeAndApplyGenericImpostor(lodMeshFilter.mesh);
+                AvatarRendererHelpers.RandomizeAndApplyGenericImpostor(impostorMesh);
             }
         }
 
+        public void CleanImpostor()
+        {
+            impostorInitialized = false;
+            Destroy(impostorMesh);
+            Destroy(impostorMaterial);
```
AjimenezDCL, line 167: I have 0 trust in Avatar Renderer. Should we add a `null` check here?

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
```diff
@@ -113,11 +113,11 @@ public override IEnumerator ApplyChanges(BaseModel newModel)
             EnablePassport();
 
             KernelConfig.i.EnsureConfigInitialized()
-                        .Then(config =>
-                        {
-                            if(config.features.enableAvatarLODs)
-                                avatarRenderer.InitializeImpostor();
-                        });
+                .Then(config =>
```
AjimenezDCL, line 116: Isn't this prone to race conditions because kernel initially sents the `enableAvatarLODs` as false until it reads the feature flag?

AjimenezDCL (approved):
Left a couple of concerns

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
BrianAmadori: I didn't put a null check because `Destroy` doesn't do anything will null arguments.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
BrianAmadori: If that happens, the `Destroy()` will be called anyways but the resources will be null and nothing will happen. (`Destroy` don't do anything with null arguments).

AjimenezDCL: I'm worried about:
1) Kernel is retrieving FeatureFlags, sends -> `enableAvatarLODs = false` in the meantime.
2) We go trough the `EnsureInitialization`
3) `avatarRenderer.InitializeImpostor()` is NOT called.
4) Kernel retrieves FeatureFlags, sends -> `enableAvatarLODs = true`.
5) AvatarShape doesn't react to it.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarRenderer.cs
```diff
@@ -132,23 +135,38 @@ void onFailWrapper(bool isFatalError)
 
         public void InitializeImpostor()
         {
+            if (impostorInitialized)
+                return;
```
pravusjif, line 139: Apparently the fetched users snapshot is not being applied to the impostor, I suspect this escape check could be the culprit, do you want me to give it a try at fixing it?

https://user-images.githubusercontent.com/1031741/134537206-4fd0de95-3177-47cd-aee3-7de6cd82f4f4.mov

pravusjif (changes requested):

PR #1239, opened 2021-09-23
fix: animation stuck fix

### LINKED ISSUE #985: NPC animations get stuck in a loop
Sometimes an entity with multiple animations gets stuck playing an animation in a loop. When this happens, you can't call the entity to play any other animations, even when you interact with the entity.

This seems to happen mostly in large scenes. It's not deterministic, in all the places where I see it happen it's just sometimes. Might be more common on slower machines.

In genesis plaza I see it a lot, the bar tender gets stuck quite often in a short animation. He should also switch animations when you talk to him, but he's stuck in a loop. There are also other places where I see this less often.

I think every time I saw this, it was with entities that aren't loaded with the initial load of the scene. In this case, the bar tender is only loaded when you go downstairs.

This happens both in preview and in production.

My theory is that the entity is never getting the Animator component loaded, and the default behavior when no animator is present is to loop the first animation.

https://images.zenhubusercontent.com/337227404/c67b8ff0-ac1d-4c57-816e-8455bdc77df3/screen_recording_.mov

Suduck:
## What does this PR change?

Fixes #985  

The current animation system has a race condition where if you try to set the animation in the middle of the charge of a GLTF the animator component will say that the animation that you are trying to play is playing but it is not playing.

As the animator component will say that the animation is playing, the DCLAnimator component will ignore the message

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=fix/animation-problem
2. Test several times that the octopus is doing the animation correctly

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Animator/DCLAnimator.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Animator/DCLAnimator.cs
@@ -47,8 +47,8 @@ namespace DCL.Components
 
         public override IEnumerator ApplyChanges(BaseModel model)
         {
-            entity.OnShapeUpdated -= OnComponentUpdated;
-            entity.OnShapeUpdated += OnComponentUpdated;
+            entity.OnShapeLoaded -= OnComponentUpdated;
+            entity.OnShapeLoaded += OnComponentUpdated;
 
             UpdateAnimationState();
 
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_NFT.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_NFT.cs
@@ -62,6 +62,7 @@ namespace DCL.Components
             alreadyLoaded = true;
 
             entity.OnShapeUpdated?.Invoke(entity);
+            entity.OnShapeLoaded?.Invoke(entity);
         }
 
         void LoadAsset() { loaderController?.LoadAsset(assetUrl, true); }
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_OBJ.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_OBJ.cs
@@ -54,6 +54,7 @@ namespace DCL.Components
             {
                 entity.OnShapeUpdated.Invoke(entity);
             }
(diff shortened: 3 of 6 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Animator/DCLAnimator.cs
```diff
@@ -47,8 +47,8 @@ public class DCLAnimationState
 
         public override IEnumerator ApplyChanges(BaseModel model)
         {
-            entity.OnShapeUpdated -= OnComponentUpdated;
-            entity.OnShapeUpdated += OnComponentUpdated;
+            entity.OnShapeLoaded -= OnComponentUpdated;
```
BrianAmadori, line 50:
```suggestion
            entity.OnShapeLoaded -= OnShapeLoaded;
```
I think we can safely rename the method to `OnShapeLoaded`

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Animator/DCLAnimator.cs
```diff
@@ -47,8 +47,8 @@ public class DCLAnimationState
```
BrianAmadori, line 47: You forgot to unsubscribe from the proper method (can't reference the code because is not included in the changeset).

This is in the current code:
`private void OnDestroy() { entity.OnShapeUpdated -= OnComponentUpdated; }`

It should be:
`private void OnDestroy() { entity.OnShapeLoaded -= OnComponentUpdated; }`

BrianAmadori (commented):
Didn't try to reproduce the bug, but found a few things.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Animator/DCLAnimator.cs
Suduck: Good catch!

BrianAmadori (approved):

AjimenezDCL (approved):
What's the difference between `OnShapeUpdate` and `OnShapeLoad`? Do we need them both?

Approving anyway.

Suduck:
> What's the difference between `OnShapeUpdate` and `OnShapeLoad`? Do we need them both?
> 
> Approving anyway.

Each time a shape has a change the OnShapeUpdate is called , where OnShapeLoad is only called when the shape is completely loaded.

When we are loading the shape the OnShapeUpdate is called more than two times  ( and the amount of times change if we are using GLTF and AB).

Some components don't care about the loading part and only want to be notified when the shape is fully loaded ( like de DCLAnimator). This way we ensure that there are no race conditions happening because we try to start something before it is fully loaded

AjimenezDCL (approved):

nearnshaw:
Thanks @Suduck  ❤️ 

This bug had been bugging me for so long, really happy that it's fixed

Issue #987, opened 2021-08-13
Implement wearables AB conversion in batch mode for automation


pravusjif:
In order to **automate wearables Asset Bundle conversion every time a wearable is approved and created** we should be able to call Unity in batch mode, providing the **wearable's urn** as a parameter and **spitting the converted Asset Bundle in a local directory**.

That way we can hook that process in a new automated CI, as suggested by @menduz

PR #1276, opened 2021-09-28
fix: non bodyshape wearables local AB conversion

pravusjif:
updated wearables dump-convert flow to run on every collection sequentially


AjimenezDCL (approved):

unity-renderer/Assets/ABConverter/Core.cs
```diff
@@ -802,12 +802,12 @@ private void MarkShaderAssetBundle()
             ABConverter.Utils.MarkAssetForAssetBundleBuild(env.assetDatabase, mainShader, MAIN_SHADER_AB_NAME);
         }
 
-        internal virtual void InitializeDirectoryPaths(bool deleteIfExists)
+        internal virtual void InitializeDirectoryPaths(bool deleteDownloadDirIfExists, bool deleteABsDireIfExists)
```
BrianAmadori, line 805:
```suggestion
        internal virtual void InitializeDirectoryPaths(bool deleteDownloadDirIfExists, bool deleteABsDireIfExists)
```
This signature change is completely justified? I'm looking at the code and observing that both arguments are called always with the same value.

BrianAmadori (changes requested):

unity-renderer/Assets/ABConverter/Core.cs
pravusjif: Check [this line](https://github.com/decentraland/unity-renderer/blob/a9ed5d9f6dde043ae85c5ef31db412ef359db1f8/unity-renderer/Assets/ABConverter/Client.cs#L532), we need to avoid deleting the ABs folder on every `DumpWearablesCollection()` call to keep the ABs from the previously-converted collections. Otherwise, after dumping all the collections sequentially, we end up with only the ABs from the latest collection in the output directory.

unity-renderer/Assets/ABConverter/Client.cs
```diff
@@ -204,6 +191,70 @@ public static void ExportSceneToAssetBundles(string[] commandLineArgs)
...
+            //NOTE(Brian): This should make the logs cleaner
+            Application.SetStackTraceLogType(LogType.Log, StackTraceLogType.None);
+            Application.SetStackTraceLogType(LogType.Warning, StackTraceLogType.None);
+            Application.SetStackTraceLogType(LogType.Error, StackTraceLogType.Full);
+            Application.SetStackTraceLogType(LogType.Exception, StackTraceLogType.Full);
+            Application.SetStackTraceLogType(LogType.Assert, StackTraceLogType.Full);
+            
+            EnsureEnvironment();
+            ExportWearablesCollectionToAssetBundles(System.Environment.GetCommandLineArgs());
+        }*/
```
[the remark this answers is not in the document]
pravusjif: Uncommented it as we will use that when we continue with the automated-wearable-collections convertion on collections deployment. Also added this info in [the pertinent issue](https://github.com/decentraland/unity-renderer/issues/987)

unity-renderer/Assets/ABConverter/Client.cs
```diff
@@ -369,11 +420,11 @@ public static Core.State DumpAsset(string assetHash, string assetFilename, strin
         /// <summary>
         /// Dump all bodyshape wearables normally, including their imported skeleton 
         /// </summary>
-        public static void DumpAllBodiesWearables(Func<string, bool> collectionFilter = null)
-        {
+        public static void DumpAllBodyshapeWearables()
```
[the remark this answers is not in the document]
pravusjif: I separated the `Client.s` into `ClientSettings.cs`, `SceneClient.cs` and `WearablesCollectionClient.cs` and tested that everything keeps working (even batch mode conversion).

unity-renderer/Assets/ABConverter/WearablesCollectionClient.cs
```diff
@@ -0,0 +1,347 @@
...
+            });
+        }
+
+        /// <summary>
+        /// Dump all non-bodyshape wearables, optimized to remove the skeleton for the wearables ABs since that is
+        /// only needed for the body shapes (and the WearablesController sets it up for non-bodyshapes in runtime).
+        /// Each collection is dumped and converted sequentially as the amount of collections has grown more that we
+        /// can handle in 1 massive dump-conversion.
+        /// </summary>
+        public static void DumpAllNonBodyshapeWearables(Func<string, bool> collectionFilter = null)
```
BrianAmadori, line 168: If we aren't using the `collectionFilter` argument maybe it's a good idea to just remove it?

BrianAmadori (approved):
🎉

unity-renderer/Assets/ABConverter/Client.cs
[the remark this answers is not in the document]
BrianAmadori: Cool!

unity-renderer/Assets/ABConverter/WearablesCollectionClient.cs
pravusjif: You are right, since now we have the other `DumpSingleWearablesCollection()` for that. I'll remove the filter!

PR #1279, opened 2021-09-29
refactor: move notification model to it own assembly

pbosio:
* Moved Notification's Model to it own assembly definition
* Created `INotification` interface

This way we can use the notification system without the need to reference `NotificationHUD`


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/NotificationHUD/Model/NotificationModel.cs
```diff
@@ -0,0 +1,31 @@
+using System;
+
+namespace NotificationModel
```
BrianAmadori, line 3:
```suggestion
namespace DCL.NotificationModel
```

BrianAmadori (approved):
❤

pravusjif (approved):

PR #1290, opened 2021-10-04
feat: update API call for builder in world

### LINKED ISSUE #1250: BIW: Update assets API to the new format
They have changed the way we use the API to cache it with CloudFlare in order to made the changes we need to do this

Lautaro
> Hi Adrian! The server with the change we’ve mentioned is live now! We went with an approach were we didn’t break API. We added the owner query parameter to the /v1/assetpacks endpoint. Whenever the endpoint is requested using /v1/assetpacks?owner=default it will return the default asset packs, giving us the opportunity to cache them at the cloudfront level, and whenever it gets requested with /v1/assetpacks?owner=usersWalletAddress it will return the user’s asset packs.
> To summarize the change, you now need to do two queries, that can be done in parallel:
> /v1/assetpacks?owner=default
> /v1/assetpacks?owner=usersWalletAddress
> and then you’ll need to merge the result of those two requests to have the same set of asset packs that you had before.
> Here’s an example on how we did it in the builder https://github.com/decentraland/builder/pull/1579/files

Suduck:
## What does this PR change?

Fixes #1250   

It refactors the way the catalog is asked to the builder API to split the call in two parts

## How to test the changes?

1. Go to: https://play.decentraland.zone/?kernel-branch=feat/add-new-message&renderer-branch=feat/update-assets-api
2. Enter builder in world and check that you have the catalog

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Configuration/Configuration.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Configuration/Configuration.cs
@@ -6,7 +6,7 @@ namespace DCL.Configuration
     {
         //Note: Don't use these URL directly, you need to get them in the BuilderInWorldUtils to take into account the ENV
         public const string BASE_URL_SCENE_OBJECT_CONTENT = "https://builder-api.decentraland.{ENV}/v1/storage/contents/";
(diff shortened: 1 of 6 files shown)


pravusjif (approved):
Worked fine for me! tested changing one of my lands in prod and published the changes with no problems.

PR #1292, opened 2021-10-04
feat: added scrollbar to dropdown menus

Kinerius:
## What does this PR change?

Adds scrollbar support to dropdown menus

## How to test the changes?

Check resolution size option here: https://github.com/decentraland/explorer-desktop/pull/12

## Our Code Review Standards


BrianAmadori (approved):
Did not find any issue with the existing dropdowns. Kind of a blind approve because we don't have any setting that has enough options to make the scollbar show up.

PR #1307, opened 2021-10-06
[Explore V2] Places sub-section implementation (UI)

sandrade-dcl:
## What does this PR change?
Create the Places sub-section in the Explore V2 feature.

**IMPORTANT**: _This PR does not include test coverage, it will be addressed in a separate PR._

## Related links
- **PRD "Explore V2"**: https://www.notion.so/decentraland/PRD-Explore-V2-ab98e922c196488fb217596c888ce3d6
- **Figma "Explore V2"**: https://www.figma.com/file/5YhRkm1NXme8MhgDxGT2zY/Explore-V2?node-id=1290%3A4164

## Test Link
https://play.decentraland.zone/?renderer-branch=feat/ExploreV2-Places-Sub-Section-UI&position=100,100

## Our Code Review Standards


sandrade-dcl:
> I see that we use auto-resize for the titles, I think that this leads to inconsistent design. We can use word-wrap with static size for more readable and prettier overall UI

Although this PR does not yet include the last visual refinements (it is still in WIP), you are right. I have changed the configuration of these texts to make them fixed in size and wrapped if the text dosn't fit 👍 Thanks for the notice!

> Also I see that the numbers for user amount have a very small font

Fixed!

> When hovering out of any place, there's no fade out animation. I think we should ensure all the elements transition with an animation if possible.

This hovering has already a fade in/out animation, but you may not notice it because the speed of the animation is a bit high. Anyway all these kind of transitions will be designed with Romi and Stefi in a next step. For now I only left a simple hovering animation as example.

> Also I'm noting a few very noticeable hiccups the first time I open the `Explore` window. However for the next times the hiccups aren't present, this causes me two concerns:
> 
> * We shouldn't have hiccups in the first place
> * If the second time there's not a hiccup, this means the places textures are not unloaded? We should always unload the textures.

I think the hiccups you're noting are related with the own instantiation of each item. For example, in the case of the Places section where there are a lot of cards to load (more then 60), they are all being instantiated at the same time each time the window is opened. I already had planned in my next steps to implement some things to improve the performance: for example apply object pooling for both "Places" and "Events" sections, maybe also we could apply some kind of throttling during the cards loading, etc.. Anyway it would be fine to have a conversation with you about these kind of solutions (@BrianAmadori).

About the loading/unloading textures... my `Image` UI component is using the `LazyTextureObserver` to set the images, so the unloading of textures is already being managed by it in each game object destruction.

sandrade-dcl:
> * The main explore & play panel seems to miss the fade-in/fade-out animations.

Good catch! already fixed 👍 

> * The carrousel dots seem to be too close to the main panel.

I have separated it a bit. Anyway it will be reviewed with Romi in the next "visual refinement" stage.

> * The mouse wheel movement is just too slow on windows + chrome, we may have to devise some acceleration logic to improve this later? Or maybe we can add a bit of speed without breaking it for osx?

Ok, I will be reviewing it with Romi in order to set a "good" configuration.

> * I think we should open the "info" panel (or a desired default action) by just clicking on the card, instead of doing it only when the user clicks on the button. I tended to always click on the card image intuitively and nothing happened. This felt a bit awkward.

I will pass this feedback to Romi, thanks Brian!

PR #1352, opened 2021-10-18
fix: flaky test fix

### LINKED ISSUE #943: Fix flaky test
These test are failing randomly, we should fix them

- [x] AudioIsLooped
 

- [x] ForgetIsCalledWhileAssetIsBeingLoaded 

- [x] FeedbackIsNotDisplayedOnParent

The following test are using local resources but they are still making the call to the API. We should ensure that they don't call the the API so in case that the API doesn't work, the test will continue working

- [x] NFTShapeIsInvalidatedWhenStartingOutOfBounds
- [x] NFTShapeIsInvalidatedWhenLeavingBounds
- [x] NFTShapeIsResetWhenReenteringBounds

- [x] NFTShapeIsInvalidatedWhenStartingOutOfBoundsDebugMode
- [x] NFTShapeIsInvalidatedWhenLeavingBoundsDebugMode
- [x] NFTShapeIsResetWhenReenteringBoundsDebugMode

Suduck:
## What does this PR change?

Fixes #943 
This PR is aimed to fix the flaky test that we have right now

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldAudioHandler.cs
```diff
@@ -89,7 +88,12 @@ public void EnterEditMode(ParcelScene scene)
         UpdateEntityCount();
 
         if (eventBuilderMusic.source.gameObject.activeSelf)
-            startBuilderMusicCoroutine = StartCoroutine(StartBuilderMusic());
+        {
+            if (startBuilderMusicCoroutine != null)
+                CoroutineStarter.Stop(startBuilderMusicCoroutine);
+
+            startBuilderMusicCoroutine = CoroutineStarter.Start(StartBuilderMusic());
```
AjimenezDCL, line 95: this is a monobehaviour, why are we using `CoroutineStarter`?

unity-renderer/Assets/Scripts/Tests/CursorControllerTests/CursorControllerTests.cs
```diff
@@ -220,6 +220,8 @@ public IEnumerator FeedbackIsNotDisplayedOnParent()
             DCLCharacterController.i.SetPosition(new Vector3(3, 2, 12));
             cameraController.SetRotation(0, 0, 0, new Vector3(0, 0, -1));
 
+            yield return null;
+            yield return null;
```
AjimenezDCL, line 224: This smells fishy, why do we need to wait for 3 frames?

unity-renderer/Assets/UnityGLTF/Scripts/GLTFComponent.cs
```diff
@@ -159,13 +159,12 @@ void ApplySettings(Settings settings)
...
         {
             if (state == State.FAILED)
-            {
                 return;
-            }
 
             state = State.FAILED;
 
-            CoroutineStarter.Stop(loadingRoutine);
+            if (loadingRoutine != null)
```
AjimenezDCL, line 166: This is actually not neeed, `CoroutineStarter` handles `null` references gracefully.

AjimenezDCL (approved):
Approving with small requests, good job!

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BuilderMode/BuilderInWorldAudioHandler.cs
BrianAmadori: I think I introduced this change to avoid issues with the disabled `MonoBehaviour` not being able to start coroutines.

unity-renderer/Assets/Scripts/Tests/CursorControllerTests/CursorControllerTests.cs
Suduck: I don't remember exactly, but IIRC it was a race condition where sometimes it took more than 1 frame to change the cursor sprite, so I set two more nulls to ensure that the race condition doesn't happen anymore

Suduck: I will add a comment for the next time to this kind of approach so we can see in the future why are we doing this

unity-renderer/Assets/UnityGLTF/Scripts/GLTFComponent.cs
Suduck: Changed!

PR #1382, opened 2021-10-21
Fix/input lock

Kinerius:
## What does this PR change?

Increases the tolerance on input spikes after locking the cursor to move the camera again, to avoid locking input on low FPS

## Our Code Review Standards


Kinerius:
woops used wrong base branch

PR #1386, opened 2021-10-22
GLTFs ON RED COLOR

Suduck:
https://play.decentraland.zone/?renderer-branch=feat/ab-view&position=-62,80


olavra:
I think we should include this as a command in console? Sometimes is really useful to enable/disable it directly when you see something weird happening that can be gone if you reload

AjimenezDCL:
Closing this because we already have this: https://github.com/decentraland/unity-renderer/pull/2058

PR #1417, opened 2021-10-27
chore: cleanup and refactor InitialScene

### LINKED ISSUE #1258: Make simpler to sync the `InitialScene` from `unity-renderer` with `MainScene` from `explorer-desktop`
When we change something in `InitialScene` from `unity-renderer` we need to do the exact change in `explorer-desktop`.

To avoid that, @Kinerius suggested creating a `Prefab` that includes everything that the `InitialScene` has and just having a `Scene` with only one prefab.

Kinerius:
## What does this PR change?

All game object dependencies from `InitialScene` are now prefabs that are being instantiated on Awake by `Main`.
Some systems got a slight refactor as now there is no need for them to be a MonoBehaviour.

Fix #1258

## How to test the changes?

There are no behavioural changes so everything should work as it always did.

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Main/Main.cs
```diff
@@ -87,7 +82,19 @@ protected virtual void SetupEnvironment()
 
         protected virtual HUDContext HUDContextBuilder() { return HUDContextFactory.CreateDefault(); }
 
-        private void Start() { Environment.i.world.sceneController.Start(); }
+        private void Start()
```
menduz, line 85: Is this still called after Awake()?

Kinerius: Yes, now that we instantiate new objects on `Awake` we have to wait a bit before they get to initialize and exist, then the `Start` is invoked and everything should work like before

menduz: 👍

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Main/Main.cs
```diff
@@ -112,5 +119,40 @@ protected virtual void OnDestroy()
...
+            var mouseCatcher = Init("MouseCatcher").GetComponent<MouseCatcher>();
+            var environment = Init("Environment").GetComponent<EnvironmentReferences>();
+            var playerReferences = Init("Player").GetComponent<PlayerReferences>();
+
+            Init("HUDController");
+            Init("HUDAudioHandler");
+            Init("NavMap");
+            Init("SettingsController");
+
+            var sceneReferences = new SceneReferences(
```
BrianAmadori, line 134: I see that this is initialized as a local instance, and the `SceneReferences` class fills out the static instance variable in its constructor. This escapes the usual idiom in which we do singletons and makes the singleton idiom pretty obscure. 

I think the intent is clearer if we just make `SceneReferences` inherit `Singleton<T>` and fill the references by just writing public fields or through a public `Initialize()` method.

BrianAmadori (changes requested):

Suduck (approved):
I am thinking that maybe since now it is an instance and it's loaded from the beginning of the app. Should we move it to the DataStore? This way we can remove the singletons. WDYT @BrianAmadori ?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Main/Main.cs
Kinerius: Done.

BrianAmadori (approved):

PR #1420, opened 2021-10-28
fix: videotexture not playing in some cases

### LINKED ISSUE #1415: Video texture not playing
This simple scene with two video screens works fine on preview, using @next, but not in production.

[https://app.zenhub.com/files/389651442/e49a3b34-083a-4c31-b2dd-76e9805c16fe/download](https://app.zenhub.com/files/389651442/e49a3b34-083a-4c31-b2dd-76e9805c16fe/download)

Deployed here:

https://play.decentraland.org/?NETWORK=mainnet&island=Ik04z&position=-133%2C-42&realm=loki

One of the screens starts the video when the scene loads, the other one when the screen is clicked.
In preview both work fine.
In production, the one that loads from the start never loads. The one that starts when clicked works around 50% of the time. If you click it several times it will often fail. There seems to be some kind of race condition there.

This is the video being used:

its size is 512 × 512, so shouldn't be capped by the video size limit

This issue was introduced in the last couple of weeks, it wasn't present before. Some previously working scenes have stopped working because of this.

Here's a second example that seems to be the same, reported by another creator
This time using smart items that used to work until before last week

scene location:
https://play.decentraland.org/?island=Ik2mv&position=-74%2C-99&realm=loki

scene code:
[https://app.zenhub.com/files/337227404/2aba97a4-5154-4156-bcb4-836bb3d9add5/download](https://app.zenhub.com/files/337227404/2aba97a4-5154-4156-bcb4-836bb3d9add5/download)

video:
https://youtu.be/W6ArVsdOX8o

Should Playing video be as touchy as I am experiencing.  If I have a list of video urls to play and pick one in world via button click i can see the Network tab fetching it and hear sound but it no longer plays the video just the sound.  This was working maybe a week ago.  Works just fine locally.  I have tried all sorts of things like not playing the video for a few seconds so it can buffer, remaking the material etc.  At one point, all I had to do was resize my browser window and the  play video would show up on the plane (I assume the resize caused it to redraw).  The symptoms and the conditions it does not show video are nothing I have been able to consistently reproduce. yet.

AjimenezDCL:
Fixes #1415 

The bug was being caused due to calling `Play` before the video was ready. This could happen randomly but not always because in some cases the ready state was set immediately, but was not guaranteed.

Also:
* Re-introduced video texture plugin optimization
* Removed `isTest` anti-pattern in `DCLVideoTexture`
* Introduced special mock class to be able to test the plugin logic without WebGL.
* Renamed all tests to be compatible with the `Should` naming scheme.


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/Tests/WebVideoPlayerShould.cs
```diff
@@ -155,68 +191,22 @@ public void OnVideoErrorPlayerWontDoAnything()
...
-            webVideoPlayer.UpdateWebVideoTexture();
-            webVideoPlayer.UpdateWebVideoTexture();
-
-            Assert.IsTrue(timesTextureWasReady == 1);
-            Assert.AreEqual(32, texture.width);
-            Assert.AreEqual(64, texture.height);
-        }
-
-        [Test]
-        public void OnVideoPlayingResizeTextureAndUpdate()
```
BrianAmadori, line 180: This was removed because texture resizing is not needed.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
```diff
@@ -66,7 +65,10 @@ public override IEnumerator ApplyChanges(BaseModel newModel)
             //If the scene creates and destroy the component before our renderer has been turned on bad things happen!
             //TODO: Analyze if we can catch this upstream and stop the IEnumerator
             if (isDisposed)
+            {
+                var toDebug = this.model as Model;
```
pbosio, line 69: is there any porpoise for that or did you forget to remove it after debugging?

pbosio (approved):
LGMT

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
BrianAmadori: It was a mistake, already removed it, ty!.

Kinerius (approved):
Nice!

PR #1427, opened 2021-10-29
fix: renderer caching in preview mode

pbosio:


CodeJuggernaut (approved):

BrianAmadori:
@menduz Can you look into this? This is supposed to avoid re-downloading the client in preview mode, but there isn't the possibility of producing cache invalidation issues for explorer?

menduz:
@pbosio could you clarify what problem is this PR intending to fix? I'm having a hard time trying to decipher the changes

menduz (approved):
Oh I think I understand now. Looks good

PR #1432, opened 2021-11-01
test: colliders dont work production

AjimenezDCL:


AjimenezDCL:
We discovered that skewed transform were breaking the baking of physics. https://forum.unity.com/threads/physics-bakemesh-question.741257/

PR #1453, opened 2021-11-03
style: explore v2

RominaMarchetti:
This branch includes the Explore V2 extreme makeover look and feel.

image
image

Notes:
- By default Event's section is opened first.
- Players and friends weren't removed from the place's card, they are visible on mouse hover.
- The Going events section will be temporary off, that's why the "want to go" button on the event's card opens the event's link.

Test link: https://play.decentraland.zone/?renderer-branch=style/ExploreV2UIPolish


sandrade-dcl (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/ExploreV2/Scripts/Sections/PlacesAndEventsSection/PlacesAndEventsSectionComponentView.cs
```diff
@@ -40,18 +40,14 @@ public override void Dispose()
 
     internal void CreateSubSectionSelectorMappings()
     {
-        subSectionSelector.RefreshControl();
         subSectionSelector.GetSection(0)?.onSelect.AddListener((isOn) => placesSubSection.gameObject.SetActive(isOn));
         subSectionSelector.GetSection(1)?.onSelect.AddListener((isOn) => eventsSubSection.gameObject.SetActive(isOn));
```
Suduck, line 44: We should identified with constant or enums which section represent each number for clarify

unity-renderer/Assets/UIComponents/Scripts/Components/SectionSelector/SectionToggle.cs
```diff
@@ -41,20 +43,25 @@ public interface ISectionToggle
...
+    [Header("Visual Configuration When Unselected")]
+    [SerializeField] private ColorBlock backgroundTransitionColorsForUnselected;
     [SerializeField] private Color unselectedTextColor;
     [SerializeField] private Color unselectedImageColor;
 
     public ToggleEvent onSelect => toggle?.onValueChanged;
 
     private void Awake() { ConfigureDefaultOnSelectAction(); }
 
+    private void OnEnable() { StartCoroutine(ForceToRefreshToggleState()); }
```
Suduck, line 63: We should use Coroutine Starter here!

Suduck (approved):
Good job! just minor request that you can address in another PR if necessary

unity-renderer/Assets/UIComponents/Scripts/Components/SectionSelector/SectionToggle.cs
AjimenezDCL: I disagree here.

`CoroutineStarter` has no benefit whatsoever besides allow you to start coroutines in a no Monobehaviour class. In this case this is just fine.

In fact using the `CoroutineStarter` has the downside of obscuring profiling. If you take a look at the profiler you will see that the `CoroutineStarter` is taking lot of time in the `MoveNext` operation of the `IEnumerators` and there's no way to know which ones are taking most of the time.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/ExploreV2/Scripts/Sections/PlacesAndEventsSection/PlacesAndEventsSectionComponentView.cs
sandrade-dcl: Done!

unity-renderer/Assets/UIComponents/Scripts/Components/SectionSelector/SectionToggle.cs
AjimenezDCL: Furthermore the `CoroutineStarter` would add you some tiresome boilerplate code. If your monobehaviour starts a coroutine, the lifecycle is attached to the monobehaviour. Disabling or removing a Monobehaviour stops and removes its running coroutines.

If you do it in a CoroutineStarter you might end up forgetting to stop the routine and it will produce a leak or some undesirable issues.

PR #1464, opened 2021-11-04
fix: now memory manager is parametrizable and the memory threshhold i…

kuruk-mm:
…s a long variable

Related with #1459

## What does this PR change?

MemoryManager can be instantiated with two parameters, but those parameters weren't being taken into account. So this PR fixed that, and the variable was a `uint` and you couldn't set a memory larger than 4gb for clean up. So I changed it to `long` (64 bits variable).

This fix doesn't have any effect on `unity-renderer` but it's useful for `explorer-desktop`.

## How to test the changes?

2. ...

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/MemoryManager/MemoryManager.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/MemoryManager/MemoryManager.cs
@@ -7,20 +7,20 @@ namespace DCL
 {
     public class MemoryManager : IMemoryManager
     {
-        private const uint MAX_USED_MEMORY = 1300 * 1024 * 1024;
-        private const float TIME_FOR_NEW_MEMORY_CHECK = 1.0f;
+        private const long MAX_USED_MEMORY = 1300 * 1024 * 1024; // 1.3GB
+        private const float TIME_FOR_NEW_MEMORY_CHECK = 1.0f; // Each second
 
         private Coroutine autoCleanupCoroutine;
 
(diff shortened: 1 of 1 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/MemoryManager/MemoryManager.cs
```diff
@@ -7,20 +7,20 @@ namespace DCL
...
     {
-        private const uint MAX_USED_MEMORY = 1300 * 1024 * 1024;
-        private const float TIME_FOR_NEW_MEMORY_CHECK = 1.0f;
+        private const long MAX_USED_MEMORY = 1300 * 1024 * 1024; // 1.3GB
+        private const float TIME_FOR_NEW_MEMORY_CHECK = 1.0f; // Each second
 
         private Coroutine autoCleanupCoroutine;
 
-        private uint memoryThresholdForCleanup = 0;
+        private long memoryThresholdForCleanup = 0;
```
sandrade-dcl, line 15: **Suggestion**: I notice these 2 variables are being set with the value of `MAX_USED_MEMORY` and `TIME_FOR_NEW_MEMORY_CHECK` constants in the class constructor and they are never change anymore, so... couldn't we remove the variables and simply use the constants directly?

sandrade-dcl (approved):
Approved! But please take a look to the suggestion I left, because I think we can clean the code a bit with that.

sandrade-dcl:
Ups... I didn't notice that you had set the "auto-merge" option activated.

PR #1495, opened 2021-11-10
fix: gltf importer json parse hiccups

BrianAmadori:
## What does this PR change?

This PR:
* Introduces a throttling-friendly version of the GLTF importer JSON parsing logic that should remove the hiccups of GLTF files with a big JSON header.
* Introduces a new generic way of using throttling in coroutines through `StartThrowingCoroutine`. This paves the way for a new centralized coroutine runner that unifies throttling logic for all the project (https://github.com/decentraland/explorer-technical-backlog/issues/22)
* Adapts all the throttling logic of the GLTF importer to use this new system.
* Add throttling tests for this new system

## How to test
Try:
https://play.decentraland.zone/?renderer-branch=fix/gltf-importer-json-hiccups&DISABLE_ASSET_BUNDLES

## Our Code Review Standards


unity-renderer/Assets/UnityGLTF/Scripts/GLTFThrottlingCounter.cs
```diff
@@ -0,0 +1,33 @@
+﻿using DCL;
+using UnityEngine;
+
+namespace UnityGLTF
+{
+    public class GLTFThrottlingCounter
+    {
+        public static GLTFThrottlingCounter i => new GLTFThrottlingCounter();
```
sandrade-dcl, line 8: Does this mean that each time we access to `GLTFThrottlingCounter.i` a new instance of `GLTFThrottlingCounter` is being created? If it is that we want, why we need to manage an `i` static instance of `GLTFThrottlingCounter`? Couldn't we simply create the instance of the class directly from the consumer that need it?

sandrade-dcl (approved):
LGTM! I only left a doubt about the use of the `GLTFThrottlingCounter` class.

unity-renderer/Assets/UnityGLTF/Scripts/Editor/GLTFImporter.cs
```diff
@@ -132,54 +132,54 @@ public override void OnImportAsset(AssetImportContext ctx)
                 var meshFilters = gltfScene.GetComponentsInChildren<MeshFilter>();
                 var vertexBuffer = new List<Vector3>();
                 meshes = meshFilters.Where(mf => mf.sharedMesh != null)
-                                    .Select(mf =>
-                                    {
-                                        var mesh = mf.sharedMesh;
+                    .Select(mf =>
+                    {
+                        var mesh = mf.sharedMesh;
```
Suduck, line 137: It's ok to use linq in the GLTFImporter? Won't it cause spikes? If not, just ignore this

unity-renderer/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -1814,13 +1805,13 @@ protected virtual IEnumerator ConstructMeshPrimitive(MeshPrimitive primitive, in
...
-        static protected IEnumerator YieldOnTimeout()
-        {
-            yield return null;
-            _timeAtLastYield = Time.realtimeSinceStartup;
-        }
+        // static protected IEnumerator YieldOnTimeout()
+        // {
+        //     yield return null;
+        //     _timeAtLastYield = Time.realtimeSinceStartup;
+        // }
```
Suduck, line 1814: remove commented code

unity-renderer/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -1814,13 +1805,13 @@ protected virtual IEnumerator ConstructMeshPrimitive(MeshPrimitive primitive, in
             }
         }
 
-        static protected bool ShouldYieldOnTimeout() { return ((Time.realtimeSinceStartup - _timeAtLastYield) > (budgetPerFrameInMilliseconds / 1000f / (float) GLTFComponent.downloadingCount)); }
+        // static protected bool ShouldYieldOnTimeout() { return ((Time.realtimeSinceStartup - _timeAtLastYield) > (budgetPerFrameInMilliseconds / 1000f / (float) GLTFComponent.downloadingCount)); }
```
Suduck, line 1808: remove commented code

unity-renderer/Assets/UnityGLTF/Scripts/GLTFThrottlingCounter.cs
Suduck: I agree! Since we are moving out from singletons this seems a little out of the common pattern

Suduck (approved):
Good job! Take a look at santi comments and let us know what do you think!

unity-renderer/Assets/UnityGLTF/Scripts/Editor/GLTFImporter.cs
pravusjif: GLTFImporter is only used in editor-mode, so it shouldn't be a problem

unity-renderer/Assets/UnityGLTF/Scripts/GLTFThrottlingCounter.cs
BrianAmadori: I think you guys solved me a bug. 🙏 

`GLTFThrottlingCounter` is supposed to be a singleton.

Regarding the choice of having to implement a singleton when we are trying to get rid of them is a bit nuanced. 
The thing is we can't make the GLTF loader know about `Main` `Environment` etc. Because its supposed to be a completely self contained piece of software. But at the same time, the loader needs persistent data, so for these use cases we end up again having to rely on static classes, singletons, etc.

At the same time, think that the GLTF loader doesn't have a "global" entry point. To use it we just create the `GLTFComponent` and its good to go. So, again we end up relying on singletons for having to share state between loaders--and that is what the `GLTFThrottlingCounter` does.

I accept suggestions on the matter, however I don't think a solution is possible that is on the scope on this PR, we can discuss this async.

BrianAmadori: Managed to address this in the end. Resolving.

PR #1508, opened 2021-11-12
Profanity avatar names filtering

lorux0:
## What does this PR change?

Adds profanity filtering to user names. So if a user's name contains an offensive word it should be filtered with ****:

- General chat
- Avatar name tags
- Player info card (description is also filtered)

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=feat/profanity-avatar-names-filtering
2. Login with a friend who has an offensive name like "shithead". Also set a description with offensive words at the profile menu.
3. Write anything on the general chat: your friend's name should be filtered ****head
5. The avatar name tag should be filtered
6. Open the profile info card by clicking on your friend's avatar. The name and description should be filtered

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/ProfanityFiltering/ProfanityFilterSharedInstances.cs
```diff
@@ -0,0 +1,13 @@
+﻿public static class ProfanityFilterSharedInstances
+{
+    private static RegexProfanityFilter regexFilterInstance;
+
+    public static RegexProfanityFilter regexFilter
```
AjimenezDCL, line 5: Maybe we should give this instance a more specific name. In case we have multiple ways of filtering by using a regex.

AjimenezDCL (approved):
Great job! I leave up to you whether to apply now the suggestion or leave it for when it's actually needed.

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/ProfanityFiltering/ProfanityFilterSharedInstances.cs
lorux0: I dont recall a better name for it now. Any suggestion? Maybe change it if a modification really comes up?

Issue #1437, opened 2021-11-01
Hiccup: LayoutRefreshWatcher "wall of death" issue


BrianAmadori:

AjimenezDCL:
Hiccups profiling: https://www.notion.so/decentraland/Hiccup-Profiling-14-10-2021-e8d8ff74e20d42c6a206f3824ba6dfa4

AjimenezDCL:
The hiccup has not been solved, but now heavy usage of UI wont affect other scenes

PR #1513, opened 2021-11-15
fix: UIShape only refreshes itself if the user is on the scene

AjimenezDCL:
#1437 
Now UIShapes are only refreshed when the user is on the scene. 

Pros:
A nearby scene with a heavy usage of the SDK UI and layouts (which kills performance) won't affect the performance of your scene.

Cons:
We no longer have prewarm of UIs as long as you walk around the world. When you enter the scene, your UI will be readjusted and it might take a frame for things to look correctly.


pravusjif (approved):
Tested in-world and had no problems

Suduck (approved):
I think that the issue generated is minor, however, we should make it work correctly in the future and since maybe we do an overhaul to the UI I think we are good.

Good job!

PR #1589, opened 2021-11-27
refactor: new service locator

BrianAmadori:
## What does this PR change?

This PR introduces a new service locator pattern that removes most of the `Environment` boilerplate and makes it far easier to add new services.

* All `Update()` and `LateUpdate()` calls for all services are handled by `UpdateEventHandler` and not called explicitly anymore.
* All `IntegrationTestSuite` test suites initialize their services through `InitializeServices` method. No more context factories.
* All services implements `IService` interface.
* If a service needs another service on its initialization stage, it should be done through `Initialize()` method. No more DI through this method is allowed, though constructor DI is still viable for tests only.

Note that the `Initialize()` method just instances all the services after they were registered. This allows for a "staging" state where a single service can be registered many times, each time, overriding the former. This allows for many use cases i.e. as a mock in a base class, then override by the derived class.

An important issue that had to be taken care of is that any service that depends on another service would have to be initialized last. So the building order would matter. To simplify things and prevent bugs that would be caused by this premise, the initialization order is not guaranteed anymore and any service can expect all services are initialized in its `Initialize()` method.

A shortcoming of this idea is that leads to close coupling on the long term, so in the future, most of the existing services will be moved into plugins (and out of this service locator). In fact, most of the current services should belong on the (still not existing) WorldRuntime plugin. When this plugin is put into place, the only relevant services for global scope are going to be very low level. i.e.: `PoolManager`, `AssetPromiseKeeper`, etc. When this happens, the service locator layer will be paper-thin and the coupling would not be an issue or suffer any consequences of scale.

## How to test the changes?

https://play.decentraland.zone/?renderer-branch=refactor/new-service-locator

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestController.cs
```diff
@@ -13,6 +13,14 @@ public class WebRequestController : IWebRequestController
         private IWebRequestAudio audioClipWebRequest;
         private List<WebRequestAsyncOperation> ongoingWebRequests = new List<WebRequestAsyncOperation>();
 
+        private WebRequestController ()
+        {
+        }
```
sandrade-dcl, line 18: It is needed to implement an empty constructor in this case?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/HUDController.cs
```diff
@@ -29,10 +30,21 @@ public class HUDController : IHUDController
         groupID = "UIHiddenNotification"
     };
 
-    public void Initialize(IHUDFactory hudFactory)
+    public HUDController ()
+    {
+    }
```
sandrade-dcl, line 35: It is needed to implement an empty constructor in this case?

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneController.cs
```diff
@@ -28,24 +31,33 @@ public void Initialize()
...
-            PoolManager.i.OnGet += Environment.i.platform.physicsSyncController.MarkDirty;
-
-            PoolManager.i.OnGet -= Environment.i.platform.cullingController.objectsTracker.MarkDirty;
-            PoolManager.i.OnGet += Environment.i.platform.cullingController.objectsTracker.MarkDirty;
+            // TODO(Brian): Move this later to Main.cs
+            if ( !EnvironmentSettings.RUNNING_TESTS )
+            {
+                if (prewarmSceneMessagesPool)
+                {
+                    for (int i = 0; i < 100000; i++)
```
sandrade-dcl, line 50: It would be nice to move the `100000` value to a constant with a descriptive name.

sandrade-dcl (approved):
Awesome work! 💯

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/NotificationHUD/Tests/NotificationHudTests.cs
```diff
@@ -30,7 +27,7 @@ protected override IEnumerator TearDown()
         [Test]
         public void NotificationHud_Creation()
         {
-            var views = GameObject.FindObjectsOfType<NotificationHUDView>();
+            var views = Object.FindObjectsOfType<NotificationHUDView>();
```
pravusjif, line 30: Why do we swap all the `GameObject` calls for `Object` calls?

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ParcelScenesCleaner.cs
```diff
@@ -41,15 +41,9 @@ public MarkedSharedComponentInfo(ParcelScene scene, string componentId)
 
         Coroutine removeEntitiesCoroutine;
 
-        public ParcelScenesCleaner ()
+        public void Initialize ()
         {
             removeEntitiesCoroutine = CoroutineStarter.Start(CleanupEntitiesCoroutine());
-            CommonScriptableObjects.rendererState.OnChange += OnRendererStateChange;
```
pravusjif, line 47: shouln't we keep this event subscription?
If we remove it we should remove its unsubcription from `Dispose()` and remove the `OnRendererStateChange()` implementation altogether.

pravusjif (changes requested):
great job! the code looks good

Some problems probably related to being outdated with master
- In the build I can't open avatars passport
- I'm not sure avatar LODs work
- Instantiating bots apparently is broken

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/NotificationHUD/Tests/NotificationHudTests.cs
BrianAmadori: Because those methods are really from `Object`, not `GameObject`. So using `GameObject` in the syntax works, but is misleading.

pravusjif (approved):
I'm approving the PR to avoid blocking it, when the current problems are fixed we can merge it

unity-renderer/Assets/Scripts/MainScripts/DCL/Environment/Factories/ServiceLocatorFactory/ServiceLocatorFactory.cs
```diff
@@ -0,0 +1,43 @@
+﻿using DCL.Controllers;
+using DCL.Rendering;
+using UnityEngine;
+
+namespace DCL
+{
+    public static class ServiceLocatorFactory
```
AjimenezDCL, line 7: I see this as an util rather than a factory, even a method in the `ServiceLocator` itself.

kuruk-mm:
Desktop build: https://github.com/decentraland/explorer-desktop/pull/77

unity-renderer/Assets/Scripts/MainScripts/DCL/Environment/TestEnvironment/ServiceLocatorFactory.cs
```diff
@@ -0,0 +1,53 @@
+﻿using DCL.Controllers;
+using DCL.Helpers.NFT.Markets;
+using DCL.Rendering;
+using NSubstitute;
+
+namespace DCL.Tests
+{
+    public static class ServiceLocatorFactory
```
AjimenezDCL, line 8: I find both ServiceLocatorFactories unneded, if we are going to keep them we should at least rename this one so we don't have a name collision. Although the namespace prevent issues, it might create unnecessary headaches or problems with IDEs

unity-renderer/Assets/Scripts/Tests/CursorControllerTests/CursorControllerTests.cs
```diff
@@ -30,29 +30,40 @@ protected override List<GameObject> SetUp_LegacySystems()
...
+            result.Register<IUpdateEventHandler>( () => new UpdateEventHandler());
+            result.Register<IWebRequestController>( WebRequestController.Create );
+            return result;
         }
 
-        protected override MessagingContext CreateMessagingContext()
-        {
-            return DCL.Tests.MessagingContextFactory.CreateMocked();
-        }
+        // protected override WorldRuntimeContext CreateRuntimeContext()
```
AjimenezDCL, line 44: I believe this is no longer needed and can be removed

AjimenezDCL (approved):
Left a couple of minor comments.

Great job! Love the implementation and all the tests ported. The only thing I don't like is that we have to manually call `Environment.Dispose` in the tests and might be prone to errors, but there's no workaround for that.

unity-renderer/Assets/Scripts/MainScripts/DCL/Environment/TestEnvironment/ServiceLocatorFactory.cs
BrianAmadori: Renamed and collapsed the namespace.

PR #1595, opened 2021-11-30
fix: rendereable incongruence

AjimenezDCL:
`AssetPromise` for GLTF and AB_GameObjects were not updating `asset.renderers` on certain cases when using the promise with a parent set.

This would end up in incongruences between `asset.container.GetComponentsInChildren<Renderer>()` and `asset.renderers`.

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Utils.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Utils.cs
@@ -527,12 +527,5 @@ namespace DCL.Helpers
         /// <param name="volume">Linear volume (0 to 1)</param>
         /// <returns>Value for audio mixer group volume</returns>
         public static float ToAudioMixerGroupVolume(float volume) { return (ToVolumeCurve(volume) * 80f) - 80f; }
-
-        public static string GetHierarchyPath(this Transform transform)
-        {
-            if (transform.parent == null)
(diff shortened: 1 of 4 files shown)


BrianAmadori (approved):
Can we please add a single test to validate this fix?

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Utils.cs
```diff
@@ -530,5 +527,12 @@ public static void SetLayerRecursively(Transform transform, int layer)
         /// <param name="volume">Linear volume (0 to 1)</param>
         /// <returns>Value for audio mixer group volume</returns>
         public static float ToAudioMixerGroupVolume(float volume) { return (ToVolumeCurve(volume) * 80f) - 80f; }
+
+        public static string GetHierarchyPath(this Transform transform)
```
BrianAmadori, line 531: If this method isn't used we should remove it?

BrianAmadori (approved):
LGTM!

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Utils.cs
AjimenezDCL: Oops I bring this util from the Avatar Refactor for debugging and I forgot to remove it. On it

PR #1602, opened 2021-12-01
chore: update abconverter to store depmap inside ABs

pravusjif:
* Updated converter to store dependencies inside ABs (`metadata.json`), instead of external depmaps
* Updated AB loading to first look for internal dependencies and if not found, use the old mechanism fetching the external depmap file
* Added Asset Bundle version + conversion timestamp in new internal `metadata.json`

The new mechanism should also be a bit more performant because:
1. We avoid the external depmap file request
2. We avoid creating a string in runtime for the depmap file name (internal `metadata.json` always has that same file name so we can have it cached in a const)

Tested converting a scene in the old way with an external depmap file (-113, -109) and converting a nearby scene with the new mechanism (-110, -110) and served both scenes ABs from a local webserver. It loaded with no problem fallbacking to the external depmap file for the scene that doesn't have the internal metadata file:
Screen Shot 2021-12-01 at 15 13 28


unity-renderer/Assets/ABConverter/AssetBundleMetadata/AssetBundleMetadata.cs
```diff
@@ -0,0 +1,8 @@
+
+[System.Serializable]
+public class AssetBundleMetadata
+{
+    public long timestamp = -1;
+    public float version = 1.0f;
```
AjimenezDCL, line 6: I think having a `string` is better here so we can have meaningful versioning i.e. `1.001.120`.

AjimenezDCL (approved):
LGTM

kuruk-mm (approved):
Looks great

PR #1613, opened 2021-12-02
fix: portable experiences taskbar button

pbosio:
context menu for quitting portable experience was not appearing after pressing the portable experience's button in the taskbar.


sandrade-dcl (approved):
It is working perfectly now!

PR #1625, opened 2021-12-07
refactor: legacy tests lifecycle

BrianAmadori:
## What does this PR change?

This PR introduces a simplification of `IntegrationTestSuite_Legacy` that aims to mitigate the following issues:

* The `SetUp` and `TearDown` between tests being very fuzzy, due to some systems being reused across tests and other systems being reinstantiated.
* Hacks were present to deal with `MainTest` being loaded on each test suite.
* Hacks were present to reduce tests duration by only loading `MainTest` for the first test on a test suite.
* `IntegrationTestSuite_Legacy` having too much responsibilities and trying to cover many use cases to deal with the use of specific systems on each test. (i.e. providing `sceneController` and `cameraController` references).

All of these issues combined are (1) making the tests codebase VERY costly to maintain and (2) are blocking the new service locator initiative* (#1589). 

After the changes introduced by this PR, all systems should be disposed and re-created between tests.

Also:
* `playmode-tests` job now takes about 7 minutes to finish (from 12 mins).
* `MainTest` and visual tests scene were removed. All the application state is arranged from the ground up and disposed without `LoadScene`.
* `IntegrationTestSuite_Legacy` no longer provides access to systems. Each test creates its own test scene and access the systems via `Environment` or `SceneReferences` as needed.
* As `IntegrationTestSuite_Legacy` is now simplified, the migration to `IntegrationTestSuite` is now cheaper because part of the migration is already done. (i.e. moving the system accessors to each suite).
* As we no longer use scenes for tests, now each suite (legacy or not) can select which systems will be created. The current implementation covers all systems, including the systems that are not part of `Environment`, `PluginSystem`, etc.

*About the service locator blocking:
This happens in part because the service locator is disposed between tests, but with the legacy integration suite, some legacy systems are not disposed in the `TearDown` phase. As the legacy systems continue to run in the background, they try to use services in the `Update`  loop, and those services are null in some frames between the `TearDown` and `SetUp` of the new test.

## How to test the changes?

2. ...

## Our Code Review Standards


pravusjif (approved):
* After fixing the use of `Assets/ABConverter/VisualTestScene.unity` in `Assets/ABConverter/VisualTests.cs` we should be good to go
* Tested the build with a complete test run and found no problems

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/PoolManagerFactory.cs
```diff
@@ -0,0 +1,21 @@
+﻿using UnityEngine;
+
+namespace DCL
+{
+    public static class PoolManagerFactory
```
lorux0, line 5: I think we should avoid using static classes, in this case, a factory. It makes it harder to extend, scale or mock

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneController.cs
```diff
@@ -43,6 +43,9 @@ public void Initialize()
 
             PoolManager.i.OnGet -= Environment.i.platform.cullingController.objectsTracker.MarkDirty;
             PoolManager.i.OnGet += Environment.i.platform.cullingController.objectsTracker.MarkDirty;
+
+            DCL.Environment.i.platform.updateEventHandler.AddListener(IUpdateEventHandler.EventType.Update, Update);
```
lorux0, line 47: it would be great to inject all needed dependencies from constructor instead of nesting singletons. It makes it harder to test and/or scale adding implicit coupling. This class is really big which would take probably a big refactor, out of the scope of this PR. Technical debt?

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Utils/WorldStateUtils.cs
```diff
@@ -129,5 +130,47 @@ static IParcelScene GetCurrentScene()
 
             return scene;
         }
+
+        public static IParcelScene CreateTestScene(LoadParcelScenesMessage.UnityParcelScene data = null)
```
lorux0, line 134: here i see another chance for improvement if static calls are removed

lorux0 (approved):

unity-renderer/Assets/DCLPlugins/BuilderInWorld/Tests/BIWCommonShould.cs
```diff
@@ -158,11 +171,18 @@ public void GetCorrectSceneSize()
         Assert.AreEqual(fourResult, new Vector2Int(2, 2));
         Assert.AreEqual(fiveResult, new Vector2Int(3, 1));
         Assert.AreEqual(sixResult, new Vector2Int(1, 3));
+
+        UnityEngine.Object.Destroy(firstScene.gameObject);
```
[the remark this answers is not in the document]
BrianAmadori: let's add this as technical debt, as this feedback is for the original implementation of the test.

unity-renderer/Assets/DCLPlugins/BuilderInWorld/Tests/BIWSceneManagerShould.cs
```diff
@@ -188,27 +207,30 @@ public void ReturnTrueWhenParcelSceneDeployedFromSDKIsCalled()
...
         Land land = new Land();
         land.parcels = new List<Parcel>() { parcel };
 
         LandWithAccess landWithAccess = new LandWithAccess(land);
-        Scene scene = new Scene();
-        scene.parcelsCoord = new Vector2Int[] { parcelCoords };
-        scene.deploymentSource = Scene.Source.SDK;
 
-        landWithAccess.scenes = new List<Scene>() { scene };
+        {
```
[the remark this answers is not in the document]
BrianAmadori: The `scene` name is used by the class scope, with this is prevented.

unity-renderer/Assets/Scripts/MainScripts/DCL/InitialSceneReferences/MainSceneFactory.cs
```diff
@@ -0,0 +1,89 @@
+﻿using System.Collections.Generic;
+using UnityEngine;
+using UnityEngine.SocialPlatforms.Impl;
+
+namespace DCL
+{
+    public static class MainSceneFactory
```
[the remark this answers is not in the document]
BrianAmadori: We don't need to mock this class, as is temporary glue until we move all the systems to stop using MonoBehaviours, when that is done we will remove it and all the application flow is going to be handled by the service locator that should be very friendly to testing.

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/PoolManagerFactory.cs
BrianAmadori: Yeah, I agree about removing the static classes, although they are fine for containers of pure methods like this that don't need mocking (like our `Utils` set of classes). The idea of this class was to move the pool prewarming out of `SceneController` as it shouldn't be its responsibility. I'm going to think how to improve this, but will do on another PR.

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneController.cs
BrianAmadori: Removing responsibilities from `SceneController` is an ongoing effort that goes back to the start of the project! We'll continue doing that for sure. About nesting singletons there's a justification but's too wally of texty to convey here, to summarize, the design of this will be greatly improved when the dcl runtime is moved to its own plugin.

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Utils/WorldStateUtils.cs
BrianAmadori: This is an utility method only reserved for tests usage, don't see the real benefit of moving to a mockable class (in fact it was moved out of `SceneController` to reduce its responsibilities, something that you correctly pointed out earlier).

Issue #1088, opened 2021-08-27
Integrate GPUSkinning to scene skinned meshes (GLTFShapes and ABs)


BrianAmadori:
We should implement GPU Skinning for scene meshes (currently is only running for avatars) so that the skinning is done in the GPU through a shader instead of in the CPU.

These 2 GPU Skinning for avatar PRs may come in handy:
https://github.com/decentraland/unity-renderer/pull/1041
and
https://github.com/decentraland/unity-renderer/pull/1140

Tasks:
- [x] Update `SimpleGPUSkinning.cs` to allow configuring which 2 UV channels (0..7) are to be used for passing the bones that into the shader.
- [x] Integrate the `GPUSkinning` shader calculations to every relevant pass in the Lit shader. If a unitypackage shader code is used for a pass, duplicate it locally and customize it.
- [x] Fix SSAO visual artifacs.
- [x] Deal with "mesh bounds" issue. [Configured enormous bounds for non-avatars since we are not culling animations]
- [x] Debug issue with broken gigantic meshes, some are skinned correctly, some don't. [The problem was scenes using skinned meshes and non-skinned meshes with the same material, so when we enabled the GPU_SKINNING keyword for those `sharedMaterials` the gpu skinning would run for non-skinned meshes and break their vertices]. Changed `sharedMaterials` usge for `materials` but that breaks our previous optimization to share materials between meshes
- [x] Move GPUSkinning to another intermediate class or place (maybe GLTFshape class?) instead of DCLAnimator, since many GLTFShapes don't have the Animator component and thus wouldn't be affected by GPUSkinning when that happens
- [ ] Implement way of keeping the "shared materials" optimization for GPUSkinned meshes [Sometimes the same material is being shared between non-skinned meshes and skinned meshes, and sometimes between more than 1 skinned meshes but having different armatures... those cases break the GPU Skinning]
- [ ] Compare performance in heavy-skinning scenes (casino, soho plaza)
- [ ] Deal with calculating the mesh bounds: maybe let the SkinnedMeshRenderer run for a frame and get that data before replacing it? [We can't just set giant bounds because the SceneBoundsChecker hides the mesh in several cases]
- [ ] Find a way of avoiding wearables GPUSkinnig Setup on the GLTF Loading pipeline for wearables, as that setup is done in the Avatar Renderer pipeline
- [ ] Patch ABs material on their loading to match the materials CRC change regarding GPU_SKINNING keyword
[...]

PR #1628, opened 2021-12-09
feat: gltfshape gpu skinning

pravusjif:
By integrating our GPUSkinning to the Lit shader used on most materials in-world, every skeletal animation will be skinned in the GPU instead of the CPU, greatly improving performance for those scenes (Avatars already had this sincde the metaverse festival)

* Found out that we can use up to 16 interpolator variables between the vertex programs and fragment programs. But we are limited to using the `TEXCOORD` from 0 to 7  as input tex coords for the vertex programs (AKA using [`Mesh.SetUVs()`](https://docs.unity3d.com/2020.3/Documentation/ScriptReference/Mesh.SetUVs.html) from C#).
Screen Shot 2021-12-08 at 17 44 45 copy

* Updated `SimpleGPUSkinning` class to allow configuring the UVChannels used for the bones on instantiation.


pravusjif:
Closing PR for reasons explained at https://github.com/decentraland/unity-renderer/issues/1088

PR #1632, opened 2021-12-09
fix: fail stream includes exception & better error handling of some assets during loading screen

lorux0:
## What does this PR change?

Reports a critical error (bypassing standard error flow) when any gltf, ab, wearable, avatar shape, renderables throws a load error while in loading screen. A popup is shown immediatelly on desktop client.

The loading screen popup implementation is at [explorer-desktop](https://github.com/decentraland/explorer-desktop) project

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=fix/init-freeze
2. Restart the browser many times and wait until the loading finishes
3. Run around the world and check objects are being loaded correctly

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_GLTF.cs
```diff
@@ -33,9 +34,11 @@ public override void Load(string targetUrl, Action<LoadWrapper> OnSuccess, Actio
 
 
             if (customContentProvider == null)
-                loadHelper = new RendereableAssetLoadHelper(this.entity.scene.contentProvider, entity.scene.sceneData.baseUrlBundles);
+                loadHelper = new RendereableAssetLoadHelper(this.entity.scene.contentProvider, entity.scene.sceneData.baseUrlBundles,
```
BrianAmadori, line 37: We shouldn't stop the client when a given asset fails importing, because this means that bad authored content can crash the client. The `GLTFShape` or `WearableController` should fail silently and continue loading.

Also, if we need some way of getting failure information, we should use `OnFail` instead of passing this new dependency around. Maybe we can change `OnFail` signature to take an `Exception` parameter if we want the details.

BrianAmadori (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadWrapper/LoadWrapper_GLTF.cs
lorux0: In case the asset fails importing in a way that was supported, the load will continue as always. Only will report a critical error when an exception that was not being handled before occurs (an exception that used to block the loading). You can check `AssetPromise_AB.OnLoad` included `DCLCoroutineRunner.Run` which handles these exceptions that im mentioning.

I dont have a clean way for injecting the critical error reporting, since load objects needs empty constructors due inheritence (`LoadWrapper`, `LoadableShape`, etc). I could try to hook it somewhere on the standard fail stream but as far as i investigated is way too complicated, and this workaround was avoiding major refactors

BrianAmadori (approved):
LGTM codewise! Didn't have time to test the build.

PR #1634, opened 2021-12-13
chore: move shaders prewarm to first loading screen

pravusjif:
Moved shader variants prewarm to first loading screen, to avoid hiccups in landing page.


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/LoadingHUD/Bridge/LoadingBridge.cs
```diff
@@ -20,5 +22,11 @@ public void SetLoadingScreen(string jsonMessage)
         if (!string.IsNullOrEmpty(payload.message))
             DataStore.i.HUDs.loadingHUD.message.Set(payload.message);
         DataStore.i.HUDs.loadingHUD.showTips.Set(payload.showTips);
+
+        if (!prewarmedShaderVariants)
+        {
+            Resources.Load<ShaderVariantCollection>("ShaderVariantCollections/shaderVariants-selected").WarmUp();
```
lorux0, line 28: whats is the initiative of moving the prewarm here? i find it less cohesive, since this class is handling the loading screen behaviour but now also has a shader prewarm which makes it harder to find

BrianAmadori: The initiative is that the landing gets completely frozen for around 5 seconds when the user is interacting with the login options.

Maybe we can do this event based and subscribe from `Main`?

BrianAmadori (changes requested):
I think we should wait at least a frame, because I experienced the hiccup just before the loading screen is supposed to be visible. Also we can use events to put the prewarm in `Main`.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/LoadingHUD/Bridge/LoadingBridge.cs
```diff
@@ -12,6 +12,8 @@ public class Payload
         public bool showTips = false;
     }
 
+    public event Action<bool> OnLoadingScreenVisibleStateChange;
```
lorux0, line 15: perhaps you can avoid creating a new event by subscribing to `DataStore.i.HUDs.loadingHUD.visible` changes.. anyway its up to you LGTM!

lorux0 (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/LoadingHUD/Bridge/LoadingBridge.cs
pravusjif: Totally, I can't believe I didn't see that a couple of lines above my event call T_T, I'll change that, thanks!

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneController.cs
```diff
@@ -77,9 +77,6 @@ public void Start()
             {
                 PoolManagerFactory.EnsureEntityPool(prewarmEntitiesPool);
             }
-
-            // Warmup some shader variants
-            Resources.Load<ShaderVariantCollection>("ShaderVariantCollections/shaderVariants-selected").WarmUp();
         }
 
         public void Dispose()
```
BrianAmadori, line 82: For correctness don't forget to unsubscribe to the event on the `Dispose()` method.

BrianAmadori (approved):
LGTM!

PR #1638, opened 2021-12-15
feat: add AttachToAvatar component

pbosio:
add component to be able to attach entities to avatar's bones


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/AvatarAttach/AvatarAttachHandler.cs
```diff
@@ -0,0 +1,121 @@
...
+        {
+            prevModel = model;
+            model = newModel;
+
+            if (model == null)
+            {
+                return;
+            }
+
+            sceneBoundsChecker?.AddEntityToBeChecked(entity);
```
BrianAmadori, line 47: This is not needed. The `sceneBoundsChecker` maintenance is already taken care of by `ParcelScene`.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/AvatarAttach/AvatarAttachHandler.cs
```diff
@@ -0,0 +1,159 @@
...
+                {
+                    componentUpdate = null;
+                    yield break;
+                }
+
+                var anchorPoint = anchorPoints.GetTransform(anchorPointId);
+
+                if (IsInsideScene(CommonScriptableObjects.worldOffset + anchorPoint.position))
+                {
+                    entity.gameObject.transform.position = anchorPoint.position;
```
BrianAmadori, line 123: Just a warning: the timing here *might* be off if we don't use `LateUpdate`, this meaning that the attachments can be off by a frame respecting the avatar animations. 

At the same time `new WaitForEndOfFrame()` will not work on CI, so if this becomes an issue, we can use the `IUpdateEventHandler` that can be accessed by doing `Environment.i.platform.updateEventHandler`.

pravusjif (approved):
LGTM nice job! Also tested the build and had no problems with current avatars

PR #1641, opened 2021-12-15
feat: biw project publish

### LINKED ISSUE #1075: Integrate Land planner in the Project publish workflow
We should make the land publisher as the default publish method

Suduck:
## What does this PR change?

This PR Fix #1075

This PR implements the basic functionality to publish a project of 1x1 in a land. It also unified the publish flow of lands and projects.

Take into account that the flow will be revamped so do not expect the publish views to work too perfect since they will be redone 

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=feat/biw-project-publish
2. Create a project or edit one of 1x1
3. Publish the project in a land 1x1

## Our Code Review Standards


unity-renderer/Assets/DCLPlugins/BuilderInWorld/Publisher/ProjectPublishHUD/Scripts/Projects/PublishProjectProgressView.cs
```diff
@@ -136,14 +140,17 @@ private IEnumerator FakePublishProgress()
         {
             while (true)
```
pravusjif, line 141: How does this while (true) loop execution stop? is this `FakePublishProgress()` being used?

unity-renderer/Assets/DCLPlugins/BuilderInWorld/Publisher/Scripts/Publisher.cs
```diff
@@ -1,19 +1,245 @@
...
+            {
+                case IBuilderScene.SceneType.PROJECT:
+                    projectPublisher.StartPublishFlow(scene);
+                    break;
+                case IBuilderScene.SceneType.LAND:
+                    landPublisher.StartPublishFlow(scene);
+                    break;
+                default:
+                    Debug.Log("This should no appear, the scene should have a know type!");
+                    break;
```
pravusjif, line 74: I would remove this `default` case if it's an impossible situation

pravusjif (changes requested):
code-wise i think it's good, i just left a couple of questions.

Regarding the BiW UI Layout a bug was instroduced:

* this is how I see my lands in prod:
Screen Shot 2022-01-06 at 13 44 41

* And this is how I 
Screen Shot 2022-01-06 at 13 48 35
see my lands in this PR

maybe it's because it's outdated with master ?

Also if I want to edit a previous builder scene, the loader alwasy freezes at 50% and never finishes, mayber that's another story though.

unity-renderer/Assets/DCLPlugins/BuilderInWorld/Publisher/ProjectPublishHUD/Scripts/Projects/PublishMapView.cs
```diff
@@ -0,0 +1,107 @@
...
+                    landsToHighlight.Add(landParcel);
+                }
+            }
+            MapRenderer.i.HighlightLandsInRed(landsToHighlight);
+        }
+
+        public void GoToCoords(Vector2Int coords)
+        {
+            //Reset scroll
+            scrollRectContentTransform.anchoredPosition = new Vector2(-1219, -1596);
```
sandrade-dcl, line 57: Could we move these `-1219, -1596` coords to a constant with a descriptive name?

unity-renderer/Assets/DCLPlugins/BuilderInWorld/Publisher/ProjectPublishHUD/Scripts/Projects/PublishProjectDetailView.cs
```diff
@@ -0,0 +1,287 @@
...
+        public event Action<PublishInfo> OnPublishButtonPressed;
+
+        [Header("Map Renderer")]
+        public PublishMapView mapView;
+
+        [SerializeField] internal Button cancelButton;
+        [SerializeField] internal Button publishButton;
+
+        [SerializeField] internal Button rotateLeftButton;
+        [SerializeField] internal Button rotateRightButton;
```
sandrade-dcl, line 65: Why don't we use `ButtonComponentViews` here? In this way we can use the buttons implemented in our UI Components library.

sandrade-dcl (approved):
Apart of what Pravus detected in the comments above, LGTM! I only left some little comments. Good work!

Suduck:
> code-wise i think it's good, i just left a couple of questions.
> 
> Regarding the BiW UI Layout a bug was instroduced:
> 
> * this is how I see my lands in prod:
>   Screen Shot 2022-01-06 at 13 44 41
> * And this is how I
>   Screen Shot 2022-01-06 at 13 48 35
>   see my lands in this PR
> 
> maybe it's because it's outdated with master ?
> 
> Also if I want to edit a previous builder scene, the loader alwasy freezes at 50% and never finishes, mayber that's another story though.

This is already happening in the publish branch, since this prefab is different from the production one, we must redo this part in the dev prefab, I will openning a new PR just for that

unity-renderer/Assets/DCLPlugins/BuilderInWorld/Publisher/ProjectPublishHUD/Scripts/Projects/PublishProjectDetailView.cs
sandrade-dcl: Ok, only for taking into account when you have to change this part: The use of a `Button` or a `BaseComponentView` is the same in terms of "overkill", both are used in the same way (you will subscribe to its `onClick` event), but the advantage of using a button from our UI library is that it will ensure to have a visual design (from the base prefab) centralized and coherent with the rest of our UIs.

PR #1647, opened 2021-12-16
refactor: extendible settings creation

lorux0:
## What does this PR change?

Makes `Settings` creation extendible, so it can be overrided in desktop.
For example: change the path where the graphic quality settings preset are, or change default general settings.

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=feat/desktop-default-settings
2. Open settings menu
3. Click on `Reset All` button
4. All settings should be set to default as expected

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Settings/Settings.cs
```diff
@@ -1,78 +1,64 @@
...
+        public readonly ISettingsRepository<AudioSettings> audioSettings;
+        
+        private readonly QualitySettingsData qualitySettingsPreset;
+        private readonly BooleanVariable autoQualitySettingsEnabled;
         private readonly AudioMixer audioMixer;
 
-        public Settings()
+        private bool isDisposed;
+
+        public static void CreateSharedInstance(ISettingsFactory settingsFactory)
```
BrianAmadori, line 28: I think we can transform this pattern into the service locator pattern when the new service locator PR is merged. 

With the new pattern, an interface would have to be extracted from this `Settings` class, and it should be like this:

On `unity-renderer`:
`Register<ISettings>( () => new Settings() )`

Then, on [SetupServices](https://github.com/decentraland/unity-renderer/blob/7e1e5971e1e97c98176245ce757dcd63e2d76636/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Scene/Main/Main.cs#L76) an alternative service locator can be built by using this line:

`serviceLocator.Register<ISettings>( () => new DesktopSettings() )`

(Here's the `ServiceLocatorFactory.CreateDefault()` [implementation](https://github.com/decentraland/unity-renderer/blob/14296c66f2e967d8339cad53317505ce0d4467bf/unity-renderer/Assets/Scripts/MainScripts/DCL/Environment/Factories/ServiceLocatorFactory/ServiceLocatorFactory.cs#L9))

BrianAmadori (approved):
LGTM!

PR #1663, opened 2021-12-21
fix: add a timer to call the owned wearables each minute

### LINKED ISSUE #1656: We should refresh the backpack each time that an user click it
If an user claim a NFT and goes to the backpack, it will refresh it to see it, but if you claim another and go to the backpack you won't see it

Suduck:
## What does this PR change?

Fix #1656

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=fix/backpack-nft
2. Test that the NFTs are loaded again after a minute

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/AvatarEditorHUDController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/AvatarEditorHUDController.cs
@@ -104,9 +104,6 @@ public class AvatarEditorHUDController : IHUD
 
     private void LoadOwnedWereables(UserProfile userProfile)
     {
-        if(string.IsNullOrEmpty(userProfile.userId))
-            return;
-        
         // If there is more than 1 minute that we have checked the owned wearables, we try it again
         // This is done in order to retrieved the wearables after you has claimed them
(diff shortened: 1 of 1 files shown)


Kinerius (approved):
Nice!

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/AvatarEditorHUDController.cs
```diff
@@ -103,11 +104,18 @@ private void LoadUserProfile(UserProfile userProfile)
 
     private void LoadOwnedWereables(UserProfile userProfile)
     {
-        if (ownedWearablesAlreadyLoaded || ownedWearablesRemainingRequests <= 0 || string.IsNullOrEmpty(userProfile.userId))
+        // If there is more than 1 minute that we have checked the owned wearables, we try it again
+        // This is done in order to retrieved the wearables after you has claimed them
+        if ((Time.realtimeSinceStartup < lastTimeOwnedWearablesChecked+60 &&
```
Kinerius, line 109: In order to avoid those explanatory comments, we can extract the condition to a self-explanatory method like

`private bool IsWearableUpdateInCooldown()
{
    return Time.realtimeSinceStartup < lastTimeOwnedWearablesChecked + 60
}`

also, the 60 could be a constant.

PR #1679, opened 2021-12-22
feat: fps panel improvements

### LINKED ISSUE #1680: Fps display not working on desktop
FPS Display is not being updated on desktop while on fullscreen

Kinerius:
## What does this PR change?

The FPS panel was not working while on Fullscreen since you have no focus on the browser/kernel/launcher, so a condition was added to avoid this.

Changed the fps label component from Text to TextMeshPro and assigned a better font with outline

Refactored label string creation for more flexibility

Fixes #1680 

Preview of the panel:
image

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/Debugging/FPSDisplay/FPSDisplay.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/Debugging/FPSDisplay/FPSDisplay.cs
@@ -144,7 +144,7 @@ namespace DCL.FPSDisplay
 
         private string SetColor(string color) { return $"<color={color}>"; }
 
-        private string AddTitle(string text) { return $"<size=110%>{text}<size=100%><br>"; }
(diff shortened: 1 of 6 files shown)


kuruk-mm:
Can we do something to avoid the text width change?

I saw that this font has different characters width sizes making the whole text change and it's a bit annoying.

And when we have numbers that change, we can add zeros like `Game Time: 13.56300` or `fps: 030`

lorux0 (approved):
Great work!

unity-renderer/Assets/Scripts/MainScripts/Debugging/FPSDisplay/FPSDisplay.cs
```diff
@@ -24,8 +35,33 @@ private void OnEnable()
...
+        {
+            if (current != null)
+            {
+                currentRealmValue = current.serverName;
+            }
+        }
+
+        private void SetupKernelConfig()
+        {
+            kernelConfigPromise = KernelConfig.i.EnsureConfigInitialized();
```
lorux0, line 57: it would be good to dispose this promise somewhere

PR #1681, opened 2021-12-22
fix: incorrect cursor locking during loading and avatar creation

lorux0:
## What does this PR change?

Disables cursor locking that was occuring on `EmotesHUDController` which caused the locking during the first loading or the first time you create an avatar as a guest.

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=fix/incorrect-mouse-locking
2. Enter as a guest from a clean session
3. While the app is loading, click anywhere on the screen. The cursor should not be locked
4. Enter into the avatar creation flow. Click anywhere on the screen. The cursor should not be locked
5. Enter the metaverse. Press `B` to open Emotes HUD. Everything should work as expected

## Our Code Review Standards


pravusjif (approved):
works great nice job!

PR #1682, opened 2021-12-22
fix: cardinal points visibility in the minimap

RominaMarchetti:
## What does this PR change?
The map rendering was overlapping the cardinal points references. Now they are visible again.

Before
Screen Shot 2021-12-22 at 18 12 54

After
Screen Shot 2021-12-22 at 18 14 21

Test link: https://play.decentraland.zone/?renderer-branch=fix/CardinalPointsVisibility


Kinerius (approved):
Thank you!

PR #1684, opened 2021-12-23
fix: portable experiences container X position

RominaMarchetti:
Task bar right margin decreased.

Screen Shot 2021-12-22 at 16 08 02

How to test? **
Test link: https://play.decentraland.zone/?renderer-branch=fix/PortableExperiencesUIPosition

1- Write the following in the browser console:
**spawnPortableExperienceScene("urn:decentraland:off-chain:static-portable-experiences:radio", "main")**
2- Check that the right margin is now correct.


Suduck (changes requested):
The pop up is not correctly visible
image

Suduck (approved):

PR #1689, opened 2021-12-27
fix: user profile activation date label

### LINKED ISSUE #1651: All "Active since" in the passport have the same date
We need to check if this date can be retreived from the profile. If not, this line needs to be removed.

lorux0:
## What does this PR change?

`Fixes #1651 `
"Active since {month} {year}" showed at PlayerProfileInfoCard was not getting refreshed accordingly to the user's activation date. This changes fixes this problem. Kernel was not providing this data from the `UserProfileBridge`, so the solution was fetching the user profile data directly from the catalyst, parsing the `timestamp` considered as the profile's activation date.

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=fix/profile-active-since
2. Click on the users around you to show the player's info card
3. In case the user has an activation data set, you should see a label indicating the date it was activated. ie: _Active since December 2021_
4. In case the user has no data about activation date, the label should be hidden

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/PlayerInfoCardHUD/PlayerInfoCardHUDController.cs
```diff
@@ -152,13 +160,34 @@ private void SetUserProfile(UserProfile userProfile)
...
         LoadAndShowWearables(userProfile);
         UpdateFriendshipInteraction();
+        view.HideActivationDate();
+
+        GetProfileFromCatalyst(userProfile)
+            .Then(catalystProfile =>
+            {
+                if (catalystProfile == null) return;
+                view.ShowActivationDate(catalystProfile.GetActivationDate());
+            });
```
pbosio, line 170: could you assign the fetch to a field to be able to (and) dispose it later? (like in `CloseCard` and `Dispose` methods)
like:
`Promise<string> profilePromise = GetProfileFromCatalyst(userProfile).Then(...)`
`profilePromise.Dispose()`
since pending promises could lead to unwanted errors during tests run?

pbosio (commented):
could we cache the response, so opening the passport for the same user doesn't trigger the fetch for the profile all over again?

lorux0:
Closed. We will be discussing a better approach to retrieve the activation date of a user

PR #1694, opened 2021-12-28
feat: cursor unlocking with right click

lorux0:
## What does this PR change?

Allows you to unlock the cursor with the right click showing a hint in the middle of the screen for a couple of seconds whenever you lock it. The hint is only shown one time per session to avoid spamming the user.
Camera flick has been fixed when you unlock via this new input.

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=feat/cursor-unlocking-right-click
2. Once you are in world click on the screen to lock the cursor
3. A hint should be shown indicating that you can unlock it with with right click
4. The hint should have a duration of 3 seconds
5. Use the right click, the cursor should be unlocked
6. Lock the cursor again by clicking the screen, the hint should not appear again
7. Refresh the tab to start a new session and repeat from step 2

## Our Code Review Standards


pravusjif (changes requested):
The feature is working great, there is only 1 bug:
I thought the "esc unlocking" was working better in this PR but then I found that locking the mouse is bugged, sometimes it locks just fine, but try locking and unlocking the cursors and you will see that many times it never really gets locked (for the browser at least), so if you have more than 1 monitor/display, the mouse gets outside the chrome window when reaching the  edges of the screen

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/MouseCatcher/CursorLockHint.cs
```diff
@@ -0,0 +1,60 @@
...
+            if (hideRoutine != null)
+                StopCoroutine(hideRoutine);
+            hideRoutine = StartCoroutine(Wait(duration, () =>
+            {
+                toastRoot.Hide();
+                hideRoutine = null;
+            }));
+        }
+
+        private IEnumerator Wait(float delay, Action onFinishCallback)
```
sandrade-dcl, line 54: Maybe it could be fine to move this kind of utils functions to some common class in our project (for example the `Utils.cs` class) in order to have it available by anyone.

sandrade-dcl (approved):
Good work! Looks and feels great the new mouse behaviour.

I only left some little suggestions.

Kinerius (approved):
NIce! This feature is very usable using the trackpad.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/TaskbarHUD/TaskbarHUDController.cs
```diff
@@ -369,7 +369,10 @@ public void OnCloseWindowToggleInputPress()
         if (mouseCatcher.isLocked)
             return;
 
-        view.chatButton.SetToggleState(true);
+        // this gives focus to chat input field and breaks the InputProcessor.Modifier.FocusNotInInput functionality
+        // when pressing escape and not having cursor locked
+        // like moving the character around without cursor locking
+        //view.chatButton.SetToggleState(true);
```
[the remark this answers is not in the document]
lorux0: I left the comment just as a warning in case it is required to activate the chat button for some reason in the future. I didnt see any troubles with the tests i made regarding to this issue, so it would be cleaner to remove the code

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/MouseCatcher/CursorLockHint.cs
lorux0: done!

pravusjif (approved):
I love the feature so I'll approve the PR because the team wants to check users feedback on this in production regardless of the ESC-unlocking bug (if the user only uses right-click to unlock, there's no bug).

PR #1716, opened 2022-01-03
fix: entity attach to first person camera

pbosio:
`Camera.main` was `null` on `Awake`
setting the `GameObject` as child of `firstPersonParent` was leaving it orphan.
`cameraPosition` is set, in the prefab, with the transform for the first person camera.
fixed using `cameraPosition` transform instead of `Camera.main.transform`


pravusjif (approved):
Epic!
This also fixes the "portal" scene that was broken somehow: https://play.decentraland.org/?island=Idqts&position=0%2C0&realm=loki&renderer-branch=fix%2Ffirst-person-camera-attach

AjimenezDCL (approved):

PR #1724, opened 2022-01-05
feat: Integrate StartMenu into PreviewMode

### LINKED ISSUE #1722: Integrate the StartMenu to PreviewMode
The idea is that the StartMenu can be usable in PreviewMode in the same way as in the normal "world" flow.

sandrade-dcl:
## What does this PR change?
Fix #1722 
The StartMenu is now usable in PreviewMode in the same way as in the normal "world" flow. In this mode there will be some sections deactivated (for example "Explore" or "Builder"), only those that have sense to use in preview mode will be activated in the menu.

image

## How to test?
1. Create a new empty scene
image

2. Play it as preview mode
image

3. Add in the url the url para: `&renderer-branch=feat/Integrate-StartMenu-into-PreviewMode`
4. Notice the StartMenu is now usable in Preview Mode.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/DebugController/Resources/DebugView.prefab
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/DebugController/Resources/DebugView.prefab
@@ -87,8 +87,8 @@ RectTransform:
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
   m_AnchorMin: {x: 0, y: 1}
   m_AnchorMax: {x: 0, y: 1}
-  m_AnchoredPosition: {x: 1339.5454, y: -32.35}
-  m_SizeDelta: {x: 360, y: 34.7}
+  m_AnchoredPosition: {x: 1439.7731, y: -32.35}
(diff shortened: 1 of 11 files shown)


lorux0 (approved):
LGTM!

pravusjif (commented):
The only problem I see is that there is a discrepancy with the settings we have in production, for example in preview mode with this renderer (tested with @latest and @next) we don't see the day-night lifecycle settings, which are important, we only see the old "night mode" toggle:
Screen Shot 2022-01-07 at 15 49 00

sandrade-dcl:
> The only problem I see is that there is a discrepancy with the settings we have in production, for example in preview mode with this renderer (tested with @latest and @next) we don't see the day-night lifecycle settings, which are important, we only see the old "night mode" toggle

@pravusjif We don't see the procedural skybox setting in Preview Mode because it is configured with the feature flag system which is not being taken into account by the Preview Mode flow. The same will occur with any other feature managed by feature flags (showed in the settings panel or elsewhere).

This is out of the scope of this PR, so we should start discussing in the team about the necessity of make the current feature flag system compaible with Preview Mode and attack it on a separate PR.

pravusjif (approved):

PR #1725, opened 2022-01-05
feat: add `CameraModeArea` ecs component

pbosio:
ECS component that forces the camera mode to the player inside a defined area.
While inside this area, player is **not** able to use input key `V` to toggle the camera freely.

issue: https://github.com/decentraland/sdk/issues/123

*small refactor: now `CommonScriptableObjects.cameraMode.Set()` can be use to change camera mode
*added ScriptableObject `CommonScriptableObjects.cameraModeInputLocked` to disable the `V` key input for camera mode toggle
*added player's avatar collider to DataStore (`DataStore.i.player.playerCollider`) since it was impossible to access it without ending up with cyclic references to assemblies


Kinerius:
We have to test this with #1694

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/CameraModeArea/CameraModeArea.cs
```diff
@@ -0,0 +1,176 @@
...
+
+namespace DCL.Components
+{
+    public class CameraModeArea : IEntityComponent, ICameraModeArea
+    {
+        [Serializable]
+        public class Model : BaseModel
+        {
+            [Serializable]
+            public class Area
```
AjimenezDCL, line 18: We already have the `BoxTriggerArea` that represents the same for the `AvatarTriggerArea`. If we can find a nice naming and a proper location for the assemblies, we could use the same class.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/CameraModeArea/CameraModeArea.cs
```diff
@@ -0,0 +1,176 @@
...
+            this.playerCollider = playerCollider;
+
+            updateEventHandler.AddListener(IUpdateEventHandler.EventType.Update, Update);
+        }
+
+        internal void Update()
+        {
+            bool playerInside = IsPlayerInsideArea();
+
+            switch (playerInside)
```
AjimenezDCL, line 114: First time I see this `switch / when` thing. Why is needed?

AjimenezDCL:
Approved with a few comments

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/CameraModeArea/CameraModeArea.cs
pbosio: I thought about that.
The main reasons I didn't create a common type were:
1- `AvatarTriggerArea` `BoxTriggerArea` implement code, it's not just a type. So I could create a type and make `BoxTriggerArea` inherit from that, but it doesn't make de code more readable cause it will hide the `area` field
2- If we want to add more fields in one of the "Areas" and not in the other it would require a little more of work and we will find our-self in the same situation as in point 1, we will need to inherit one base type from another
So I decided to keep them separated

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/CameraModeArea/CameraModeArea.cs
pbosio: It's the same as:
```
if (playerInside && !isPlayerInside)
{
    // ...
}
else if (!playerInside && isPlayerInside)
{
    // ...
}
```
but way more cool and easier to read (was a rider's suggestion) :P

AjimenezDCL: Aahhh the naming confused me, I didnt notice `playerInside` is different than `isPlayerInside`.

pravusjif: Interesting, I too didn't know we could have multiple conditions in the switch cases this way, nice

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/CameraModeArea/CameraModeArea.cs
```diff
@@ -0,0 +1,176 @@
...
+            isPlayerInside = playerInside;
+        }
+
+        internal void Dispose()
+        {
+            OnAreaDisabled();
+            updateEventHandler.RemoveListener(IUpdateEventHandler.EventType.Update, Update);
+        }
+
+        private bool IsValidCameraMode(in CameraMode.ModeId mode)
```
pravusjif, line 132: How can the camera mode be invalid? I always find bit shifting/comparing a bit hard to read/maintain

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/CameraModeArea/CameraModeArea.cs
```diff
@@ -0,0 +1,176 @@
...
+
+            Collider[] colliders = Physics.OverlapBox(center, areaModel.area.box * 0.5f, rotation,
+                PhysicsLayers.avatarTriggerMask, QueryTriggerInteraction.Collide);
+
+            if (colliders.Length == 0)
+            {
+                return false;
+            }
+
+            return colliders.Any(collider => collider == playerCollider);
```
pravusjif, line 163: If this is the only `LINQ` usage in the script, can we turn that into a regular `for` structure and remove `LINQ`?

pravusjif (approved):
Looks good, I've been waiting for this feature for a long time, to apply it in Salmonomicon, good job!
Left a couple of questions

PR #1726, opened 2022-01-06
feat: add SSL to the WebSocket with self-signed certificate generated randomly at runtime

### LINKED ISSUE #1717: Find a way to connect the kernel to the renderer with websocket using .ORG
After the new security measures, The kernel is not able to connect to the renderer by WebSockets using `play.decentraland.org`

Temporary to solve this, we can use `play.decentraland.zone` but we need to figure out how to connect with `play.decentraland.org`.

kuruk-mm:
## What does this PR change?

Fix #1717

Just review this commit: https://github.com/decentraland/unity-renderer/pull/1726/commits/697a2a73c786bff885aad5d66520abdc5ce2d64a

Reminder: (I put this in the Debug.Log)
```
To be able to connect with SSL you should start Chrome with the --ignore-certificate-errors argument specified, or in Firefox set the configuration option `network.websocket.allowInsecureFromHTTPS to true, then use the ws:// rather than the wss:// address.
```

If we try to enter `play.decentraland.org` without SSL, we will get an error message.

Without SSL and `play.decentraland.zone` the port finder will only try port 5000.

And I set `play.decentraland.zone` without SSL as default

## Our Code Review Standards


pravusjif:
for launching chrome with flags on other OS
https://confluence.uconn.edu/ikb/desktop-support/software-and-applications/troubleshooting-certificate-errors-in-web-browsers

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/KernelCommunication/WebSocketCommunication/WebSocketCommunication.cs
```diff
@@ -26,37 +26,66 @@ public class WebSocketCommunication : IKernelCommunication
...
 
+        string wssUrl = wssServerUrl + wssServiceId;
         return wssUrl;
     }
 
     public WebSocketCommunication()
     {
         InitMessageTypeToBridgeName();
 
+        var debugConfig = GameObject.Find("DebugConfig").GetComponent<DebugConfigComponent>();
```
lorux0, line 84: i would try to find another way of getting `DebugConfigComponent` because finding objects by its name could lead into errors. You could have any other object called the same way in the scene or someone could change the name of it easily. It would be better to inject the component into `WebSocketCommunication` constructor. `DebugConfigComponent` could have its own singleton or a shared instance somewhere.
Also why you change the configuration here? I would be better to change it from `DebugConfigComponent` internally.

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/KernelCommunication/WebSocketCommunication/WebSocketCommunication.cs
```diff
@@ -26,37 +26,66 @@ public class WebSocketCommunication : IKernelCommunication
 
     private string StartServer(int port, int maxPort)
     {
+        var debugConfig = GameObject.Find("DebugConfig").GetComponent<DebugConfigComponent>();
```
lorux0, line 29: i would try to find another way of getting DebugConfigComponent because finding objects by its name could lead into errors. You could have any other object called the same way in the scene or someone could change the name of it easily. It would be better to inject the component into WebSocketCommunication constructor. DebugConfigComponent could have its own singleton or a shared instance somewhere.

lorux0 (approved):

pravusjif:
This is not working for me in macOS.

Tried the following:
1. changed the Baser Url Custom to .org
2. Toggled ON the Web Socket SSL setting
3. Toggled OFF the Open Browser When Start setting
4. Hit play in the editor to leave it waiting
5. opened chrome with the required flag (`/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --ignore-certificate-errors`) and entered https://play.decentraland.org/?NETWORK=mainnet&realm=loki&position=0%2C0&ws=ws://localhost:5000/dcl (changed the `ws=wss://localhost:5000/dcl` to `ws=ws://localhost:5000/dcl`)

If these steps are the ones needed to enter .org from the editor maybe we should put them in some documentation as well? (maybe not readme to avoid exposing it too much? I don't know)

Screen Shot 2022-01-07 at 15 21 45

kuruk-mm:
> This is not working for me in macOS.
> 
> Tried the following:
> 
> 1. changed the Baser Url Custom to .org
> 2. Toggled ON the Web Socket SSL setting
> 3. Toggled OFF the Open Browser When Start setting
> 4. Hit play in the editor to leave it waiting
> 5. opened chrome with the required flag (`/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --ignore-certificate-errors`) and entered https://play.decentraland.org/?NETWORK=mainnet&realm=loki&position=0%2C0&ws=ws://localhost:5000/dcl (changed the `ws=wss://localhost:5000/dcl` to `ws=ws://localhost:5000/dcl`)
> 
> If these steps are the ones needed to enter .org from the editor maybe we should put them in some documentation as well? (maybe not readme to avoid exposing it too much? I don't know)
> 
> Screen Shot 2022-01-07 at 15 21 45

We tested with Pravus, and we don't know precisely why Chrome is not opening with that parameter on Mac, and it only happens with Mac. So we tested enabling `chrome://flags/#allow-insecure-localhost` in Chrome, and it worked. I added that in the `Debug.Log` comment

PR #1730, opened 2022-01-06
fix: remove AttachToAvatar when Transform is added

pbosio:
`AttachToAvatar` and `Transform` components can't coexist in the same Entity.
We must make sure that `AttachToAvatar` is removed from the Entity when a `Transform` is added to it.

closes https://github.com/decentraland/sdk/issues/159


AjimenezDCL (approved):

pravusjif (approved):
"portal" scene still works fine after this change

PR #1766, opened 2022-01-13
feat: Start Menu improvements

### LINKED ISSUE #1755: Start Menu list of improvements
- [x] Player name tag should loose the double #0000 or the single when reserved name

- [x] Clicking on the black side of the header while in Map, triggers a Click that does nothing

- [x] World must be renderer somehow while changing graphic settings (maybe a little screen or window)
- [x] When we press ESC when a modal is open, should be closed only that modal and not the whole Start Menu
- [x] When pressing ESC in-world without anything else opened, it should open the start menu.
- [x] Remove the StartMenu tutorial step for users that already did the tutorial.

sandrade-dcl:
## What does this PR change?
Fix #1755 

- [x] Player name shouldn't show the address tag when it is a claimed name.
- [x] Clicking on the black side of the header while in Map, triggers a Click that does nothing.
- [x] World must be renderer somehow while changing graphic settings (maybe a little screen or window) -> **_DEACTIVATED TEMPORALLY_**
- [x] When we press ESC when a card modal is open (for example: place or event cards), should be closed only that modal and not the whole Start Menu.
- [x] When pressing ESC in-world without anything else opened, it should open the start menu.
- [x] Remove the StartMenu tutorial step for users that already did the tutorial.

## How to test the changes?
1. Go to: https://play.decentraland.zone/?renderer-branch=feat/Start-Menu-list-of-improvements
2. Test and check each point described above.

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/ExploreV2/Scripts/MainMenu/ExploreV2Menu/ExploreV2MenuComponentController.cs
```diff
@@ -467,17 +481,68 @@ internal bool NeedToRefreshRealms(IEnumerable<RealmModel> newRealmList)
...
     {
+        if (DataStore.i.exploreV2.isSomeModalOpen.Get())
+            return;
+
         if (isOpen.Get())
             exploreV2Analytics.SendStartMenuVisibility(false, fromShortcut ? ExploreUIVisibilityMethod.FromShortcut : ExploreUIVisibilityMethod.FromClick);
 
-        SetVisibility(false);
+        if (!fromShortcut)
+            SetVisibility(false);
```
pravusjif, line 499: since the `else` has brackets, can we add brackets here as well?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/ExploreV2/Scripts/MainMenu/ExploreV2Menu/ExploreV2MenuComponentController.cs
```diff
@@ -467,17 +481,68 @@ internal bool NeedToRefreshRealms(IEnumerable<RealmModel> newRealmList)
...
+        else
+        {
+            if (isOpen.Get())
+                SetVisibility(false);
```
pravusjif, line 503: since the `else` has brackets, can we add brackets here as well?

unity-renderer/Assets/Scripts/MainScripts/DCL/DataStore/DataStore_ExploreV2.cs
```diff
@@ -11,6 +11,7 @@ public class DataStore_ExploreV2
         public readonly BaseVariable<int> currentSectionIndex = new BaseVariable<int>(0);
         public readonly BaseVariable<bool> placesAndEventsVisible = new BaseVariable<bool>(false);
         public readonly BaseVariable<bool> profileCardIsOpen = new BaseVariable<bool>(false);
+        public readonly BaseVariable<bool> isSomeModalOpen = new BaseVariable<bool>(false);
```
pravusjif, line 14: I find the name of this variable a bit strange/confusing, can there be 2 modals open at the same time? If that's the case then maybe it's better to have a `List` of open modals here and then check the `Count()` for that list instead of this bool. Otherwise if you have 2 open modals and you close the 2nd, it would override this bool setting it as false when in reality there's still another modal open.

If only 1 modal can be opened at a time then I think it's fine as it is.

pravusjif (approved):
Nice improvement! I left a couple of comments.

On the build I only saw an issue in the Graphics Settings World Preview when using LOW specs (the SSAO appears to have an issue there, if you close the settings the SSAO works fine so it only happens in that world preview):
Screen Shot 2022-01-19 at 10 58 00

Personally I don't see the use of having a world preview if we don't have a "Save button" to apply the changes after seeing the "preview", because right now it's not actually a preview, it's the actual applied settings XD, but that's just me.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/ExploreV2/Scripts/MainMenu/ExploreV2Menu/ExploreV2MenuComponentController.cs
sandrade-dcl: I love your TOC ❤️ Thanks! Done in https://github.com/decentraland/unity-renderer/pull/1766/commits/2d28daf31911a106f173fd5a1b48b38783649ce1

unity-renderer/Assets/Scripts/MainScripts/DCL/DataStore/DataStore_ExploreV2.cs
sandrade-dcl: Yeah, I understand what go you mean. At the moment and inside the StartMenu's scenario we cannot have more than 1 modal open at the same time (at least we have not had the necessity yet), but if we would need it in the future then we would convert this variable to something more complex like you suggest. I'll take it into account, thanks for the feedback!

Suduck (approved):

sandrade-dcl:
> On the build I only saw an issue in the Graphics Settings World Preview when using LOW specs (the SSAO appears to have an issue there, if you close the settings the SSAO works fine so it only happens in that world preview): Screen Shot 2022-01-19 at 10 58 00

After analyze this problem with the SSAO (Brian confirmed me that it is something already investigated by him in the past and it seems to be an internal problem in Unity), we're going to think another way to show this world previsualization from Settings panel, but without using a window with a RenderTexture. Meanwhile, in order to avoid blocking this PR, we have decided to deactivate it temporally.

PR #1783, opened 2022-01-17
fix: main player comms reported rotation

### LINKED ISSUE #476: Fix users comms reported rotation
Through comms we always report the CAMERA rotation but we should be reporting the CHARACTER rotation, otherwise when players are using the third person camera, they think the rest of the users see them in the direction they point their avatar but they actually see the rotation of the camera...

pravusjif:
Fixes #476

**WHY**
Through comms we always report the CAMERA rotation but we should be reporting the CHARACTER rotation, otherwise when players are using the third person camera, they think the rest of the users see them in the direction they point their avatar but they actually see the rotation of the camera...

**WHAT**
Re-implemented mini fix originally implemented by Santi at https://github.com/decentraland/unity-renderer/pull/864


lorux0 (approved):
works great, good job!

Suduck (approved):

sandrade-dcl (approved):

PR #1784, opened 2022-01-18
feat: display debug bounding boxes

pbosio:
Display debug bounding boxes of entities shapes.

Screen Shot 2022-01-17 at 22 49 39

How to test:
1: goto https://play.decentraland.zone/?position=-143,-125&kernel-branch=feat/show-shapes-bounding-box&renderer-branch=feat/show-shapes-bounding-box
2: open browser's console and put `clientDebug.ToggleSceneBoundingBoxes('-143,-125',true)`
bounding boxes should be visible now
3: now put in the console `clientDebug.ToggleSceneBoundingBoxes('-143,-125',false)`
bounding boxes should have disappear


olavra:
Does the command need to receive the coordinates? couldnt it just simply enable it for everything?

pbosio:
> Does the command need to receive the coordinates? couldnt it just simply enable it for everything?

The command supports adding multiple scenes by coordinates or sceneId.
It's always better to have more control, since having a lot of noise makes debugging more difficult.
In any case, activating for all visible scenes would be as easy as writing this in the console
```
for(let sceneId of globalThis.sceneWorkers.keys()){
    if (sceneId !== "dcl-gs-avatars")
        clientDebug.ToggleSceneBoundingBoxes(sceneId,true)
}
```
so, if needed, we can also add it as another `clientDebug` command in kernel

unity-renderer/Assets/DCLPlugins/DebugPlugins/ShapesBoundingBoxDisplayer/EntityWireframe.cs
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    void IShapeListener.OnShapeUpdated(IDCLEntity entity)
+    {
+        entityWireframe ??= Object.Instantiate(wireframeOriginal);
+
+        Transform wireframeT = entityWireframe.transform;
+
+        wireframeT.position = entity.meshesInfo.mergedBounds.center;
+        wireframeT.localScale = entity.meshesInfo.mergedBounds.size * 1.01f;
```
Suduck, line 30: can we move this to a constant? This way we have more info about what is this magic number

Suduck: On another hand , we are not setting the rotation of the wireframe, if the entity is rotated in different axies, will it work?

Suduck (approved):

sandrade-dcl (approved):
Very useful tool!

unity-renderer/Assets/DCLPlugins/DebugPlugins/ShapesBoundingBoxDisplayer/EntityWireframe.cs
pbosio: no need to set rotation there.
size and position are realted to the `mergedBounds`. After those lines the wireframe is added as child of the entity's `gameObject` so it will rotate with the entity

PR #1811, opened 2022-01-20
chore: toggle non animator meshes animation with mesh visibility

pravusjif:
**WHY**
Currently when a GLTFShape visibility is changed, if it has an Animation, it always keeps being animated. For example if a mesh visibility is turned off, its animation keeps animating unnecessarily.

This should only apply to meshes on entities that don't have an Animator component (since in those cases the parcel devs don't have control over the animations so they wouldn't care of its state).

**WHAT**
Added Animation to MeshesInfo and added animation toggling on mesh visibility toggling, only for entities that don't have an Animator component.


BrianAmadori (changes requested):
I think we should add some tests to validate this behaviour keeps working in the future.

Suduck (approved):
Codewise looks good

davidejensen (approved):

BrianAmadori (approved):

PR #1817, opened 2022-01-21
fix: main player comms reported rotation

### LINKED ISSUE #476: Fix users comms reported rotation

pbosio:
Fixes #476

WHY
Through comms we always report the CAMERA rotation but we should be reporting the CHARACTER rotation, otherwise when players are using the third person camera, they think the rest of the users see them in the direction they point their avatar but they actually see the rotation of the camera...

<s>kernel PR need to be merged first: https://github.com/decentraland/kernel/pull/181</s>

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
@@ -524,6 +524,7 @@ public class DCLCharacterController : MonoBehaviour
         var reportPosition = characterPosition.worldPosition + (Vector3.up * height);
         var compositeRotation = Quaternion.LookRotation(characterForward.HasValue() ? characterForward.Get().Value :cameraForward.Get());
         var playerHeight = height + (characterController.height / 2);
+        var cameraRotation = Quaternion.LookRotation(cameraForward.Get());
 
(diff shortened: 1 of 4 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
```diff
@@ -532,7 +533,7 @@ void ReportMovement()
         //                  - Scenes not being sent for loading, making ActivateRenderer never being sent, only in WSS mode.
         //                  - Random teleports to 0,0 or other positions that shouldn't happen.
         if (initialPositionAlreadySet)
-            DCL.Interface.WebInterface.ReportPosition(reportPosition, compositeRotation, playerHeight);
+            DCL.Interface.WebInterface.ReportPosition(reportPosition, compositeRotation, playerHeight, cameraRotation);
```
BrianAmadori, line 536: Please add some tests to avoid future regressions, as this change already broke the client in the past.

BrianAmadori (changes requested):

AjimenezDCL (approved):

kuruk-mm:
This PR fix https://github.com/decentraland/unity-renderer/issues/1822

BrianAmadori (approved):

PR #1821, opened 2022-01-24
refactor: gif processing for better performance and multithreading

### LINKED ISSUE #1842: Refactor UniGif for Perfomance
Right now WebGL build has a Gif Processor that works natively on the browser, this won't work on Desktop nor Unity, so in the past, we added UniGif, a library to turn a gif file into multiple textures.

Sadly UniGif is slow, processing a gif of 2Mb resulted in a 10000ms delay of the main thread so we need to refactor that system to be more performant by using the power of multithreading.

Kinerius:
## What does this PR change?

UniGif was deleted, now we use Unity Gif Decoder from https://github.com/3DI70R/Unity-GifDecoder

Fixes #1842 

## How to test the changes?

Desktop Only:
Run through the world and pass-by scenes with lots of gifs, there shouldn't be performance spikes.

A good example is -14, -12 ( it has 4 gifs and a broken URL)

WebGL:
Should work like before

Unity:
There's a test scene named UniGifStressTest, you can try it by toggling the multithreaded bool to see the difference in performance using the profiler.

## Our Code Review Standards


kuruk-mm (commented):
Can we add the Unity-GifDecoder as a UPM package?

We can fork the repo if you have modified something from the original source code.

kuruk-mm (approved):

PR #1843, opened 2022-01-26
fix: added OS based system to force scroll rect sensibility

### LINKED ISSUE #1819: Scrolling through passport wearables is not reponsive
Current list of collectibles in a player passport can be scrolled using the mouse wheel, but the speed is too slow to be usable.
It would be good to check other parts of the application, like the backpack, explore or chat and reconsider scrolling speed.
We also need to have in consideration different scrolling method, depending of the OS, or device (trackpad, mouse wheel)

**Steps to reproduce**
- Find a player with custom wearables
- Open its passport
- Select Collectibles
- Scroll using the mouse wheel
- The movement speed of the list is ridiculous

**Expected behaviour**
List should be scrollable in a meaningfull way using the speed of the wheel to determine distance

davidejensen:
## What does this PR change?

Fix #1819
This PR adds an initial simple system to add a component to scroll rects to force the scroll sensitivity based on the OS in use, the scroll values are defined as BaseVariable in DataStore_HUDs

...

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=fix/scroll-sensitivity-in-player-card-collectibles
2. Test the 2 following places where the component has been used:
* With different OS try to scroll in the chat box in the bottom left
* Find a player with more then 6 customizable items and inspect them, go to the collectible tab in the player card and test the scroll behavior

## Our Code Review Standards


DavideJensen:
Made the changes, still needs to be tested on different platforms, I tryed already mac and windows with different mouse

sandrade-dcl (approved):
The code LGTM!

Anyway, before merge it, I think we should do an analysis of all the existing scrolls that are working in the project and apply the `ScrollRectSensitivityHandler` to each of them (I'm sure we're missing a lot of them that don't have the script applied). Now that we have a different sensitivity configuration for each OS, I think it is the moment to decide a correct value for each one and apply it everywhere. In my case (I'm using Windows) I feel that the scroll moves very slow with the current configured values.

I approve the PR in order to not block it, but please let's talk about this in the next daily, ok?

Thanks and good work!

DavideJensen:
> The code LGTM!
> 
> Anyway, before merge it, I think we should do an analysis of all the existing scrolls that are working in the project and apply the `ScrollRectSensitivityHandler` to each of them (I'm sure we're missing a lot of them that don't have the script applied). Now that we have a different sensitivity configuration for each OS, I think it is the moment to decide a correct value for each one and apply it everywhere. In my case (I'm using Windows) I feel that the scroll moves very slow with the current configured values.
> 
> I approve the PR in order to not block it, but please let's talk about this in the next daily, ok?
> 
> Thanks and good work!

@sandrade-dcl yeah the PR is not ready to be merged because I still need to do additional testing on Linux OS, we can talk tomorrow about this as a post daily if you want

sandrade-dcl:
> @sandrade-dcl yeah the PR is not ready to be merged because I still need to do additional testing on Linux OS, we can talk tomorrow about this as a post daily if you want

Ok! Let's talk tomorrow about this 👍

DavideJensen:
List of scroll rectangles on which to apply the new scroll behavior:

- [ ] Collectibles panel
- [ ] Chat
- [ ] Event list 
- [ ] Explore -> highlights
- [ ] Explore -> places
- [ ] Explore -> events
- [ ] Backpack -> collectibles
- [ ] Settings -> graphics

BrianAmadori (approved):
The applied settings work fine on windows

Issue #1759, opened 2022-01-12
Desktop: The app crash on close


kuruk-mm:
In Windows, the App Crash almost every time that you close, or close with Alt+F4.

Probably it's the following:
```
Setting up 3 worker threads for Enlighten.
  Thread -> id: 2aac -> priority: 1
  Thread -> id: 37d8 -> priority: 1
  Thread -> id: 19dc -> priority: 1
System.NullReferenceException
  at (wrapper managed-to-native) UnityEngine.Behaviour.set_enabled(UnityEngine.Behaviour,bool)
  at DCL.SettingsPanelHUD.SettingsPanelHUDView.SetTutorialButtonEnabled (System.Boolean isEnabled) [0x00000] in <366503faae314bb29313bb5b71c6d75e>:0
  at DCL.SettingsPanelHUD.SettingsPanelHUDController.SetTutorialButtonEnabled (System.Boolean isEnabled) [0x00006] in <366503faae314bb29313bb5b71c6d75e>:0
  at DCL.Tutorial.TutorialController.SetTutorialDisabled () [0x000dd] in <4e86139f7b554f179d1fe866dc0fbca1>:0
  at DCL.Tutorial.TutorialController.Dispose () [0x00000] in <4e86139f7b554f179d1fe866dc0fbca1>:0
  at DCL.PluginInfo.Disable () [0x00009] in <2de2437650fd4c76ae0bbe1c8f55d608>:0
  at DCL.PluginSystem.Dispose () [0x00022] in <2de2437650fd4c76ae0bbe1c8f55d608>:0
  at DCL.Main.OnDestroy () [0x00044] in <2830c777109f440ca91d7504a1b86bcd>:0
  at DCL.MainDesktop.OnDestroy () [0x00000] in <d48a4262455a42c8b52d1d5139cdabca>:0
```

kuruk-mm:
We're going to kill the process to avoid crash reports.

The behavior on the Web when you close the tab is similar to killing the process. We don't have data that we can lose when you close the App by force.

This can be temporary until we fix the closing pipeline.

PR #1846, opened 2022-01-26
fix: avoid the websocket server to close inactive connections

kuruk-mm:
fix: SetTutorialButtonEnabled is used after the view is disposed of, add check to avoid null reference

Fix partially: https://github.com/decentraland/unity-renderer/issues/1759

## Our Code Review Standards


Kinerius (approved):
Nice

PR #1853, opened 2022-01-27
feat: add `excludeIds` field to `AvatarModifierArea`

pbosio:
* added avatar's `collider` to `Player` model class
* added `excludeIds` field to `AvatarModifierArea` component.

Note: Modifiers won't affect avatars for users which id are contained in `excludeIds`


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/AvatarModifierArea.cs
```diff
@@ -117,10 +129,13 @@ private HashSet<GameObject> DetectAllAvatarsInArea()
 
         Vector3 center = entity.gameObject.transform.position;
         Quaternion rotation = entity.gameObject.transform.rotation;
-        return cachedModel.area?.DetectAvatars(center, rotation);
+        return cachedModel.area?.DetectAvatars(center, rotation, excludedColliders);
```
AjimenezDCL, line 132: I would inverse the logic here. It feels more natural to have a function like `DetectAvatars(center, rotation, avatars)` than an exclusion list

AjimenezDCL (dismissed):

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/AvatarModifierArea.cs
```diff
@@ -152,8 +170,84 @@ private void ApplyCurrentModel()
...
+        if (!excludeIds.Contains(player.id))
+        {
+            return;
+        }
+
+        if (excludedColliders == null)
+        {
+            excludedColliders = new HashSet<Collider>() { player.collider };
+        }
+        else if (!excludedColliders.Contains(player.collider))
```
pravusjif, line 248: I think this check can be avoided right? since the collider can only exist once in the hashset

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/TriggerArea.cs
```diff
@@ -25,6 +28,10 @@ public override HashSet<GameObject> DetectAvatars(Vector3 center, Quaternion rot
         HashSet<GameObject> result = new HashSet<GameObject>();
         foreach (Collider collider in colliders)
         {
+            if (excludeColliders != null && excludeColliders.Contains(collider))
```
pravusjif, line 31: since null checks can be a bit costly, can we cache the `excludeColliders != null` in a bool just before the foreach? and use that bool inside the foreach, what do you think?

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/AvatarModifierArea.cs
pravusjif: Uhmmm I think in the end that would be almost the same, because the provided `avatars` parameter  value would be like an "allowed avatars", and would also have to be checked on every collider found in the trigger area. I find it more natural that the `DetectAvatars` detects everything (not only the allowed avatars) and filters out the excluded ones, as it is in the current implementation, but it's just my opinion

pravusjif (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/TriggerArea.cs
pbosio: 👍 Changed here: https://github.com/decentraland/unity-renderer/pull/1853/commits/343a0512997ef9de3670de0da08f71f8c6a0780b

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/AvatarModifierArea.cs
pbosio: yeah, you are right! I thought that `add` could throw an exception if element is already there (as Dictionary does) but it doesn't.
Changed here: https://github.com/decentraland/unity-renderer/pull/1853/commits/343a0512997ef9de3670de0da08f71f8c6a0780b

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/AvatarModifierArea.cs
pbosio: Having an "allow list" would make the code more complex:
1- for backward compatibility we will need to check if the allow list exists in the component and that it is valid.
2- since the sdk component is designed to have an exclusion list, we will need to convert that exclusion list into an allow list every time an avatar shape is instantiated.
we decided to have an exclusion list since it's the most common use case, to avoid, in most scenarios, the scene dev to keep track of every avatar that is inside the scene and modify the list accordingly constantly. This also means that the component would be updated less frequently

AjimenezDCL: I'm just talking about the `DetectAvatars` function. Everything would be the same but that function receive the Avatar colliders to detect in that area.

You just get all the avatars, remove the exclusion list and pass it to the area to detect

pbosio: `DetectAvatars` does a collision test with everything in `AVATAR_TRIGGER_LAYER` layer
we are not handling any list of avatar's colliders at this point
except for filtering that collision test result with `excludedColliders`

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/AvatarModifierArea.cs
```diff
@@ -36,6 +42,8 @@ public AvatarModifierArea()
             { "DISABLE_PASSPORTS", new DisablePassportModifier() }
         };
         model = new Model();
+
+        DataStore.i.player.otherPlayers.OnRemoved += OnOtherPlayersRemoved;
```
BrianAmadori, line 46:
```suggestion
        DataStore.i.player.otherPlayers.OnRemoved += OnOtherPlayersRemoved;
```

I think that this is not symmetric with `OnDestroy`. The constructor can be called and the `OnDestroy` won't. For monobehaviours it's safer to use `Awake()` and `OnDestroy()` for cases like this.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/AvatarModifierArea.cs
BrianAmadori: @pbosio and @AjimenezDCL agreed on this feedback not applying for this PR. Marking this conversation as solved.

BrianAmadori (approved):

PR #1854, opened 2022-01-27
refactor: move components to assemblies

BrianAmadori:
## What does this PR change?

This PR doesn't change any code. But restructures the assemblies so the SDK component implementations can be moved to individual assemblies. 

Core changes:
* `BaseVariable` classes were incorrectly placed in `ScriptableObject` assembly. Moved them over to `BaseVariable` assembly.
* moved `BaseShape` to its own assembly
* moved `CollidersManager` to its own assembly (and placed `CollidersInfo` in this same assembly as well)
* moved `BaseComponent` and `BaseDisposable` to `WorldRuntimeInterfaces` assembly (maybe the assembly name doesn't make much sense now?)
* Moved `PrimitiveMeshBuilder` to `Utils`.

This allowed moving the following components to their own assemblies:
* TextShape
* DCLFont
* ParametrizedShape
* Billboard

## Our Code Review Standards


Kinerius (approved):
Nice!

Issue #1357, opened 2021-10-18
Move SDK components to individual assemblies to finish cleanup of MainScripts


BrianAmadori:
This is needed to close https://github.com/decentraland/explorer-technical-backlog/issues/12

This is also needed to cleanly address the new SDK 7 version

- [x] LoadableShape (GLTFShape, OBJShape, NFTShape)
- [x] ParametrizableShape (BoxShape, SphereShape, CylinderShape, etc)
- [x] DCLTransform
- [x] Builder in world components
- [x] All UI components
- [x] TextShape
- [x] DCLFont
- [x] Audio components
- [x] Animation components
- [x] UUID event components

PR #1862, opened 2022-01-28
refactor: move audio components to own assembly

BrianAmadori:
## What does this PR change?

This PR moves audio components to their own assembly. Had to remove `GlobalScene` references from audio components to make this work and replaced them with equivalent behaviour.

Actionable for https://github.com/decentraland/unity-renderer/issues/1357

## How to test the changes?

We may want to test portable experiences that use sound and enable they play the sound correctly, although this should be already covered by tests.

Steps for test:
1. Enter the world.
2. Go to Google console (F12).
3. Write `spawnScenePortableExperienceSceneFromUrn("urn:decentraland:off-chain:static-portable-experiences:radio", "main")`
4. You will see the PEX is executed (don't worry about the infinite loading spinner, it is because it is a old PEX).
image
5. Select any station in the PEX UI.
6. The music is not playing, when it should.

You can also test it with the "Portal" PEX (`spawnScenePortableExperienceSceneFromUrn("urn:decentraland:off-chain:static-portable-experiences:portal-experience", "main")`), when a sound should be playing each time we shoot:
image

## Our Code Review Standards


sandrade-dcl (changes requested):
Removing the logic where we were checking the `isPortableExperience` flag, the audio in the Portable Experiences has stopped working. We should to keep this logic.

Steps for reproduce it:
1. Enter the world.
2. Go to Google console (F12).
3. Write `spawnPortableExperienceScene("urn:decentraland:off-chain:static-portable-experiences:radio", "main")`
4. You will see the PEX is executed (don't worry about the infinite loading spinner, it is because it is a old PEX).
image
5. Select any station in the PEX UI.
6. The music is not playing, when it should.

You can also test it with the "Portal" PEX (`spawnPortableExperienceScene("urn:decentraland:off-chain:static-portable-experiences:portal-experience", "main")`), when a sound should be playing each time we shoot:
image

Suduck (approved):
Codewise LGTM!

pravusjif (commented):
The code looks good, but the Portal PEX audio **is not working**, although the radio PEX does work. In master build both work fine.
Also tested other audio from scenes in-world and it works fine.

sandrade-dcl (changes requested):
As Pravus also said, the audio of the **Portal PEX** continues without working (I guess because of some of the changes applied in the `DCLAudioSource` class).

Maybe the problem is this code block that you removed in the `UpdateAudoSourceVolume()` function?

```
if (scene is GlobalScene globalScene && globalScene.isPortableExperience)
{
    audioSource.volume = newVolume;
    return;
}
```

The audio of the **Radio PEX** is working fine, but I guess it is because it is a stream audio, that is different to the audio reproduced by the Portal PEX.

BrianAmadori:
You were right @sandrade-dcl! But the thing is I cannot reference `GlobalScene` because of assembly dependency issue. So i used the `isPersistent` flag instead.

sandrade-dcl (approved):
It is working fine now! Good work!

PR #1863, opened 2022-01-28
refactor: move builder components to own assembly

BrianAmadori:
## What does this PR change?

This PR moves Gizmos and SmartItem components under the same assembly named `DCL.Components.BuilderInWorld`

Actionable for #1357 

## Our Code Review Standards


Suduck (changes requested):
The DCLGizmos is not part of builder in world, is part of the old builder. We should move to the older builder since in the future it will be deleted (probably) and we don't want to maintain code that we don't use

BrianAmadori:
It's ok if I just rename the assembly to `Builder` and still put both components on the same assembly? As in any case we will end up removing the gizmos class.

Suduck (approved):

Suduck:
Then it's ok I guess!

PR #1874, opened 2022-02-01
chore: avatar visibility now uses a flag based approach

AjimenezDCL:
## What does this PR change?
Avatar visibility was not really scalable. Adding layers meant adding a new method to both `IVisibility` and `Visibility`. The raising need of different systems to switch an avatar visibility were causing collisions on the visibility state.

Now Visibility just handle constrains that can be added or removed by other systems so an avatar is only visible when all its constrains are cleared.

## How to test the changes?
Avatars should behave as usual

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/AvatarSystem/Definitions/IVisibility.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/AvatarSystem/Definitions/IVisibility.cs
@@ -6,9 +6,13 @@ namespace AvatarSystem
     public interface IVisibility : IDisposable
     {
         void Bind(Renderer combinedRenderer, Renderer[] facialFeatures);
-        void SetExplicitVisibility(bool explicitVisibility);
-        void SetLoadingReady(bool loadingReady);
-        void SetCombinedRendererVisibility(bool combinedRendererVisibility);
-        void SetFacialFeaturesVisibility(bool facialFeaturesVisibility);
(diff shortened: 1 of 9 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/AvatarSystem/Definitions/IVisibility.cs
```diff
@@ -6,9 +6,13 @@ namespace AvatarSystem
...
-        void SetCombinedRendererVisibility(bool combinedRendererVisibility);
-        void SetFacialFeaturesVisibility(bool facialFeaturesVisibility);
+        public void AddGlobalConstrain(string key);
+        public void RemoveGlobalConstrain(string key);
+
+        public void AddCombinedRendererConstrain(string key);
+        public void RemoveCombinedRendererConstrain(string key);
+
+        public void AddFacialFeaturesConstrain(string key);
+        public void RemoveFacialFeaturesConstrain(string key);
```
Suduck, line 16: interfaces shouldn't have public method

Suduck (approved):
Good job!

unity-renderer/Assets/Scripts/MainScripts/DCL/AvatarSystem/Definitions/IVisibility.cs
AjimenezDCL: wtf is wrong with me... nice catch

PR #1875, opened 2022-02-02
fix: avoid fps affecting movement speed

### LINKED ISSUE #1753: Avatar movement speed is affected by FPS
Considering the context of low FPS:
- Scene Load raidus = 6
- 100 avatars around you
- Ultra settings

The avatar movement speed is considerably slower than on good FPS conditions.

**Acceptance Criteria**
- [ ] Avatar walk / run speed is always constant independently of the FPS

DavideJensen:
## What does this PR change?

Fix #1753
With this PR the avatar movement speed should not be influenced by FPS anymore

...

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=fix/fps-affecting-avatar-movement
2. Go to a crowded area and maximize the settings
3. Walk around
4. Lower the settings to minimum
5. Walk around
6. Verify the consistency between speeds

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
@@ -252,7 +252,7 @@ public class DCLCharacterController : MonoBehaviour
 
     internal void LateUpdate()
     {
-        deltaTime = Mathf.Min(deltaTimeCap, Time.deltaTime);
+        deltaTime = Time.deltaTime;//Mathf.Min(deltaTimeCap, Time.deltaTime);
 
         if (transform.position.y < minimumYPosition)
         {
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
@@ -35,8 +35,6 @@ public class DCLCharacterController : MonoBehaviour
 
     new Collider collider;
 
-    float deltaTime = 0.032f;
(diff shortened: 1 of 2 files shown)


DavideJensen:
I still miss the whole point of why a fixed delta time has been used in the past considering that in other context Time.deltaTime is used without any worries, probably was due to legacy stuff, now everything looks like it's properly working

AjimenezDCL (approved):
It was introduced with the messages budget, but I wonder why. Good job!

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
```diff
@@ -268,7 +264,7 @@ internal void LateUpdate()
         {
             velocity.x = 0f;
             velocity.z = 0f;
-            velocity.y += gravity * deltaTime;
+            velocity.y += gravity * Time.deltaTime;
```
BrianAmadori, line 267: IIRC, the reason of capping the `deltaTime` was to avoid the player "teleporting" in cases of big hiccups. So maybe it would be wise to keep deltaTime capping logic but with a bigger threshold?

DavideJensen: We can talk about this in the daily/post daily today

AjimenezDCL: What would cause the teleporting? 

For the local player, In a hiccup the engine will stop processing input, wont it?
For another player the position won't be sent to kernel and in any case we have the interpolation.

Am I missing something?

We could also manually introducing a big hiccup and check the behaviour.

BrianAmadori: We already talked about it and concluded that this may not be a big enough issue to justify the cap for now. We are going to test this further to be sure. Resolving this conversation.

PR #1902, opened 2022-02-09
chore: simplify asset bundles loading toggling

pravusjif:
* Made scene ABs and wearable ABs usage more consistent with feature flags
* Refactored to have an easier way to avoid ABs for the desktop client

WIP: Still have to solve assembly definition files nightmare hell to finish


AjimenezDCL:
What about this, are we gonna finish it anytime soon? @pravusjif

pravusjif:
For now we can keep using the asset bundle's url param in desktop client.

PR #1911, opened 2022-02-10
fix: characters disclaimer copy

RominaMarchetti:
## What does this PR change?
The text indicating which characters are not allowed was override from this `non-alphanumeric characters or spaces allowed`
to this `No alphanumeric characters or spaces allowed`.

## How to test?
1- Play as Guest in the following link: https://play.decentraland.zone/?renderer-branch=fix/AlphanumericCopy
2- In the second step of creating your account, try to use space or alphanumeric characters in the Name input.

Screen Shot 2022-02-10 at 14 33 24


nlorux:
The phrase says "no alphanumeric characters allowed" but is confusing to me, since you can write `oiejf235` as your name, so alphanumerics are allowed. Only special characters or "non-alphanumeric" are not allowed, like #@$^ or spaces

PR #1923, opened 2022-02-11
fix: Name input warning for special characters

RominaMarchetti:
The string `non-alphanumeric characters or spaces allowed` has been replaced by `Letters and numbers only`
2022-02-10_17h23_04

### How to test:
Enter this link and create a new account (go as Guest, is easier).
When you try to use an special character or spaces in your name this new message should appear under the text field.


RominaMarchetti:
@lorux0 Louise, who is a native English speaker, helped me with the copy.
You were right about that I was changing the meaning of the sentece for the oposite D:

davidejensen (approved):

PR #1925, opened 2022-02-14
fix: scenes stuck loading on GLTF promises silent forget

### LINKED ISSUE #1890: Debug bug of scenes that never finish loading (with asset bundles disabled)

pravusjif:
This PR fixes #1890 

**WHY**
* When traversing long distances without stopping, (as shown in #1890 with the railroad scene) the user can be loading and unloading a lot of scenes around him very quickly. In that scenario, many asset promises get cancelled while loading.
* When a same asset is requested from different scenes, if the first instance of the asset promise is still loading, the subsequent ones get into a WAITING blocked state that gets resolver AFTER the first one finishes downloading the asset and loading it. Those first asset promises are what we call **master promises**, and when those promises [are told to be forgotten](https://github.com/decentraland/unity-renderer/blob/ab815ad8c95bf5ddd1fb43dc80a799e28ea72af0/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Common/AssetPromiseKeeper.cs#L114) while still downloading/loading (for example when unloading the scene that requested it), if they are blocking other promises, we forget them in a special way we call ["silent forget"](https://github.com/decentraland/unity-renderer/blob/ab815ad8c95bf5ddd1fb43dc80a799e28ea72af0/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Common/AssetPromiseKeeper.cs#L135) that flags the promise as forgottten but allows it to finish downloading so that it unblocks the other promises that depend on that one
* The problem we were having is that when a far away scene gets unloaded (and its loading promises are forgotten) the scene gameobjects get deactivated, and since we were triggering the [`LoadAssetCoroutine` coroutine](https://github.com/decentraland/unity-renderer/blob/ab815ad8c95bf5ddd1fb43dc80a799e28ea72af0/unity-renderer/Assets/UnityGLTF/Scripts/GLTFComponent.cs#L222) from the [GLTFComponent monobehaviour itself](https://github.com/decentraland/unity-renderer/commit/fa747cbe5d378af438b718ca10dd5f8bf9ec4493), the running coroutine that [was waiting](https://github.com/decentraland/unity-renderer/blob/ab815ad8c95bf5ddd1fb43dc80a799e28ea72af0/unity-renderer/Assets/UnityGLTF/Scripts/GLTFComponent.cs#L296) to get a chance to download the promise asset, gets interrupted (since the GO gets deactivated by the scene unloading).
* By using the `CoroutineStarter` as we always do, the problem is solved because the coroutine that waits and ends up finishing the asset loading is never interrupted due to the scene unloading.
* The bug was more evident without using ABs, that's why it came out while developing the desktop client

**WHAT**
Fixed coroutines being started from monobehaviours (that were getting deactivated on silent forgets) instead of CoroutineStarter

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/GLTF/Tests/BlockedAndMasterPromisesShould.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/GLTF/Tests/BlockedAndMasterPromisesShould.cs
@@ -70,6 +70,60 @@ namespace AssetPromiseKeeper_GLTF_Tests
             Assert.IsTrue(keeper.library.Contains(asset));
             Assert.AreEqual(1, keeper.library.masterAssets.Count);
         }
+        
+        [UnityTest]
+        public IEnumerator SucceedWhenMastersParentIsDeactivated()
+        {
+            var keeper = new AssetPromiseKeeper_GLTF();
+            keeper.throttlingCounter.enabled = false;
+
+            string url = TestAssetsUtils.GetPath() + "/GLB/Lantern/Lantern.glb";
+            GameObject parent = new GameObject("parent");
+
+            AssetPromise_GLTF prom = new AssetPromise_GLTF(scene.contentProvider, url);
+            prom.settings.parent = parent.transform;
+
+            AssetPromise_GLTF prom2 = new AssetPromise_GLTF(scene.contentProvider, url);
+            bool failEventCalled2 = false;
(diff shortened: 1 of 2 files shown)


Kinerius (approved):
Nice catch!!

BrianAmadori (approved):
Awesome fix. Would love to get a small test to cover the case. (Maybe load a GLTF and deactivate the `GLTFComponent` gameObject after a frame?).

sandrade-dcl (approved):

BrianAmadori (approved):

kuruk-mm (approved):

PR #1942, opened 2022-02-17
fix: desktop crash on application quit

pravusjif:
1. `Main` now sets the `DataStore.i.common.isWorldBeingDestroyed` in true on the `Application.quitting` event.
2. Every other place we need to check if the application is quitting we just check that datastore value
3. Debugged in desktop client to see which points made the build crash on application quit and added the checks there. Tried to centralize the checks as much as possible to avoid duplicate checks.
4. Even though in-editor there are some nullrefs on playmode exit, I tested that those don't provoke a crash in the desktop application


unity-renderer/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -207,6 +207,11 @@ private GLTFSceneImporter(ILoader externalDataLoader)
 
         public void Dispose()
         {
+#if UNITY_STANDALONE || UNITY_EDITOR
+            if (DataStore.i.common.isWorldBeingDestroyed.Get())
+                return;
+#endif
+
```
Kinerius, line 214: What if we just put only one of these returns at the `catch` of `LoadScene()`? So it doesn't throw. 
Every Exception ends there.

unity-renderer/Assets/DCLPlugins/BuilderInWorld/Scripts/BuilderInWorldAudioHandler.cs
```diff
@@ -228,7 +228,15 @@ private void OnEntityBoundsCheckerStatusChanged(DCL.Models.IDCLEntity entity, bo
 
     private bool EntityHasBeenAddedSinceLastUpdate() { return (entityHandler.GetCurrentSceneEntityCount() > entityCount); }
 
-    private void OnDestroy() { Dispose(); }
+    private void OnDestroy()
+    {
+#if UNITY_STANDALONE || UNITY_EDITOR
+        if (DataStore.i.common.isWorldBeingDestroyed.Get())
```
BrianAmadori, line 234: I'd rename this to `isApplicationQuitting`. 

`isWorldBeingDestroyed` can mean too many things.

BrianAmadori (approved):

unity-renderer/Assets/DCLPlugins/BuilderInWorld/Scripts/BuilderInWorldAudioHandler.cs
Suduck: Agree! `isWorldBeingDestroyed `sound like `IsInMordor`, it could mean too many things xD

Suduck: Another proposal could be something like 

`isApplicationOnShutdown`

Suduck (approved):

PR #1953, opened 2022-02-20
test: curved world effect in shader

pravusjif:
added basic curved world effect to lit and triplanar shaders


AjimenezDCL:
Closing this but leaving the branch just in case we want to come back to take a look at it.

PR #1982, opened 2022-03-04
chore: move GLTFs texture resize to the GPU for desktop client

pravusjif:
* In many parts of the world we are resizing every texture loaded that is bigger than 1024x1024. We even have lots of textures in the first base wearables that get resized every time we enter the world if any avatar is using them. (With Asset Bundles this resizing is avoided as those textures were already resized when converting the asset to AB).

* If we can move to the GPU the textures resize we do in runtime for the Desktop Client, it will run smoother now that we have removed the runtime compressions for the Desktop Client as well.

* This optimization of using `Graphics.CopyTexture()` (copying everything directly in the GPU without passing through the CPU) instead of reading from the `RenderTexture` with `Texture2d.ReadPixels()` + `Texture2d.Apply()` that's quite expensive, can only be done for the Desktop client as it supports using the CopyTexture() method and the Web client doesn't.

A place where textures resizing can be seen is at:
https://play.decentraland.zone/?renderer-branch=chore/MoveGLTFsTextureResizeToGPUForDesktop&DISABLE_ASSET_BUNDLES&DISABLE_WEARABLE_ASSET_BUNDLES&position=-32,68&LOS=2

If these images can be seen, the resizing worked fine:
chrome_zBmTglA1LL

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/TextureHelpers/TextureHelpers.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/TextureHelpers/TextureHelpers.cs
@@ -67,8 +67,28 @@ public static class TextureHelpers
     {
         Texture2D texture = new Texture2D(sourceTexture.width, sourceTexture.height, sourceTexture.format, false);
         
-        // TODO: WebGL doesn't support this, should we add the ReadPixels/GetPixels approach as in Resize() ?
-        Graphics.CopyTexture(sourceTexture, texture);
+        bool supportsGPUTextureCopy = SystemInfo.copyTextureSupport != CopyTextureSupport.None;
+        if (supportsGPUTextureCopy)
+        {
+            Graphics.CopyTexture(sourceTexture, texture);
+        }
+        else
+        {
+            RenderTexture rt = RenderTexture.GetTemporary(sourceTexture.width, sourceTexture.height);
+            rt.filterMode = FilterMode.Point;
+            FilterMode sourceFilterMode = sourceTexture.filterMode; 
+            sourceTexture.filterMode = FilterMode.Point;
+
+            RenderTexture.active = rt;
+            Graphics.Blit(sourceTexture, rt);
+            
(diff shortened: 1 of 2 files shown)


BrianAmadori (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/TextureHelpers/TextureHelpers.cs
```diff
@@ -54,7 +66,10 @@ public static Texture2D Resize(Texture2D source, int newWidth, int newHeight)
     public static Texture2D CopyTexture(Texture2D sourceTexture)
     {
         Texture2D texture = new Texture2D(sourceTexture.width, sourceTexture.height, sourceTexture.format, false);
-        Graphics.CopyTexture(sourceTexture, texture); // TODO: does this work in WebGL?
+        
+        // TODO: WebGL doesn't support this, should we add the ReadPixels/GetPixels approach as in Resize() ?
+        Graphics.CopyTexture(sourceTexture, texture);
```
kuruk-mm, line 71: Should we change this with:

```
        bool supportsGPUTextureCopy = SystemInfo.copyTextureSupport != CopyTextureSupport.None;
        if (supportsGPUTextureCopy)
        {
            Graphics.CopyTexture(rt, nTex);
        }
        else
        {
            nTex.ReadPixels(new Rect(0, 0, newWidth, newHeight), 0, 0);
            nTex.Apply();
        }
```

and in the resize function just use:

```
    Graphics.CopyTexture(rt, nTex);
```

?

kuruk-mm (approved):
Great! I let you a little comment

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/TextureHelpers/TextureHelpers.cs
pravusjif: Yeah that's why I updated the comment on that line, I just talked to Brian about this, I'll apply that change today before merging.

pravusjif: done!

PR #1985, opened 2022-03-04
chore: toggle world blocker colliders on renderer state change

pravusjif:
**WHY**
When we are initializing the world, sometimes it happens that we end up loading (and prioritizing) the parcel next to the one we specified in the URL instead of the desired one because the character collides with the world blockers when the rendering is OFF. The problem is specially annoying for debugging entering to a specific scene with LOS=0 for example.

**WHAT**
Added world blocker colliders enabled state initialization and toggling based on renderer state

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BlockerController/Interfaces/IBlockerInstanceHandler.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BlockerController/Interfaces/IBlockerInstanceHandler.cs
@@ -8,7 +8,8 @@ namespace DCL.Controllers
         void DestroyAllBlockers();
         Dictionary<Vector2Int, IPoolableObject> GetBlockers();
         void HideBlocker(Vector2Int coords, bool instant = false);
-        void ShowBlocker(Vector2Int pos, bool instant = false);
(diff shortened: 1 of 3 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/BlockerController/Interfaces/IBlockerInstanceHandler.cs
```diff
@@ -8,7 +8,8 @@ public interface IBlockerInstanceHandler
         void DestroyAllBlockers();
         Dictionary<Vector2Int, IPoolableObject> GetBlockers();
         void HideBlocker(Vector2Int coords, bool instant = false);
-        void ShowBlocker(Vector2Int pos, bool instant = false);
+        void ShowBlocker(Vector2Int pos, bool instant = false, bool colliderEnabled = true);
+        void ToggleBlockersCollision(bool newState);
```
AjimenezDCL, line 12:
```suggestion
        void SetCollision(bool newState);
```

We are not actually toggling, just setting the new collision state.

AjimenezDCL (approved):
Approving with a small coment

lorux0 (approved):

PR #1987, opened 2022-03-07
fix: set camera rotation from kernel message

pbosio:
fixes https://github.com/decentraland/sdk/issues/121
camera was not rotating correctly when receiving rotation message from kernel
*modified `OnSetRotation` math from first and third camera state controller

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Camera/CameraStateFPS.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Camera/CameraStateFPS.cs
@@ -38,10 +38,13 @@ namespace DCL.Camera
 
             if (payload.cameraTarget.HasValue)
             {
-                var newPos = new Vector3(payload.x, payload.y, payload.z);
                 var cameraTarget = payload.cameraTarget.GetValueOrDefault();
-                var dirToLook = (cameraTarget - newPos);
-                eulerDir = Quaternion.LookRotation(dirToLook).eulerAngles;
+
+                var horizontalAxisLookAt = payload.y - cameraTarget.y;
+                var verticalAxisLookAt = new Vector3(cameraTarget.x - payload.x, 0, cameraTarget.z - payload.z);
+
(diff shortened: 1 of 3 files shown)


pravusjif (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Camera/CameraStateFPS.cs
```diff
@@ -38,10 +38,13 @@ public override void OnSetRotation(CameraController.SetRotationPayload payload)
...
-                var newPos = new Vector3(payload.x, payload.y, payload.z);
                 var cameraTarget = payload.cameraTarget.GetValueOrDefault();
-                var dirToLook = (cameraTarget - newPos);
-                eulerDir = Quaternion.LookRotation(dirToLook).eulerAngles;
+
+                var horizontalAxisLookAt = payload.y - cameraTarget.y;
+                var verticalAxisLookAt = new Vector3(cameraTarget.x - payload.x, 0, cameraTarget.z - payload.z);
+
+                eulerDir.y = Vector3.SignedAngle(Vector3.forward, verticalAxisLookAt, Vector3.up);
+                eulerDir.x = Mathf.Atan2(horizontalAxisLookAt, verticalAxisLookAt.magnitude) * Mathf.Rad2Deg;
```
sandrade-dcl, line 47: Suggestion: I have seen that this fragment of code is exactly the same in `CameraStateFPS` and `CameraStateTPS`... maybe we could extract it to a function into some common "util" class?

sandrade-dcl (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Camera/CameraStateFPS.cs
pbosio: I would like to keep it as it is because I think that the rotation for the third person camera should be modified, since it will point the center of the camera to the target position and most of the time the avatar is blocking the visual to that position. So I will like to open the discussion to modify that behavior later :)

PR #1990, opened 2022-03-09
fix: non global-scene avatars

pbosio:
for the moment we do not officially support using `AvatarShape` through the sdk, but devs can still use it.
so here are some fixes to mitigate the possibility of the explorer malfunctioning when an `AvatarShape` is created by a scene.
changes:
*avoid adding scene avatars to `DataStore` variable `otherPlayers`
*fix issues generated by avatar model's field `id` not being set
*use local scene position instead of world position to translate avatars created by a scene (we might want, in the near future, to start supporting devs using `AvatarShape` in their scenes)
*removed some duplicated code regarding avatar movement (the calculation was already done by `AvatarMovementController`)
*relocated subscription to entity's transform changes, since the subscription was done after the avatar is loaded, we were sometimes "losing" that event
*deleted `lastAvatarPosition` field since it was not being used
fixes https://github.com/decentraland/sdk/issues/145


pravusjif (approved):
Nice! this fixes lots of long time AvatarShape problems

AjimenezDCL (approved):
Good job!

PR #1995, opened 2022-03-11
fix: missing nullref check in nftshape info success event

pravusjif:
added missing nullref check, saw this nullref appearing when moving out of the museum district scene.

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShapeLoaderController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShapeLoaderController.cs
@@ -150,6 +150,10 @@ public class NFTShapeLoaderController : MonoBehaviour
 
     private void FetchNFTInfoSuccess(NFTInfo nftInfo)
     {
+        // the FetchNFTInfoSuccess action can get called at the NFTInfoLoadHelper after destroying this controller
+        if (this == null)
+            return;
+        
         loadNftAssetCoroutine = StartCoroutine(LoadNFTAssetCoroutine(nftInfo));
     }
 
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShapeLoaderController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShapeLoaderController.cs
@@ -1,14 +1,10 @@
 using DCL.Components;
 using DCL.Helpers.NFT;
 using System.Collections;
-using System.Collections.Generic;
-using System.IO;
 using System.Text.RegularExpressions;
 using UnityEngine;
 using DCL;
-using Newtonsoft.Json;
 using NFTShape_Internal;
-using UnityGLTF.Loader;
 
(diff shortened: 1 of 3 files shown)


sandrade-dcl (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShapeLoaderController.cs
```diff
@@ -150,6 +150,10 @@ private void FetchNFTContents()
 
     private void FetchNFTInfoSuccess(NFTInfo nftInfo)
     {
+        // the FetchNFTInfoSuccess action can get called at the NFTInfoLoadHelper after destroying this controller
```
BrianAmadori, line 153: Not happy with this fix, but approving for stability. The best way to fix this shouldn't be to ensure the subscriptor is unsubscribed from this event when is destroyed?

BrianAmadori (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShapeLoaderController.cs
pravusjif: I thought about that as well but the thing with this `FetchNFTInfoSuccess()` "Action" is that it's not used through event subscription, it's being injected into another class method as a callback to be invoked when successfull: https://github.com/decentraland/unity-renderer/blob/074d7043c2eacdc262eddf9252236d374620031e/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShapeLoaderController.cs#L148

And even when this `NFTShapeLoaderController` becomes null when being destroyed, that callback injected of course can still be invoked afterwards

pravusjif: Let me see if I can make a little refactor to solve that, since I think each `NFTShapeLoaderController` uses its own `NFTInfoLoadHelper` we should be able to add an event in the helper and unsubscribe the controller when destroyed

pravusjif: Done!

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/NFTShape/NFTShapeLoaderController.cs
```diff
@@ -145,11 +148,11 @@ public void UpdateBackgroundColor(Color newColor)
     private void FetchNFTContents()
     {
         ShowLoading(true);
-        nftInfoLoadHelper.FetchNFTInfo(darURLRegistry, darURLAsset, FetchNFTInfoSuccess, FetchNFTInfoFail);
+        nftInfoLoadHelper.FetchNFTInfo(darURLRegistry, darURLAsset);
```
BrianAmadori, line 151: I was thinking what if we do many requests one after another? But I think it would work fine, they would just use the same callback at the end anyways and be affected by the unsubscription.

BrianAmadori (approved):
<3

PR #2005, opened 2022-03-16
feat: remove tags from the chat's bodies

AjimenezDCL:
Users are taking advantage of the rich text to do stuff like increasing the size of their messages, or color them.
We cannot disable rich text due to its usage to bold stuff like the names, but we can take advantage of the `<noparse>` feature to avoid rich text in the body of the message. 

## Test
Test with `kernel-version=1.0.0-1970282103.commit-16f79be` , before the soft filtering of `< >` were set.
https://play.decentraland.zone/?NETWORK=mainnet&island=I47wbd&kernel-version=1.0.0-1970282103.commit-16f79be&position=100%2C100&realm=loki


davidejensen (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatUtils.cs
```diff
@@ -0,0 +1,10 @@
+public static class ChatUtils
+{
+    public static string AddNoParse(string message)
+    {
+        var filteredMessage = message.Replace("noparse", "");
```
Suduck, line 5: this won't replace the text if the user is saying "noparse" in the chat?

Suduck (approved):
Approved but one concern arises!

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatUtils.cs
davidejensen: @AjimenezDCL should we replace "<noparse>" and "</noparse>" instead?

AjimenezDCL: sure, the chances of somebody using "noparse" in a real sentences are quite slim, but I will change it anyway

AjimenezDCL: I'm going to replace:

`<noparse`  and `</noparse`.

Because othrwise we should replade: `<noparse>`, `<noparse >`, `</noparse>` and `</noparse >`

PR #2027, opened 2022-03-21
fix: memory leak caused by cancellations

Kinerius:
## What does this PR change?

Fixes a Memory Leak caused by textures being created asynchronously and the process being canceled before those textures were handled correctly for their deletion.
This was introduced with #1761 

## How to test the changes?

Unity Only:
- Open up APK_GLTF_InteractiveTest scene and run it
- Open up the Profiler and check out current Texture and Mesh memory and **save those numbers**.
- Press Z to spawn GLTFs and X to delete the last one loaded, you can press C for an automated and randomized Loading and Deletion and press again to toggle it Off.
- Press X until no GLTFs are loaded
- Check the Texture and Memory usage and compare it with the initial values, they must be equal or somewhat close (< 1 MB of tolerance )

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/UnityGLTF/Scripts/GLTFComponent.cs
+++ b/unity-renderer/Assets/UnityGLTF/Scripts/GLTFComponent.cs
@@ -100,16 +100,12 @@ namespace UnityGLTF
         private Settings settings;
 
         private  CancellationTokenSource ctokenSource;
-        private bool isDestroyed;
 
         public Action OnSuccess { get { return OnFinishedLoadingAsset; } set { OnFinishedLoadingAsset = value; } }
 
         public Action<Exception> OnFail { get { return OnFailedLoadingAsset; } set { OnFailedLoadingAsset = value; } }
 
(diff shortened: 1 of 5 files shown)


unity-renderer/Assets/UnityGLTF/Scripts/GLTFComponent.cs
```diff
@@ -375,23 +367,16 @@ private void EnqueueDownload()
...
-        }
+        public long GetMeshesMemorySize() { return meshesEstimatedSize; }
 
         private void OnDestroy()
         {
 #if UNITY_STANDALONE || UNITY_EDITOR
             if (DataStore.i.common.isApplicationQuitting.Get())
                 return;
 #endif
-            isDestroyed = true;
```
BrianAmadori, line 394: What about the handling of the `isDestroyed` flag? We shouldn't be removing it altogether? As it is right now we are creating dead code with this change.

BrianAmadori (approved):

PR #2029, opened 2022-03-22
feat: biw rotation and cache 

### LINKED ISSUE #2023: BIW: Thumbnail is not updated in some specific projects
In some specifics projects, the thumbnail is not updated when the project is updated

Suduck:
## What does this PR change?

Fixes #2023 

This PR fixes a problem with the cache of the images that we have in projects and some refinements from the dry run of the feature

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=feat/biw-projects
2. Test that the project images are downloaded correclty
3. Test that the rotation of the project doesn't affect the scene

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/DCLPlugins/BuilderInWorld/HUD/ProjectsPanelHUD/Scripts/Views/Projects/ProjectCardView.cs
+++ b/unity-renderer/Assets/DCLPlugins/BuilderInWorld/HUD/ProjectsPanelHUD/Scripts/Views/Projects/ProjectCardView.cs
@@ -1,13 +1,11 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Configuration;
-using DCL;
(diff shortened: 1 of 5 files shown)


sandrade-dcl (approved):
Only a suggestion: Why don't you use the `Image` UI Component that is already implementing all the logic related to the promises for loading the image? I think you could simplify your code a lot.

Suduck:
> Only a suggestion: Why don't you use the `Image` UI Component that is already implementing all the logic related to the promises for loading the image? I think you could simplify your code a lot.

The problem is that the promises don't send the "no-cache" header to the request, we need to specify it for the request for this thumbnail otherwise sometimes it will be cached by chrome. The promises are not an option since there is no way to specify it

Also, take into account that this is "legacy" code that was here before the UI library, so this contains the old format

PR #2032, opened 2022-03-22
revert: GLTF Multithreading 

Kinerius:
## What does this PR change?

Reverts #1761 while keeping most newest improvements

## Our Code Review Standards


unity-renderer/Assets/UnityGLTF/Scripts/Loader/WebRequestLoader.cs
```diff
@@ -62,43 +63,42 @@ public string GetWrappedUri(string uri)
...
 #endif
             
-            token.ThrowIfCancellationRequested();
-            
             bool error = false;
             string errorMessage = null;
 
+            if (asyncOp == null)
+            {
+            }
```
BrianAmadori, line 100: we should remove this line
```suggestion
```

BrianAmadori:
Note for reviewers: we talked and agreed to left the memory estimation code out until we add back the multithreading feature. That's the reason some tests were marked as [Explicit].

unity-renderer/Assets/UnityGLTF/Scripts/Loader/WebRequestLoader.cs
Kinerius: Yes, the deletion got reverted 😆

pravusjif (approved):

PR #2034, opened 2022-03-22
feat: compress map texture chunks to reduce memory usage

Kinerius:
## What does this PR change?

Reduces the memory usage of the Map Textures by ~75%

Before ( 7x7x10.6 = 519.4 MB ) 
image

After: (7x7x2.7 = 132.3 MB )
image

**IMPORTANT NOTE:**  The side effect of doing this is that there's a new hiccup when opening the map. 

## Our Code Review Standards


Suduck (approved):

davidejensen (changes requested):
Works fine when the zoom is at a normal level or zoomed in:
Screenshot 2022-03-23 at 09 07 12
Breaks when going in the last 2 levels of zoom:
Screenshot 2022-03-23 at 09 07 25

davidejensen:
@Kinerius I think I fixed the issue I encountered by moving the compress after the .Apply, the results in terms of memory is the same of the improved one (2.7MB per chunk) but now the visual bug is no longer there. I've committed the changes to test the build version as well, let's discuss today in the daily about this

davidejensen (approved):

PR #2043, opened 2022-03-23
fix: GLTF mesh creation throttling regression

### LINKED ISSUE #1980: GLTF throttling budget is not being shared across diferent importing instances when it should
When GLTF imported throttling was moved to Tasks, the budget counter stopped being static. This means that is no longer shared between the many GLTFs that are imported along the client at the same time. 

This should be fixed to decrease hiccups on browsers and on desktop mesh construction logic.

Kinerius:
## What does this PR change?

Fixes #1980 throttling regression.

Bonus:
Removed all local throttles as they do little to no difference on performance rather than fix very specific use cases. 
Also added an option to force the coroutines to be run synchronously to fix the Asset Bundle converter. 

## How to test the changes?

Run through the world with Asset Bundles disabled, meshes should be correctly throttled on their creation and should be fewer hiccups.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/ABConverter/Core.cs
+++ b/unity-renderer/Assets/ABConverter/Core.cs
@@ -713,6 +713,7 @@ namespace DCL.ABConverter
             loader.maximumLod = 300; // todo: fix this later
             loader.forceGPUOnlyMesh = false;
             loader.forceGPUOnlyTex = false;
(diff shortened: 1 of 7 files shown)


lorux0 (approved):
looks good to me!

PR #2045, opened 2022-03-23
fix: components not being removed on `ComponentRemoved` message

pbosio:
Pretty much every `IEntityComponent` was not being removed when receiving `ComponentRemoved` message from kernel (except from shapes and pointer events)

*added code to try to remove component by name when message is received
*modified `BaseComponent` class to ensure that every inheritor must set `componentName` property
*added class name to components that didn't have it

the following ecs script as an example, show billboard component is never removed
```
let active = true
const e1 = new Entity();
e1.addComponent(new Transform({ position: new Vector3(16 * 0.5, 1, 9), scale: new Vector3(16, 1, 14) }))
e1.addComponent(new BoxShape()).withCollisions = false
e1.addComponent(new Billboard())

Input.instance.subscribe("BUTTON_DOWN", ActionButton.PRIMARY, false, ev => {
    if (active) {
        e1.removeComponent(Billboard)
        log(`remove billboard`)
    }
    else {
        e1.addComponent(new Billboard())
        log(`create billboard`)
    }
    active = !active
})

engine.addEntity(e1)
```


BrianAmadori (changes requested):
The implementation looks good, however we shouldn't be covering this sensible feature with tests?

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Animator/DCLAnimator.cs
```diff
@@ -43,6 +43,8 @@ public class DCLAnimationState
         Dictionary<string, AnimationClip> clipNameToClip = new Dictionary<string, AnimationClip>();
         Dictionary<AnimationClip, AnimationState> clipToState = new Dictionary<AnimationClip, AnimationState>();
 
+        public override string componentName => "animator";
```
Suduck, line 46: We should always uses camelCase or Snake, so this  should be Animator or the rest should be audioSource

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/AvatarAttach/AvatarAttachComponent.cs
```diff
@@ -56,7 +56,7 @@ IEnumerator IComponent.ApplyChanges(BaseModel newModel)
 
         void IComponent.RaiseOnAppliedChanges() { }
 
-        Transform IMonoBehaviour.GetTransform() => handler.entity?.gameObject.transform;
+        Transform IMonoBehaviour.GetTransform() => null;
```
Suduck, line 59: Is this change expected?

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioSource.cs
```diff
@@ -31,6 +31,8 @@ public class Model : BaseModel
         public long playedAtTimestamp = 0;
         private bool isOutOfBoundaries = false;
 
+        public override string componentName => "AudioSource";
```
Suduck, line 34: This too

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Audio/DCLAudioStream.cs
```diff
@@ -28,6 +28,8 @@ public class Model : BaseModel
         private bool isDestroyed = false;
         private Model prevModel = new Model();
 
+        public override string componentName => "AudioStream";
```
Suduck, line 31: This too

Suduck (approved):
Good job! Small change requsted. Also, Do you consider that we should move all the names to constants ? Maybe in the CLASS_ID file? I let that decision to you!

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/AvatarAttach/AvatarAttachComponent.cs
pbosio: yes, `IMonoBehaviour.GetTransform` is only used for "pointer events" components so is not necessary to set something there

pbosio:
> Good job! Small change requsted. Also, Do you consider that we should move all the names to constants ? Maybe in the CLASS_ID file? I let that decision to you!

I don't think we will gain something by moving those name to constants in other file since there is no hardcoded string in our project making any reference to them and I don't think it would mitigate the possibility of any unintended error. So I would prefer to keep them there explicitly in every component class

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/AvatarAttach/AvatarAttachComponent.cs
pbosio: Also, this component does not create any `GameObject` so it makes no sense setting something there :P

pravusjif:
> > Good job! Small change requsted. Also, Do you consider that we should move all the names to constants ? Maybe in the CLASS_ID file? I let that decision to you!
> 
> I don't think we will gain something by moving those name to constants in other file since there is no hardcoded string in our project making any reference to them and I don't think it would mitigate the possibility of any unintended error. So I would prefer to keep them there explicitly in every component class

IT's true that many other component classes not modified in this PR already have the componentName override like `public override string componentName => "NFT Shape";`, In the future we can change that on all components to avoid memory fragmentation by instantiating a new string every time the `componentName` is used, right now it may not be needed

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ParcelScene.cs
```diff
@@ -684,6 +684,32 @@ private void RemoveEntityComponent(IDCLEntity entity, string componentName)
                         }
                     }
                     return;
+                default:
```
pravusjif, line 687: With this new `default` logic, do the `OnClick.NAME`, `OnPointerDown.NAME`, `OnPointerUp.NAME`, `OnPointerHoverEnter.NAME`, `OnPointerHoverExit.NAME` cases of that `switch` still make sense? couldn't those work fine with the new `default` as well?

pbosio: yes and no :P
having those as switch cases will make it faster than the default case which iterate through the component list to find the component to remove.
Ideally, for better performance, we should add a case in that switch for every component and the default case is mostly to handle the components that we forgot to add there

pbosio:
> > > Good job! Small change requsted. Also, Do you consider that we should move all the names to constants ? Maybe in the CLASS_ID file? I let that decision to you!
> > 
> > 
> > I don't think we will gain something by moving those name to constants in other file since there is no hardcoded string in our project making any reference to them and I don't think it would mitigate the possibility of any unintended error. So I would prefer to keep them there explicitly in every component class
> 
> It's true that many other component classes not modified in this PR already have the componentName override like `public override string componentName => "NFT Shape";`, In the future we can move those to CONST on all components to avoid memory fragmentation by instantiating a new string every time the `componentName` is used, right now it may not be needed

hopefully, in the future, we won't need that component name at all :)

pravusjif (changes requested):
The code looks good but after a little but when Genesis Plaza loads, we get this error in the console and the automatic doors from the ground floor don't work, they never open and you can get trapped in the main building:
Screen Shot 2022-03-29 at 15 05 30

pbosio:
> The code looks good but after a little but when Genesis Plaza loads, we get this error in the console and the automatic doors from the ground floor don't work, they never open and you can get trapped in the main building: Screen Shot 2022-03-29 at 15 05 30

fixed. seems like I broke something with quests when I did some changes to add tests

pravusjif (approved):

Issue #1456, opened 2021-11-03
GLTF vs AB debugging view


olavra:
As a developer, I want to be able to see the assets that has been loaded as Asset bundles or GLTFs by enabling a tool in the explorer.
Ideally this could be a console command or/and URL param that replaces diffuse color of the assets by red (GLTF) or green (AB).

PR #2058, opened 2022-03-24
debug: ab gltf debug viewer

Suduck:
## Context 

We have AB ( asset bundles) to improve the performance of the scenes. These assets bundles are converted after the scene has been published and this conversion can fail completely or by parts, when this happens, we use the original GLTFs instead of the asset bundles.

So you can have a scene with half of the items as asset bundles and the other half as GLTF.

With this PR, you can debug the scene and see which items are GLTF and which ones are AB.

This PR was rush during the Fashion week to ensure that the scenes were converted correctly, so not a correct approach was done in order to have it as quickly as possible

## How to see that

When you use the shortcut, it will change the materials of all GLTFComponents to red or green based on their origin

Red material: GLTF 
Green material: Asset Bundles

`Left shift + A ` : Changes the materials of the current scene
`Left shift + Q ` : Changes the materials of all the loaded scenes 
`Left shift + F` : Revert all the changes done to the materials

## What does this PR change?

This PR should NEVER be merged since it is a very hacky way to see the AB. 

It adds the ability to check the conversion of the scene to asset bundles


AjimenezDCL:
Waiting until this is done: https://github.com/decentraland/unity-renderer/issues/1456

PR #2062, opened 2022-03-24
fix: AvatarModifierArea losing modifiers after model update

pbosio:
Avatar inside `AvatarModifierArea` were losing it modifiers every time the component was updated.
For example, in the following code, the avatar inside the area becomes visible every time the area size changes and the avatar must move out and in the area to become invisible again.
```

let active = true
const e1 = new Entity();
const m = e1.addComponent(new AvatarModifierArea({ area: { box: new Vector3(16, 1, 14) }, modifiers: [AvatarModifiers.HIDE_AVATARS] }))
e1.addComponent(new Transform({ position: new Vector3(16 * 0.5, 1, 9), scale: new Vector3(16, 1, 14) }))
e1.addComponent(new BoxShape()).withCollisions = false

Input.instance.subscribe("BUTTON_DOWN", ActionButton.PRIMARY, false, ev => {
    if (active) {
        m.area = { ...m.area, box: new Vector3(16, 2, 14) }
        log(`change avatar area`)
    }
    else {
        m.area = { ...m.area, box: new Vector3(16, 1, 14) }
        log(`change avatar area again`)
    }
    active = !active
})

engine.addEntity(e1)
```

root cause of the issue was that, even though `RemoveAllModifiers()` was called on every component update, the list containing the avatars inside the area was not being cleaned. so when the code checks again to add the modifiers it would skip applying the modifiers because it assumes that the avatar has the modifier already applied

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/AvatarModifierArea.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarModifiers/AvatarModifierArea.cs
@@ -138,6 +138,7 @@ public class AvatarModifierArea : BaseComponent
     private void RemoveAllModifiers()
     {
         RemoveAllModifiers(DetectAllAvatarsInArea());
+        avatarsInArea = null;
     }
 
     private void RemoveAllModifiers(HashSet<GameObject> avatars)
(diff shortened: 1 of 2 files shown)


Suduck (approved):

BrianAmadori (approved):
LGTM codewise

PR #2073, opened 2022-03-25
fix: apply texture resizing to UIImage

AjimenezDCL:
Apply the resizing factor to UIImages


pravusjif:
https://play.decentraland.org/?position=-100%2C127&realm=dg&renderer-branch=chore/uiimage-resize-factor

AjimenezDCL:
Closing because it will be solved here #2388

PR #2081, opened 2022-03-28
fix: set websocket ssl protocol to tls12 over the default (tls10)

### LINKED ISSUE #2047: We can no longer access .org with unity editor
It's important we are able to, since some scenes like the DG scenes only show some meshes if they are accessed through .org

kuruk-mm:
Fix #2047 

Looks like TLS 1.0 is being deprecated, so when we tried to connect to Unity we got a `unsupported protocol` from the Client.

By default, `websocket-sharp` has TLS 1.0. So here we're enabling it to use TLS 1.2.

## Our Code Review Standards


Kinerius (approved):
It works!

We still require some manual checks like enabling SSL and setting up `allow-insecure-localhost` flag on the browser but it's understandable.

PR #2083, opened 2022-03-28
refactor: plugin system register with the type as key instead of the …

### LINKED ISSUE #2056: PluginSystem: Allow to override the plugin from Desktop
Currently in explorer-desktop we cannot re-use the plugin system from unity-renderer because the key of the registered plugins is the lambda instance.

So from Desktop, we need to re-implement the changes from unity-renderer each time the PluginSystemFactory is changed.

To fix it, we could add as ID the class template as we do in `ServiceLocatorFactory`.

kuruk-mm:
…pluginbuilder

Fix #2056

# Why?

To be able to override the behavior from Desktop easily

## Our Code Review Standards


BrianAmadori (changes requested):

BrianAmadori (approved):
LGTM

PR #2093, opened 2022-03-29
test: no legacy animations

AjimenezDCL:
Testing the usage of non-legacy animations
## Our Code Review Standards


BrianAmadori (commented):
Maybe we should close this PR?

PR #2117, opened 2022-04-04
refactor: nft counting tests

BrianAmadori:
## What does this PR change?

NFTShape tests were failing because there are quite a few systems that couldn't be mocked correctly. Those systems are on the NFTShape domain and doesn't make sense to mock them through the global `ServiceProvider`. So it was decided that this was a perfect opportunity to start moving the shape to a plugin, where a proper DI approach is implemented to be mocked by tests later.

* Add proper mocking for NFTShape through DI
* NFTShape supports DI
* Move NFTShape to its own plugin (partially, still needs to move scripts to plugin's folder)
* Removed `Explicit` for most of NFT tests
* Add external Register/Unregister support to component factory, so more components can be added as separate plugins
* Add test for metrics counting with NFTs
* Removed `NFTShapeConfig` class and ScriptableObject asset. Now configuration is given by to the proper `DataStore_NFTShape` to abide to DataStore pattern.

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=refactor/nft-counting-tests
2. ...

## Our Code Review Standards


Kinerius (approved):
LGTM

PR #2128, opened 2022-04-05
chore: improve character ground check extra distance

pravusjif:
Tried other solutions to fix the problem we get on **the desktop client** character controller getting "floaty" for a second when reaching the ground but since those other solutions (like using `FixedUpdate()`) require to re-balance gravity, etc. I went for the simpler solution that doesn't affect the web client and solves the issue at the desktop client: decreasing a bit the ground check extra distance value.

In case you want to test that the character controller jump height and slope walking keeps working as expected you can enter this scene at ropsten network:
https://play.decentraland.zone/?NETWORK=ropsten&position=59%2C-67&realm=thor&renderer-branch=chore%2FImproveCharacterGroundCheckExtraDistance
chrome_z8HENOj61g

And test that the last slope (47.5º) cannot be "walked on" (all the previous can):
chrome_V0VFWSnZ92

And the 190cm and 200cm columns shouldn't be reachable by jumping from the ground (the other 2 should):
chrome_2kz6hQBzmv

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
@@ -11,7 +11,7 @@ public class DCLCharacterController : MonoBehaviour
     [Header("Movement")]
     public float minimumYPosition = 1f;
 
-    public float groundCheckExtraDistance = 0.25f;
+    public float groundCheckExtraDistance = 0.1f;
     public float gravity = -55f;
     public float jumpForce = 12f;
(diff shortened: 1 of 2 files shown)


Kinerius (commented):
I think this is worsening a bit the "Flying animation" issue by a small margin, maybe we should address that issue first.

pravusjif:
> I think this is worsening a bit the "Flying animation" issue by a small margin, maybe we should address that issue first.

Oh you mean the current user avatar in third person, right? (since other avatars grounding checks are not affected by the modified value) then just in case let's wait for this other PR to be merged first then: https://github.com/decentraland/unity-renderer/pull/2131

Kinerius (approved):
I think we can merge this without waiting for the other PR

PR #2131, opened 2022-04-05
fix: free fall animation when grounded

Kinerius:
## What does this PR change?

Fixes the freefall animation when grounded, only on current player.

## Our Code Review Standards


pravusjif (commented):
Nice catch, I tested that this PR effectively improves the current user avatar grounded state for its avatar shape when in 3rd person (screenshots below).

I also tested locally removing the changes made to the `DCLCharacterController` and the improvement **is still working** so I'd remove those changes and only keep the changes made to the `AvatarAnimatorLegacy`, since adding the `characterController.Move()` `ColliderFlags` return value usage doesn't add anything besides an extra check (although super cheap). In another PR we can test the theory of replacing the raycasts sequence we currently have for ground detection and use those collider flags + 1 extra check to get the moving platforms info. 

Screen Shot 2022-04-05 at 15 54 09

Screen Shot 2022-04-05 at 15 55 23

Screen Shot 2022-04-05 at 15 56 00

Screen Shot 2022-04-05 at 15 57 01

Screen Shot 2022-04-05 at 15 57 16

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarAnimatorLegacy.cs
```diff
@@ -65,6 +65,9 @@ public class BlackBoard
     Vector3 lastPosition;
     bool isOwnPlayer = false;
     private AvatarAnimationEventHandler animEventHandler;
+    
+    private float lastOnAirTime = 0;
+    private const float FORCE_GROUND_TIME = 0.25f;
```
AjimenezDCL, line 70: Nitpicking here, let's move it to the top of the class.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
```diff
@@ -472,14 +473,18 @@ public Transform CastGroundCheckingRays()
         return null;
     }
 
-    public bool CastGroundCheckingRays(float extraDistance, float scale, out RaycastHit hitInfo) { return CastGroundCheckingRays(transform, collider, extraDistance, scale, groundLayers, out hitInfo); }
+    public bool CastGroundCheckingRays(float extraDistance, float scale, out RaycastHit hitInfo)
+    {
+        bool raycastSucceded = CastGroundCheckingRays(transform, collider, extraDistance, scale, groundLayers, out hitInfo);
+        return IsLastCollisionGround() || raycastSucceded;
```
AjimenezDCL, line 479:
```suggestion
        if(CastGroundCheckingRays(transform, collider, extraDistance, scale, groundLayers, out hitInfo))
                return true;
        return IsLastCollisionGround();
```

More nitpicking, Isn't this equivalent and more readable?

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarAnimatorLegacy.cs
```diff
@@ -113,6 +116,12 @@ void UpdateInterface()
 
         //NOTE(Brian): Vertical speed
         float verticalVelocity = flattenedVelocity.y;
+
+        if (Mathf.Abs(verticalVelocity) > float.Epsilon)
```
AjimenezDCL, line 120: What happens if I climb some stairs? wouldnt my vertical velocity be positive and (in some cases) higher than `float.Epsilon`?

AjimenezDCL (approved):

pravusjif (changes requested):
I re-tested with 2 sessions in ropsten 64,-64 and now the grounding works better for other avatars as well, nice job!

Although I still believe we shouldn't be introducing the changes made in the `DCLCharacterController` as they don't add anything besides an extra check (only used for the current avatar), since the raycasts are still being cast (and we need them to get the moving platforms transform), as I mentioned in my previous comment

Kinerius:
> I re-tested with 2 sessions in ropsten 64,-64 and now the grounding works better for other avatars as well, nice job!
> 
> Although I still believe we shouldn't be introducing the changes made in the `DCLCharacterController` as they don't add anything besides an extra check (only used for the current avatar), since the raycasts are still being cast (and we need them to get the moving platforms transform), as I mentioned in my previous comment

The extra check fixes cases where the raycasts does not detect the ground correctly, here's a screenshot of Not having that extra check

image

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarAnimatorLegacy.cs
Kinerius: This is an additional check based on velocity, the raycast is still there, so we are not introducing another case.
Anyways I've tested it by removing all the "isOwnPlayer" conditionals and the only issue I see is when we go down from stairs, the character moves too fast and we have some frames where we are not grounded and the free fall animation triggers, this won't be too noticeable because the interpolation tends to make everyone fly anyways.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarAnimatorLegacy.cs
Kinerius: Nice catch, fixed.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarAnimatorLegacy.cs
Kinerius: I've added some comments to clarify its purpose

pravusjif:
> > I re-tested with 2 sessions in ropsten 64,-64 and now the grounding works better for other avatars as well, nice job!
> > Although I still believe we shouldn't be introducing the changes made in the `DCLCharacterController` as they don't add anything besides an extra check (only used for the current avatar), since the raycasts are still being cast (and we need them to get the moving platforms transform), as I mentioned in my previous comment
> 
> The extra check fixes cases where the raycasts does not detect the ground correctly, here's a screenshot of Not having that extra check
> 
> image

got it, nice catch!

pravusjif (approved):

PR #2147, opened 2022-04-08
refactor: change entityId to long instead of string

Suduck:
## What does this PR change?

This PR is an advance for ECS 7.0, it will convert all the entityIds from strings to longs

These changes can be implemented without any problem in production. 

The idea is to change forever the way that we handle the entity IDs, to use long instead of string because it is more performance and it is a requirement for the ECS 7.0.

This change will be retro compatible with all the ECS versions, and no changed is needed in the kernel part.

We are using the HashCode of the strings to represent the id of the entities.

Since the message system is handling a TAG of the entities that are tied to the string identifier from kernel, we need to do this implementation at the scene level 

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=refactor/ultimate-entity-in64
2. Test that everything is working as expected

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/WebInterface/Interface.cs
```diff
@@ -1451,15 +1451,15 @@ public static void NotifyStatusThroughChat(string message)
             SendMessage("VideoProgressEvent", progressEvent);
         }
 
-        public static void ReportAvatarRemoved(string entityId, string avatarId)
+        public static void ReportAvatarRemoved(long entityId, string avatarId)
         {
             avatarStatePayload.type = "Removed";
             avatarStatePayload.entityId = entityId;
```
menduz, line 1457: should this type change be considered in kernel?

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ParcelScene.cs
```diff
@@ -405,7 +405,7 @@ public void SetEntityParent(string entityId, string parentId)
 
             if ( DCLCharacterController.i != null )
             {
-                if (parentId == "FirstPersonCameraEntityReference" || parentId == "PlayerEntityReference") // PlayerEntityReference is for compatibility purposes
+                if (parentId == "FirstPersonCameraEntityReference".GetHashCode() || parentId == "PlayerEntityReference".GetHashCode()) // PlayerEntityReference is for compatibility purposes
```
menduz, line 408: Can this magic hash codes be constants? I'd like to have them outside of the range of "user entities", maybe a negative number or super high number. I consider having GetHashCode is a dangerous blackbox

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ParcelScene.cs
```diff
@@ -429,7 +429,7 @@ public void SetEntityParent(string entityId, string parentId)
                 }
             }
 
-            if (parentId == "0")
+            if (parentId == "0".GetHashCode())
```
menduz, line 432: We should do special code for parent entity, the ID of the root entity will be 0 in the new ECS as it is in the current one. `"0".GetHashCode() != 0`

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Tests/SceneMetricsCounterShould.cs
```diff
@@ -133,16 +133,16 @@ public void CountTexturesMemoryWhenAddedAndRemoved()
...
 
         Rendereable rendereable1 = new Rendereable();
-        rendereable1.ownerId = "A";
+        rendereable1.ownerId =0;
         rendereable1.textures.Add(tex1);
         rendereable1.textures.Add(tex2);
 
         Rendereable rendereable2 = new Rendereable();
-        rendereable2.ownerId = "B";
+        rendereable2.ownerId =0;
```
menduz, line 141: is this OK that A,B,C are all 0 here?

menduz (changes requested):
See comments, I'd like to have a stable number for well known entities instead of string.GetHashCode(), i.e.: root=0, AvatarPositionEntityReference etc. 

Those stable numbers will be used by the new ECS and will be part of the "Protocol Definition" for Decentraland

unity-renderer/Assets/Scripts/MainScripts/DCL/WebInterface/Interface.cs
Suduck: Yes, this type should change in kernel as well

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ParcelScene.cs
Suduck: Yes, I can move them to constant, however we should not use a negative or super high number rather than GetHashCode, because we can end using another string by accident. GetHashCode is the best way to ensure retrocompatibility right now.

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ParcelScene.cs
Suduck: I don't know if I understood correctly what you mean there. We need to maintain retrocompatibility with the old ECS, and the old ECS will send the "0" string. Do you mean change the way the parent works?

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneController.cs
```diff
@@ -228,14 +228,14 @@ public bool ProcessMessage(QueuedSceneMessage_Scene msgObject, out CustomYieldIn
                     case MessagingTypes.ENTITY_CREATE:
                         {
                             if (msgPayload is Protocol.CreateEntity payload)
-                                scene.CreateEntity(payload.entityId);
+                                scene.CreateEntity(payload.entityId.GetHashCode());
```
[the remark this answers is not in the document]
Suduck: Do you mean to add new messages for special cases?

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ParcelScene.cs
BrianAmadori: @menduz In the "legacy" scene runtime, the 10 first id numbers are reserved for these cases:
https://github.com/decentraland/kernel/blob/main/packages/scene-system/sdk/SceneRuntime.ts#L62

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneController.cs
```diff
@@ -228,14 +228,14 @@ public bool ProcessMessage(QueuedSceneMessage_Scene msgObject, out CustomYieldIn
...
 
                             break;
                         }
                     case MessagingTypes.ENTITY_REPARENT:
                         {
                             if (msgPayload is Protocol.SetEntityParent payload)
-                                scene.SetEntityParent(payload.entityId, payload.parentId);
+                                scene.SetEntityParent(payload.entityId.GetHashCode(), payload.parentId.GetHashCode());
```
BrianAmadori, line 238: We should NOT use hashcode for converting the string ids to longs. Note that Kernel already uses long for ids internally when passing the messages to Unity. The original string ids are actually base-27 expressions. So the conversion should be made converting from base 27 to base 10. We should untangle this conversion from kernel or re-convert here using this criteria.

BrianAmadori (changes requested):

AjimenezDCL (approved):
I let a couple of small requests but I agree what the concerns raised by Mendez and Brian

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneController.cs
menduz: This is not entirely valid, entity ids happen to be auto increments but that is not the case 100% of times. Hash code seems a better fit for the general case

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ParcelScene.cs
menduz: Please let's create a "conversion function" that takes the strings and gives us predictable constant values for well known entities such as "0" and the others mentioned across the PR. That function should be used to convert string->long without risking any protocol due to magic hashcodes

menduz: I'll take the chance to write down a spec document for the SDK entities today to have this thing documented as part of the "Decentraland Protocol"

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneController.cs
menduz: Oh, in a second look it seems like there is a double mapping in the kernel for this

unity-renderer/Assets/Scripts/MainScripts/DCL/WebInterface/Interface.cs
menduz: I propose you to create a different function with the new types so we don't break backwards compatibility

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ParcelScene.cs
BrianAmadori: I already wrote it down on another comment but just in case: the conversion is a base conversion, ids are just constructed by converting the incremental ids to base 36. In the actual implementation 0 to 10 are reserved for special cases. I recognize that only reserving 10 ids can be a limitation so you may want to change it for the SDK 7, just stating how it works right now.

unity-renderer/Assets/Scripts/MainScripts/DCL/WebInterface/Interface.cs
Suduck: After creating another function with new types I just realized that this field has not usage on the kernel side, I have made a PR to remove that field in kernel and removed it from this part as well

unity-renderer/Assets/Scripts/MainScripts/DCL/WebInterface/Interface.cs
```diff
@@ -1451,15 +1451,15 @@ public static void NotifyStatusThroughChat(string message)
...
             avatarStatePayload.avatarShapeId = avatarId;
             SendMessage("ReportAvatarState", avatarStatePayload);
         }
 
-        public static void ReportAvatarSceneChanged(string entityId, string avatarId, string sceneId)
+        public static void ReportAvatarSceneChanged(long entityId, string avatarId, string sceneId)
```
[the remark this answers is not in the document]
Suduck: This field is no longer usage so I has removed it from kernel and Unity

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneController.cs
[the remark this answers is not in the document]
Suduck: but how do we ensure that `0xffffffff` is not a result from a sepecfic string in `entityId.GetHashCode()`?

I mean, imagine that `GetHashCode` of the string `"E8F"` gives the result `4294967295`  0xffffffff representation in long is `4294967295` 

This function will return the same number `4294967295` for the string "0" and for the string "E8F". It is very unlikely to happen but it can happen.


I can add the numbers of the GetHashCode of those strings, but if the GetHashCode implementation of the string class in the project changes, this won't work

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneController.cs
Suduck: Should we then switch the approach GetHashCode for a base36 to base10 conversion? It is efficient enough?

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneController.cs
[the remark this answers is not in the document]
Suduck: Brians says that the first 10 ids are reserved ids so I think we can safely do 

`
        private const long CONST_SCENE_ROOT_ENTITY = 0;
        private const long CONST_AVATAR_ENTITY_REFERENCE = 1;
        private const long CONST_AVATAR_POSITION_REFERENCE = 2;
        private const long CONST_FIRST_PERSON_CAMERA_ENTITY_REFERENCE = 3;
        private const long CONST_THIRD_PERSON_CAMERA_ENTITY_REFERENCE = 4;`

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ParcelScene.cs
Suduck: As you wish, I think that get hash has more performance than convert to base 10, but just let me know which one you prefer and I will change it

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ParcelScene.cs
```diff
@@ -95,6 +101,25 @@ public virtual void SetData(LoadParcelScenesMessage.UnityParcelScene data)
...
+                    return CONST_FIRST_PERSON_CAMERA_ENTITY_REFERENCE;
+                case "AvatarEntityReference":
+                    return CONST_AVATAR_ENTITY_REFERENCE;
+                case "AvatarPositionEntityReference":
+                    return CONST_AVATAR_POSITION_REFERENCE;
+                case "PlayerEntityReference":
+                    return CONST_THIRD_PERSON_CAMERA_ENTITY_REFERENCE;
+            }
+
+            return entityId.GetHashCode() << 9;
```
menduz, line 121: A comment here saying that we are reserving the first 512 entities for special IDs wont hurt. Will also prevent accidentally changing the << 9

menduz (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ParcelScene.cs
```diff
@@ -95,6 +101,25 @@ public virtual void SetData(LoadParcelScenesMessage.UnityParcelScene data)
 
             OnSetData?.Invoke(data);
         }
+        
+        internal long EntityFromLegacyEntityString(string entityId)
```
BrianAmadori, line 105: This is a pure function, we can make it static (and this also saves the need of having to create a `ParcelScene` component to test this).

pbosio (changes requested):
Genesis Plaza doesn't seem to be loading properly using this branch.
this:
Screen Shot 2022-04-26 at 14 48 35

master:
Screen Shot 2022-04-26 at 14 50 16

menduz:
so much space

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneController.cs
```diff
@@ -226,14 +226,14 @@ public bool ProcessMessage(QueuedSceneMessage_Scene msgObject, out CustomYieldIn
...
                             if (msgPayload is Protocol.CreateEntity payload)
-                                scene.CreateEntity(payload.entityId.GetHashCode());
+                                scene.CreateEntity(ParcelScene.EntityFromLegacyEntityString(payload.entityId));
 
                             break;
                         }
                     case MessagingTypes.ENTITY_REPARENT:
                         {
                             if (msgPayload is Protocol.SetEntityParent payload)
-                                scene.SetEntityParent(payload.entityId, payload.parentId);
```
menduz, line 236: Should this have tests?

PR #2151, opened 2022-04-08
test: HUD canvas scalers with constant pixel size

pravusjif:
Did quick update of several HUD canvases we have in our build, changing their Canvas Scaler "UI Scale Mode" from "scale with screen size" to "constant pixel size" only for testing as per requested by @menduz.

The affected canvases were:
* AvatarEditor
* ControlsHUD
* EmotesWheelHUD
* ExternalUrlPromptHUD
* LoadingHUD
* MinimapHUD
* NFTPRomptHUD
* PlayerInfoCardHUD
* SettingsPAnelHUD
* TeleportPromptHUD
* USersAdoundListHUD
* DebugView (network info panel)

Take into account that if we want to go this way in the future, we should create the pertinent issue to change ALL canvases used in the project (that would also involve Romi making an evaluation pass on all the UI because some thing are resized by this change). Also in case we want this change, we should change the canvases we are creating programmatically and check the popular heavy-ui scenes to see how they are affected (DG Casinos, DG Poker, Wonder Mine).

Some examples of little changes I saw on some panels when changing this setting:

Screen Shot 2022-04-08 at 17 33 40

Screen Shot 2022-04-08 at 17 32 55

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/DebugController/Resources/DebugView.prefab
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/DebugController/Resources/DebugView.prefab
@@ -471,7 +471,7 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 0cd44c1031e13a943bb63640046fad76, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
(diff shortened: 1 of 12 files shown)


AjimenezDCL:
@pravusjif can this be closed?

PR #2163, opened 2022-04-13
fix: memory leak

### LINKED ISSUE #2082: AssetBundle Meshes are not being destroyed correctly after usage
**Snapshot at 0,0 ( LOCAL_COMMS )**

image

**Snapshot at 0,0 after teleporting to 100,100 and coming back ( LOCAL_COMMS )**

image

**Comparison**

image

We can see that:
- The first instance is still alive. And has only 1 reference on the second snapshot.   (this is wrong!)
- The pooled object got deleted and created again
- A new instance was created

Here's more info on where is this Mesh referenced:
image

All of the occurrences I've found are references to meshes that were created by AssetBundles

Kinerius:
## What does this PR change?

Fixed memory leaks caused by references in a dictionary that was holding objects with references that should be cleaned up.

This should fix #2082 

## How this change was measured?

I ran some tests with this steps:
- I made use of the `LOCAL_COMMS` URL parameter that prevents the loading of other avatars, this removes "noise" from the profiling.
- The first snapshot is taken at 0,0 (Genesis Plaza) when the boards finish loading.
- The second snapshot is taken after I teleport to 100,100 (mostly barren land) and go back to 0,0 and repeat this process 2 times. So when the snapshot is taken, Genesis plaza was loaded and destroyed 3 times in total.
- Between A and B I've closed unity to prevent the profiler to have garbage

Snapshots A has the fix
Snapshots B doesn't.

image

Although Unity memory profiler has its issues, it still gives us some info to measure.

## How to test the changes?

As always, run through the world, teleport to popular places a bunch of times and it should be less likely for a crash to occur.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadableShape.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/LoadableShapes/LoadableShape.cs
@@ -26,7 +26,7 @@ namespace DCL.Components
 
         protected Model previousModel = new Model();
 
-        protected static Dictionary<GameObject, LoadWrapper> attachedLoaders = new Dictionary<GameObject, LoadWrapper>();
+        private static readonly Dictionary<GameObject, LoadWrapper> attachedLoaders = new Dictionary<GameObject, LoadWrapper>();
 
         public static LoadWrapper GetLoaderForEntity(IDCLEntity entity)
         {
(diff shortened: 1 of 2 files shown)


eordano:
Great work!

pravusjif (approved):
I run the old test of going to Soho Plaza (-80, 0) and coming back to Genesis Plaza several times, compared master branch with this one and in this one I don't get the "infinite memory accumulation" problem, nice job!

Issue #2132, opened 2022-04-05
Fix Scene Boundaries Checker problem that provokes the Black And White bug among others


pravusjif:
The original issue was created in the desktop client: https://github.com/decentraland/explorer-desktop/issues/164 but since this issue is more of the core of the client and also happens in the web client I'm migrating the issue.

Bug effect:
![image](https://user-images.githubusercontent.com/6534068/158173022-00afe138-c864-4e5e-8a4c-a90f40c2972b.png)

![Screen Shot 2022-03-22 at 14 33 36](https://user-images.githubusercontent.com/1031741/159541931-c4c26e7d-9723-4e1f-b662-cf1b097a04dc.png)

![Screen Shot 2022-03-22 at 14 35 25](https://user-images.githubusercontent.com/1031741/159541969-e1a688e9-c54a-4c2b-825f-1ade902cd0bb.png)

https://go.userzoom.com/tracker_shares/353e471f-b2ac-4043-be2e-e35590108245

------------------
Specs of a user that reported the issue:
### Operating System
Windows 10  (10.0.19044) 64bit

### Wallet
Metamask

### CPU
Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz

### GPU
NVIDIA GeForce GTX 1080 with Max-Q Design 8105MB

### RAM
16333MB

### Kernel Version
1.0.0-1970282103.commit-16f79be

### Unity Version
desktop-main.commit-9993f39

### Realm
unicorn

### NameTag
h1p2bl7#b347

------------------

Most reproducible case (I could reproduce this in Windows and Maxc using the web client):

https://user-images.githubusercontent.com/1031741/161770499-caa95c7c-7b4b-4f88-9d77-162f85c0bf1c.mp4

- Stand in the same spot as the video in the 14,-7 scene
- Look in the direction shown in the video
- Set "Ultra" graphics
-  Unload surrounding scenes and re-load them as shown in the video

pravusjif:
After being able to reproduce the issue in Unity Editor I'm slowly getting to the bottom of the issue, it involves several problems:

## Content-side
Genesis Plaza is nesting lots of entities and for a couple of seconds some of them (the ones in the Events and POIs columns at the spawn point) are even having a scale of 1 million (in X, Y and Z) before resizing to their normal size....

[...]

Issue #978, opened 2021-08-12
Attaching meshes to the main character get rendered for some frames when the character hasn't entered the scene


pravusjif:
If a scene attaches something to the main player, if the scene is loaded while the player is outside the scene, the attached meshes are rendered and appear at the main character after turning off because of the scene boundaries checker.

This happens with 100% repro rate if one walks from genesis plaza to this gay pride scene at -50, 1 (make sure the scene is not loaded and gets loaded from stacth when testing)

![Screen Shot 2021-08-12 at 15.57.26.png](https://images.zenhubusercontent.com/5d9200f4f58e25000108abac/7c20bd86-433f-4894-9cbf-b9a40ab1f05d)

pravusjif:
We should re-check if this happens

pravusjif:
[This PR](https://github.com/decentraland/unity-renderer/pull/1732) may have solved the issue, it shoud be re-validated

pravusjif:
This would be solved by: https://github.com/decentraland/unity-renderer/pull/2164

pravusjif:
fixed at https://github.com/decentraland/unity-renderer/issues/2132

pravusjif:
This issue keeps happening, I finally found what's going on and it needs an important refactor to be fixed.

Basically: the [`MaterialTransitionController`](https://github.com/decentraland/unity-renderer/blob/dev/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/MaterialTransitionController/MaterialTransitionController.cs) placeholder renderer is not being toggled based on the entity "inside scene boundaries" state. Also, we are showing the final renderer (not the placeholder one) with a one-time hologram effect when the [`MaterialTransitionController`](https://github.com/decentraland/unity-renderer/blob/dev/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/MaterialTransitionController/MaterialTransitionController.cs) finishes, thus showing the asset regardless of being outside its scene boundaries, for example in the scene mentioned in this issue's original description.

I created this issue to tackle the root of the problem in the future: https://github.com/decentraland/unity-renderer/issues/2886

xavi-dcl:
Closing as we have #2886

Issue #844, opened 2021-07-23
When going around the tram line, loading objects are obstructing the road.


BrianAmadori:
When going at a fast speed back and forth over the new genesis plaza tram line, objects obstructing the road can be seen on the way back. This may be related to a pool recycling issue.

This bug already happens in master but when we apply the memory leak fix proposed in this PR #808, it happens more frequently.

olavra:
The repro rate of this issue is quite low and the time investment to solve it is too high. Going back to the backlog for another time

pbosio:
I will leave little more info about this issue:
* GLTF (`GLTFSceneImporter`) load it meshes progressively and those aren't check against scene bounds until the whole GLTF finishes loading
NOTE: I've tested adding submeshes to `SceneBoundsChecker` as soon as they where created by `GLTFSceneImporter` but performance was terrible and client was running frame by frame in the best scenario
* GLTFShape (`LoadableShape`) set the mesh as visible (depending on model coming from the sdk) as soon as it finish loading and then it's added to `SceneBoundsChecker` which is running a time budget queue so it may take a while before that Shape is processed
NOTE: we can't set the shape as invisible and leave it to the `SceneBoundsChecker` to set it as visible after processed cause Preview, Builder and Builder in World does not hide renderers when out of bounds

pravusjif:
This would be solved by: https://github.com/decentraland/unity-renderer/pull/2164

pravusjif:
fixed at https://github.com/decentraland/unity-renderer/issues/2132

Issue #480, opened 2021-05-21
Entities out of bounds on player load for a second


nearnshaw:
People from the community ran into a weird case that seems to be 100% reproduceable

Go to
 1) https://play.decentraland.org/?position=68%2C75&realm=loki-amber

2) Get in 3rd person view

3) Walk south a couple of parcels (till the xmas version of genesis plaza starts loading)

4) Notice that a christmas bag appears on your player for a second

This bag clearly comes from the plaza right south of this location (the Xmas version of Genesis plaza)
The bag's position was attached to the player, surely the engine is showing it before checking if the player is in the scene

https://images.zenhubusercontent.com/337227404/0304a764-027d-4894-b5b7-6f87b577b963/xmas_bag_480p.mov

BrianAmadori:
The scene limit checks are lazy (this means they are checked little by little over time), I'm sure this is a byproduct of that feature.

olavra:
Christmas scenes are not longer deployed and cant be reproduced. Closing the ticket.

nearnshaw:
The xmas scene mentioned here is still deployed in the indicated coordinates

olavra:
@nearnshaw  This is a weird christmas scene
![image](https://user-images.githubusercontent.com/41490935/128501046-50aced3c-066a-4066-9cc9-65ec7b9e0637.png)

nearnshaw:
lol yes, that xmas scene was removed several weeks ago from that spot

pravusjif:
This will be fixed when we finish this other issue: https://github.com/decentraland/unity-renderer/issues/2132

PR #2164, opened 2022-04-13
fix: out of scene bounds entities check

pravusjif:
## WHY
We have several issues that degrade the experience due to the scene bounds checker not evaluating entities at the pertinent moment:

- https://github.com/decentraland/unity-renderer/issues/2132
- https://github.com/decentraland/unity-renderer/issues/844
- https://github.com/decentraland/unity-renderer/issues/480
- https://github.com/decentraland/unity-renderer/issues/978

## WHAT
- Implements a new preliminary scene bounds check for entities using scenes "outer bounds", that is cheaper
    - Calls that preliminary check on transforms update and on shapdes update.
- Added missing scene bounds checks for Parametrized Shapes
- Added missing scene bounds check for entities after they are reparented
- Adds scene bounds check for entities with shape but without transform
- Increases a bit the CPU time budget for messages processing and for the Scene Bounds Checker processing (they use the same budget)
- Renamed some Scene Bounds Checker methods that were ambiguous and confusing
- Adds test scene bounds checking test coverage 

### https://github.com/decentraland/unity-renderer/issues/2132 issue
1. Spawn [at -12, -10](https://play.decentraland.zone/?renderer-branch=fix/OutOfSceneBoundsEntitiesCheck&LOCAL_COMMS&position=-12%2C-10) and wait until Genesis Plaza finishes loading (make sure the Scene Load Radius value in the settings is 4) 
2. Open Settings, change the Scene Load Radius to 1 and wait 5~ seconds
3. Change Scene Load Radius to 4 and close the settings panel, wait for 5~10 seconds after the Genesis Plaza finishes loading and notice that he "black and white" bug doesn't happen anymore
4. Steps 2 and 3 can be reproduced many times and the black and white bug won't happen again 
[Doing those steps from master's build will trigger the black and white bug]

### https://github.com/decentraland/unity-renderer/issues/844 issue
1. Spawn [at -12, -10](https://play.decentraland.zone/?renderer-branch=fix/OutOfSceneBoundsEntitiesCheck&LOCAL_COMMS&position=-12%2C-10)
2. Walk towards the tramline at -12, 0
3. Walk through the tramline to the west, and notice that no mesh is instantiated int the middle of the street

### https://github.com/decentraland/unity-renderer/issues/480 and https://github.com/decentraland/unity-renderer/issues/978 issues
1. Spawn [at -40, 2](https://play.decentraland.zone/?renderer-branch=fix/OutOfSceneBoundsEntitiesCheck&LOCAL_COMMS&position=-40%2C2)
2. Walk to -50, 1 and notice that the paint bucket that attaches to the player in that scene doesn't appear before you enter the scene


pravusjif:
### Update
- Most cases of meshes outside their scene bounds are due to entities having their shape component added BEFORE their Transform component: resulting in the entity instantiated in the lower-left position of the scene base parcel (the scene local origin 0,0,0) until that entity has its Transform component and gets repositioned, rotated and scaled. 
- In Genesis Plaza many entities are having that problem, that's why they appear first in the scene's 0,0,0. Even the street parcels have that problem...
- When a scene with those kind of entities is being loaded far away from the player, due to the message processing throttling the affected entities take more time to receive their Transform component after they have their shape and are instantiated in the scene's local origin position. 

We should have the Scene Bounds Checker hide those entities when they are in the scene local origin position as they have most of their mesh outside the scene boundaries. The problem that arises with that is that after flagging those entities as "outside scene bounds" and disabling their renderers and colliders, it may take a while until they are re-evaluated and set as "inside scene bounds" again (re-enabling their renderers and colliders): there's time until those entities receive their first Transform component message (the messaging processing throttling for far away scenes prolongs that waiting time) and after that there's time until the Scene Bounds Checker actually re-evaluates those entities and detects them as being inside the scene bounds (the cpu budget throttling for the Scene Bounds Checker prolongs that waiting time when there are lots of entities to be evaluated)

### Current problem being debugged
For some reason the first time that Genesis Plaza loads (starting outside the plaza, as every test run for this implementation) the meshes are quickly processed and detected as inside boundaries, but if we reload the scene in the same session (so it starts using the cached assets in the asset library) it detects the entities as outside boundaries (because of the shapes without transform issue) and it takes A LOT of time to evaluate those entities again, due to CPU budget throttling, etc.

### Other problem found
I believe [SceneBoundsFeedbackStyle_Simple](https://github.com/decentraland/unity-renderer/blob/master/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneBoundariesController/SceneBoundsFeedbackStyle_Simple.cs) shouldn't just toggle the renderers enabled value, it should take into account that shape's model `visible` value and use that instead of always enabling the renderers. To fix that a mini refactor is needed in that class.

pravusjif:
Testing the same path with this PRs build and PROD's build, the Performance Meter Tool show a (significant?) drop in performance...
Screen Shot 2022-06-07 at 01 49 27

Screen Shot 2022-06-07 at 01 43 15

Kinerius (approved):
Looks good to me :D

Kinerius:
image

Just to add more info on why it looks good for me, the performance difference with master is insignificant

pravusjif:
By doing performance comparisons on [this other approach](https://github.com/decentraland/unity-renderer/issues/2433#issuecomment-1194156935) I found out that the first time we run the Performance Meter Tool with 30 seconds or more, we have a huge hiccup that affects its result. If we run it again after that in the same session, it doesn't happen.

The result I reported here of performance score 76 against 90 with this approach is not certain, since the new performance comparisons show this approach a lot more close to production's performance score.

Kinerius:
>  I found out that the first time we run the Performance Meter Tool with 30 seconds or more, we have a huge hiccup that affects its result. If we run it again after that in the same session, it doesn't happen.

Yeah we have to fix that or change the score so it's based on the 99/50/1 percentiles, those reduce the noise a lot

pravusjif:
In the end the https://github.com/decentraland/unity-renderer/issues/2132 and https://github.com/decentraland/unity-renderer/issues/978 issues haven't been solved by this PR.

- Black and white issue: it is only happening using graphics setting in ULTRA
- Bucket attached to player appearing out of bounds: this has been fixed for the first-time scene instantiation but if the scene is reloaded the bug is triggered... maybe it has something to fo with asset library asset reutilization.

Discovered at: https://github.com/decentraland/unity-renderer/pull/2811#issuecomment-1214669985

PR #2165, opened 2022-04-13
refactor: move character controller to new assembly

BrianAmadori:
## What does this PR change?

This PR moves the character controller to a new assembly. Therefore, it should be encapsulated.

The main affected functionality is the **Avatar** and **FirstPersonCamera** parenting, to avoid cyclic references the transforms for each case had to be moved to a new `DataStore` class.

It also contains a small fix to an inconsistency I found: When any entity is parented to the avatar and then unparented, it doesn't get removed from the "persistent" list of the `SceneBoundsChecker`. In a worst case scenario, this can reduce the accuracy of the SBC system badly and maybe impact performance. Added coverage for this fix.

## How to test the changes?

You can use `spawnScenePortableExperienceSceneFromUrn("urn:decentraland:off-chain:static-portable-experiences:portal-experience", "main")` on console to test the avatar attachment with portable experiences.

Aside of that, if the character controller works and the client doesn't crash at first glance, it should be fine.

## Our Code Review Standards


pbosio (approved):
also tested at -134,75 tennis court and it's working fine

PR #2172, opened 2022-04-18
refactor: move components handling out of entities and scene classes (I)

pbosio:
*Add class `ECSComponentsManagerLegacy` which encapsulate all components related code that was previously handled by both `DecentralandEntity` and `ParcelScene` classes.
*Every scene has it own instance of `ECSComponentsManagerLegacy`

I benchmark how much performance would be lost with this refactor, surprisingly (not entirely sure why) looks like we are reducing an average of -1ms in the components handling calls vs what we have in master

next steps:
~~*move `UUIDComponents` to their own assembly -> remove callbacks from `ECSComponentsManagerLegacy`~~ (DONE)
*TBD: check refactor for `ParcelScene` -> move `ECSComponentsManagerLegacy` to `DataStore` (I did some quick test on this a there is no performance lost in comparison with what we have on master)


Suduck (approved):

AjimenezDCL (approved):
I like the refactor, as we discussed privately, if somehow we could abstract the shared components from the components manager that would be a jackpot. GJ regardless!

PR #2200, opened 2022-04-22
fix: portable experience lifecycle issues

BrianAmadori:
## What does this PR change?

This PR cover the following fixes:
- fix: experiences viewer not showing pex as it should
- fix: global scene unloading not working at all, including pexs

Tests added to ensure coverage.

## How to test the changes?

For the first fix, ensure your portable experience is showing up in the pex viewer.

Not sure how to reproduce the unload global scenes, because this never worked at all!. The code to unload global scenes was gated under a `!isPersistent` check. We have to just make sure the avatar scene and pex are never unloaded by kernel after teleports?

## Our Code Review Standards


sandrade-dcl (approved):
Code reviewed and build tested. It is working fine! The Experiences Viewer works again and the PEXs are being loaded and unloaded correctly. Thanks!

PR #2210, opened 2022-04-25
feat: new asset promise keeper and new resource system

Suduck:
## What does this PR change?

This PR adds two new types of asset promise keeper and add a resource system on top of that. This resource system will increase complexity as we advance in the new ECS.

This also add a new model for the components PBRMaterial and DCLTexture that will be used in the future 

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Materials/Data/MaterialModel.cs
```diff
@@ -0,0 +1,83 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using DCL.Helpers;
+using UnityEngine;
+
+[System.Serializable]
+public class MaterialModel : BaseModel
```
pbosio, line 8: Model shouldn't have inheritance

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Materials/Data/MaterialModel.cs
```diff
@@ -0,0 +1,83 @@
...
+    
+    [Range(0, 4)]
+    public int transparencyMode = 4; // 0: OPAQUE; 1: ALPHATEST; 2: ALPHBLEND; 3: ALPHATESTANDBLEND; 4: AUTO (Engine decide)
+    
+    public TextureModel albedoTexture;
+    public TextureModel alphaTexture;
+    public TextureModel emissiveTexture;
+    public TextureModel bumpTexture;
+
+    public override BaseModel GetDataFromJSON(string json) { return Utils.SafeFromJson<MaterialModel>(json); }
```
pbosio, line 44: Model shouldn't contain a de-serialization method

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Materials/Data/MaterialModel.cs
```diff
@@ -0,0 +1,83 @@
...
+    public override bool Equals(object obj)
+    {
+        if (ReferenceEquals(null, obj))
+            return false;
+        if (ReferenceEquals(this, obj))
+            return true;
+        if (obj.GetType() != this.GetType())
+            return false;
+        return Equals((MaterialModel) obj);
+    }
```
pbosio, line 82: Could we move those to a helper?
We should try to keep models as just data without any methods

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Textures/Data/TextureModel.cs
```diff
@@ -0,0 +1,46 @@
+using System.Collections;
+using System.Collections.Generic;
+using DCL.Helpers;
+using UnityEngine;
+
+[System.Serializable]
+public class TextureModel : BaseModel
```
pbosio, line 7: Same comments as for `MaterialModel`

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Material/AssetPromise_Material.cs
```diff
@@ -1,17 +1,187 @@
...
+                promiseDclTexture.OnSuccessEvent += (x) =>
+                {
+                    SetMaterialTexture(material, materialProperty, x.texture2D);
+                };
+                promiseDclTexture.OnFailEvent += (x, error) =>
+                {
+                    onFail?.Invoke(error);
+                };
+
+                AssetPromiseKeeper_DCLTexture.i.Keep(emissionAssetPromiseDclTexture);
```
pbosio, line 176: if we are "Keeping" textures here... shouldn't we "Forget" them when Material promise is forgotten also?

pbosio, line 176:
```suggestion
                AssetPromiseKeeper_DCLTexture.i.Keep(promiseDclTexture);
```

unity-renderer/Assets/Scripts/MainScripts/DCL/DCLResources/ResourceManagerService.cs
```diff
@@ -0,0 +1,65 @@
...
+
+public interface IResourceManagerService : IService
+{
+    /// <summary>
+    /// Get a material based in a model, if it is created, it will return the already created material,
+    /// If not, it will create one based on it
+    /// </summary>
+    /// <param name="model"></param>
+    /// <returns></returns>
+    UniTask<Material> GetMaterial(MaterialModel model);
```
pbosio, line 15: Shouldn't we add a cancellation token for those tasks?

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Materials/Data/MaterialModel.cs
```diff
@@ -0,0 +1,83 @@
...
+            hashCode = (hashCode * 397) ^ directIntensity.GetHashCode();
+            hashCode = (hashCode * 397) ^ castShadows.GetHashCode();
+            hashCode = (hashCode * 397) ^ transparencyMode;
+            hashCode = (hashCode * 397) ^ (albedoTexture != null ? albedoTexture.GetHashCode() : 0);
+            hashCode = (hashCode * 397) ^ (alphaTexture != null ? alphaTexture.GetHashCode() : 0);
+            hashCode = (hashCode * 397) ^ (emissiveTexture != null ? emissiveTexture.GetHashCode() : 0);
+            hashCode = (hashCode * 397) ^ (bumpTexture != null ? bumpTexture.GetHashCode() : 0);
+            return hashCode;
+        }
+    }
```
pbosio, line 68: if we need to add this, I would call it with another name, without overriding `GetHashCode`
 `GetHashCode` should not depend on field values and should be immutable for the whole life of an instance.
 overriding it could generate a lot of issues that are not easy no find or debug.
 ie:
```
 Obj obj = new Obj()
 HashSet hashSet  = add(obj)
 obj.someField = change
 hashSet.contains(obj) == false
```

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Material/AssetPromise_Material.cs
```diff
@@ -1,17 +1,187 @@
...
+            if (emissionAssetPromiseDclTexture != null)
+                AssetPromiseKeeper_DCLTexture.i.Forget(emissionAssetPromiseDclTexture);
+            if (alphaAssetPromiseDclTexture != null)
+                AssetPromiseKeeper_DCLTexture.i.Forget(alphaAssetPromiseDclTexture);
+            if (baseAssetPromiseDclTexture != null)
+                AssetPromiseKeeper_DCLTexture.i.Forget(baseAssetPromiseDclTexture);
+            if (bumpPromiseDclTexture != null)
+                AssetPromiseKeeper_DCLTexture.i.Forget(bumpPromiseDclTexture);
+            
+            CoroutineStarter.Stop(loadCoroutine);
```
AjimenezDCL, line 40: We should set `loadCoroutine` to `null` after stopping it.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Material/AssetPromise_Material.cs
```diff
@@ -1,17 +1,187 @@
...
+        }
+
+        protected override void OnLoad(Action OnSuccess, Action<Exception> OnFail)
+        {
+            loadCoroutine = CoroutineStarter.Start(CreateMaterial(model, OnSuccess, OnFail));
```
AjimenezDCL, line 45: if `loadCoroutine` exists we should Stop it. Otherwise we will have leaks.

unity-renderer/Assets/Scripts/MainScripts/DCL/DCLResources/ResourceManagerService.cs
```diff
@@ -0,0 +1,65 @@
+using System.Collections;
+using System.Collections.Generic;
+using Cysharp.Threading.Tasks;
+using DCL;
+using UnityEngine;
+
+public interface IResourceManagerService : IService
```
AjimenezDCL, line 7: Why do we need this? cannot we just use the keeper?

AjimenezDCL (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Materials/Data/MaterialModel.cs
Suduck: If we don't need heritage, we could get rip of those

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Material/AssetPromise_Material.cs
Suduck: This is only called on the `OnLoad `function. If it is already loading, the `OnCancelLoading `will clean it up, and if it is already loaded the `Cleanup `method will take care of them.

I may be wrong but `Forget `is when the load didn't started yet right?

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Materials/Data/MaterialModel.cs
Suduck: We need this override since the Dictionaries uses it under the hood to work, We can't escape from overridering it. 

In order to escape, we should create our own implementation of Dictionary and use it in all the asset promises keeper.

The only thing that we can do is generate the hashCode at the creation of the object

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Materials/Data/MaterialModel.cs
Suduck: We can move the first Equals , but the other need to stay there since it is the override for Equals

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Materials/Data/MaterialModel.cs
```diff
@@ -0,0 +1,83 @@
...
+            hashCode = (hashCode * 397) ^ emissiveColor.GetHashCode();
+            hashCode = (hashCode * 397) ^ emissiveIntensity.GetHashCode();
+            hashCode = (hashCode * 397) ^ reflectivityColor.GetHashCode();
+            hashCode = (hashCode * 397) ^ directIntensity.GetHashCode();
+            hashCode = (hashCode * 397) ^ castShadows.GetHashCode();
+            hashCode = (hashCode * 397) ^ transparencyMode;
+            hashCode = (hashCode * 397) ^ (albedoTexture != null ? albedoTexture.GetHashCode() : 0);
+            hashCode = (hashCode * 397) ^ (alphaTexture != null ? alphaTexture.GetHashCode() : 0);
+            hashCode = (hashCode * 397) ^ (emissiveTexture != null ? emissiveTexture.GetHashCode() : 0);
+            hashCode = (hashCode * 397) ^ (bumpTexture != null ? bumpTexture.GetHashCode() : 0);
```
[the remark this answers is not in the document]
Suduck: I can do it, but if someone changes the model on runtime, the hash code will change and we won't be able to detect it to regenerate it

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/DCLTexture/AssetPromise_DCLTexture.cs
```diff
@@ -0,0 +1,102 @@
+﻿using System;
+using UnityEngine;
+
+namespace DCL
+{
+    public class AssetPromise_DCLTexture : AssetPromise<Asset_DCLTexture>
```
BrianAmadori, line 6: I have a big concern with this rename:

1. Is making this PR diff hard to follow
2. Why `DCLTexture`? All the other AssetPromise don't have the `DCL` prefix on it and they don't need it.
3. `DCLTexture` on the component name was needed to avoid naming conflict with `UnityEngine.Texture`. This is not the case.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Material/AssetPromise_Material.cs
```diff
@@ -1,17 +1,201 @@
...
+            if (model.alphaTexture != null)
+                yield return  alphaAssetPromiseDclTexture;
+            if (model.albedoTexture != null)
+                yield return  baseAssetPromiseDclTexture;
+            if (model.bumpTexture != null)
+                yield return  bumpPromiseDclTexture;
+            if (model.emissiveTexture != null)
+                yield return  emissionAssetPromiseDclTexture;
+
+            asset.material = material;
```
BrianAmadori, line 111: After the material is constructed, is *very important* to use `SRPBatchingHelper.OptimizeMaterial` on it. If that is not used then the SRP batching will not be effective, and this can have a huge impact in draw calls.

BrianAmadori (dismissed):
For now I'm having those two requests, I'll give this PR another pass later.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Material/AssetPromise_Material.cs
pbosio: for every `Keep` there must be a `Forget` in any scenario.
here it looks like they will only be forgotten if the load is cancelled, but what about after everything is loaded? when are texture forgotten?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/DCLTexture/AssetPromise_DCLTexture.cs
Suduck: At the beginning it was going to be an asset promise keeper of a DCLTexture and in the end it is doing that purpose somehow. However, I will call it `TextureResource `if that is ok with you!

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Material/AssetPromise_Material.cs
Suduck: Added!

unity-renderer/Assets/Scripts/MainScripts/DCL/DCLResources/ResourceManagerService.cs
Suduck: Removed

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Material/AssetPromise_Material.cs
Suduck: This is done in the `CleanPromises()` in the override of the `Cleanup`

image

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Textures/Data/TextureModel.cs
```diff
@@ -4,7 +4,7 @@
 using UnityEngine;
 
 [System.Serializable]
-public class TextureModel : BaseModel
+public class TextureModel
```
AjimenezDCL, line 7: Now that we are not coupling this models to the components maybe we should relocate them to a hierarchy outside of `DCL/Components`.

AjimenezDCL (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Textures/Data/TextureModel.cs
Suduck: Yep that is the idea, I was waiting for the Brian PR to move all the new ECS to a new plugin

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/TextureResource/AssetPromise_TextureResource.cs
```diff
@@ -0,0 +1,102 @@
+﻿using System;
+using UnityEngine;
+
+namespace DCL
+{
+    public class AssetPromise_TextureResource : AssetPromise<Asset_TextureResource>
```
BrianAmadori, line 6: Now I understand, I thought of it as a rename, but it was a different promise altogether.

The `AssetPromise_Texture` code already reuses textures with the same settings, I'm pretty sure that this promise can be removed and everything will be reused correctly.

The would not include the base64 code as its a deprecated feature and should not be part of SDK 7.

PR #2214, opened 2022-04-25
feat: CRDT protocol

pbosio:
add CRDT protocol code.
notes:
*This PR does not implement any CRDT usage
*Test were created to have matching tests with the typescript side of CRDT


unity-renderer/Assets/Scripts/MainScripts/DCL/CRDTProtocol/Tests/CRDTTests.cs
```diff
@@ -0,0 +1,117 @@
...
+using System.IO;
+using DCL;
+using Newtonsoft.Json;
+using NUnit.Framework;
+using UnityEditor;
+using UnityEngine;
+
+namespace Tests
+{
+    public class CRDTTests
```
AjimenezDCL, line 12: I understand the usage of the `.test` files which are just copied from kernel to ensure kernel-unity are aligned, but the parsing it's making too much noise. It's also not really following the AAA paradigm that we usually use for tests here.

It also makes the tests difficult to read and prevents them from becoming the onboarding point for the system.

I'd just implement each test separately. We could even use `[TestCase(CRDTMessage[] inputs, CRDTMessage output)]` and have a single test with a bunch of cases which shouldn't give much maintenance overhead to keeping up with kernel tests.

AjimenezDCL (changes requested):

Suduck (commented):
Good job!

unity-renderer/Assets/Scripts/MainScripts/DCL/CRDTProtocol/Tests/CRDTTests.cs
pbosio: I understand that is always preferable to have tests easy to read, easy to understand.
But we can't lose sight of tests primary objective that is making sure that nothing is broken.
Recreating each test from typescript to c# has a lot of maintainability costs and it's very susceptible to human error.
Having the same tests for typescript and c# means that we can be sure that any change on either side won't break compatibility with the other

AjimenezDCL: Yes, I can see the advantages from that. Okay, what about this middle ground.

Let's move the Parse to a method in a Utils class (inside the same CRDTTests assembly like this:
```csharp
public CRDTTestData[] Parse(string filePath);
```

where `CRDTTestData` is:
```csharp
public class CRDTTestData
{
    public CRDTMessage[] inputs;
    public CRDTMessage output;
}
```

And then let's have this test:
```csharp
[Test]
[TestCase("File 1")]
[TestCase("File 2")]
[TestCase("File 3")]
public void TestCRDTFiles(string filePath)
{
   var tests = CRDTTestsUtils.Parse(filePath);
   foreach(var test in tests)
   {
        AssertTest(test);
   }
}

private void AssertTest(CRDTTestData test)
{
   CRDTMessage output = null;
    foreach(var input in test.inputs)
      output = crdtProtocol.Update(input);
       
    Assert.AreEqual(output, test.output);
}
```

It's easier to maintain, we have more visibility about the files being tested and it's also easier to understand.

pbosio: I've refactored the test to make it more readable.
I did not went for the `[TestCase("File 1")]` approach because we need to be able to iterate through every `*.test` file in that test directory since we may consider fetching those test files from another repository

AjimenezDCL (approved):

Issue #2201, opened 2022-04-22
Profanity Filter is slow


Kinerius:
Every time a chat message is sent or a user name appears, a regex is being used to see if the message contains something offensive. This was a nice and quick implementation to solve a problem.

Now we need to address its performance impact.

![image.png](https://images.zenhubusercontent.com/61080c59fd1a374ce3f3b027/24620126-a34e-4457-a8ca-8122b5e91e35)

Steps to solve this:

- [x] Refactor RegexProfanityFilter usages to support async calls (https://github.com/decentraland/unity-renderer/pull/2213)
- [x] Refactor RegexProfanityFilter itself to split words and regex into steps, so every word will be checked by multiple small regexes over time, if the accumulated processing time exceeds a threshold it will wait for a frame. (#2215)
- [ ] Upgrade RegexProfanityFilter to enqueue filter requests so it will work on one string at a time so multiple calls on this on the same frame will enqueue the requests and will solve them one by one.

Kinerius:
I've just checked and performing a regex on messages is 20 times slower when deep profiling, parsing a really long message takes 28ms and 533ms when deep profiling.
Although 28 is still a very high number.

Kinerius:
I did a Proof of Concept of splitting the regex into several steps to reduce its performance impact over time, here are the results:

TEST STRING ( with some bad words, sorry for that )  
```C#
"Lofuckrem homo homogeneous ass asset ipsfuckum dolfuckor sifuckt ametfuck, fuckconsectetur adipiscing elit. Sed sed volutpat tortor, fuck dignissim elit. Suspendisse ut leo sit amet felis porttitor sodales in in elit. Maecenas viverra, nisi ac ornare volutpat, nisl ipsum faucibus est, et efficitur ante mauris a sem. Phasellus iaculis nisl eget tellus aliquam pulvinar. Curabitur pulvinar placerat dui id malesuada. Curabitur at fuck est. Phasellus dictum tempus massa, id scelerisque mi efficitur ut. Quisque venenatis sodales felis non pellentesque. Maecenas feugiat nulla rhoncus fermentum pellentesque. Sed mollis sapien risus. Etiam dignissim ligula at ullamcorper laoreet. Fuck et elit mi. Duis ex erat, gravida eu eros sed, rhoncus pellentesque mauris. Nam in elit ornare, iaculis lorem et, efficitur orci. Quisque eget augue commodo, laoreet orci elementum, consectetur justo."
```

[...]

PR #2215, opened 2022-04-25
feat: Throttled Profanity Filter

Kinerius:
## What does this PR change?

Adds a Throttling Profanity Filter to replace the old one, this one is much more performant and it throttles itself over several frames if needed.

Part of https://github.com/decentraland/unity-renderer/issues/2201 

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/ProfanityFiltering/IProfanityFilter.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/ProfanityFiltering/IProfanityFilter.cs
@@ -2,5 +2,5 @@
 
 public interface IProfanityFilter
 {
-    public UniTask<string> Filter(string message);
+    UniTask<string> Filter(string message);
 }
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/ProfanityFiltering/IProfanityFilter.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/ProfanityFiltering/IProfanityFilter.cs
@@ -1,0 +1,6 @@
+﻿using Cysharp.Threading.Tasks;
+
+public interface IProfanityFilter
+{
+    public UniTask<string> Filter(string message);
+}
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/ProfanityFiltering/ProfanityFilterSharedInstances.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/ProfanityFiltering/ProfanityFilterSharedInstances.cs
@@ -1,13 +1,13 @@
 ﻿public static class ProfanityFilterSharedInstances
 {
-    private static RegexProfanityFilter regexFilterInstance;
+    private static IProfanityFilter regexFilterInstance;
 
-    public static RegexProfanityFilter regexFilter
+    public static IProfanityFilter regexFilter
     {
         get
         {
-            return regexFilterInstance ??= new RegexProfanityFilter(
-                new ProfanityWordProviderFromResourcesJson("Profanity/badwords"));
+            return regexFilterInstance ??= new ThrottledRegexProfanityFilter(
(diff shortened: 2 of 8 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/ProfanityFiltering/ProfanityFilterSharedInstances.cs
```diff
@@ -1,13 +1,16 @@
 ﻿public static class ProfanityFilterSharedInstances
```
BrianAmadori, line 1: It's better to use a contained composite class instead of static to avoid state corruption between tests and lifecycle anomalies.

Kinerius: I agree, but I would change that on another PR

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/ProfanityFiltering/IProfanityFilter.cs
```diff
@@ -0,0 +1,6 @@
+﻿using Cysharp.Threading.Tasks;
+
+public interface IProfanityFilter
+{
+    public UniTask<string> Filter(string message);
```
lorux0, line 5: no need to declare `public`

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/ProfanityFiltering/ThrottledRegexProfanityFilter.cs
```diff
@@ -0,0 +1,77 @@
...
+            return message;
+
+        var stopwatch = new Stopwatch();
+        stopwatch.Start();
+
+        foreach (Regex regexStep in regexSteps)
+        {
+            await CheckTimerAndSkipFrame(stopwatch);
+
+            if (regexStep.IsMatch(message))
```
lorux0, line 48: do you need to do this check before replacing? it may hurt the performance since it could be doing double checking with `regex.Replace`

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/ProfanityFiltering/ThrottledRegexProfanityFilter.cs
```diff
@@ -0,0 +1,77 @@
...
+                message = regexStep.Replace(message, match => new StringBuilder().Append('*', match.Value.Length).ToString());
+            }
+        }
+
+        return message;
+    }
+
+    private async Task CheckTimerAndSkipFrame(Stopwatch stopwatch)
+    {
+        if (stopwatch.ElapsedMilliseconds > 1)
```
lorux0, line 60: couldnt this be an inaccurate check? why dont you wait a frame for every chunk?

lorux0:
works great!

lorux0 (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/ProfanityFiltering/ThrottledRegexProfanityFilter.cs
Kinerius: Some checks take less than 1 millisecond, skipping a frame for every chunk would delay even the smallest string by 
`chunks(20) * frameTime(2 to 8ms)`, and the performance impact would be the same.

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/ProfanityFiltering/ThrottledRegexProfanityFilter.cs
Kinerius: Just checked the source code, you are right, `regex.Replace` does a `Regex.Match` before doing the replace

PR #2216, opened 2022-04-26
fix: now that hasConnectedWeb3 is fixed we can use it.

AjimenezDCL:
in decentraland/kernel#261 `hasConnectedWeb3` is reliable and we can use it to know if the user is a guest.

Do not merge until kernel is ready.

## Our Code Review Standards


menduz:
Now that the new comms saga is in prod, this should be merged to fix a legacy bug

pravusjif (approved):
tested backpack in guest mode and also signed in and saw no problems

Issue #2204, opened 2022-04-22
JSON Deserializing throttle not working properly


Kinerius:
GLTF Json deserialization was modified to be throttled over time but it seems to be not working properly, we need to investigate what is causing this and how we can fix it.

![image.png](https://images.zenhubusercontent.com/61080c59fd1a374ce3f3b027/a41e842a-abd2-41e3-b503-b258002fdb0b)

Steps for resolving this issue

- [x] Find a way to throttle this correctly
- [ ] Do tests to find a sweet spot between performance impact and loading times

Kinerius:
As mentioned in the PR, this overhead came from the profiler, there is nothing to fix here.

PR #2222, opened 2022-04-27
fix: gltf json throttling

Kinerius:
## What does this PR change?

A simple fix for https://github.com/decentraland/unity-renderer/issues/2204

## Observations

Some JSON parsing hiccups went from 36ms to just 8ms ( one step of the parse ) so it is definitively throttled now 

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/UnityGLTF/Scripts/Serialization/Extensions/GLTFJsonExtensions.cs
+++ b/unity-renderer/Assets/UnityGLTF/Scripts/Serialization/Extensions/GLTFJsonExtensions.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Diagnostics;
 using DCL;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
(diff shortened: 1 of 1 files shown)


unity-renderer/Assets/UnityGLTF/Scripts/Serialization/Extensions/GLTFJsonExtensions.cs
```diff
@@ -81,15 +85,13 @@ public static IEnumerator ReadListDelayed<T>(this JsonReader reader, Func<T> des
...
                     break;
-
-                throttlingCounter++;
-                if ( throttlingCounter > throttlingCounterInterval )
+                
+                if ( sw.ElapsedMilliseconds > MAX_READ_TIME_IN_MS )
                 {
                     yield return skipFrameIfDepletedTimeBudget;
-                    throttlingCounter = 0;
+                    sw.Restart();
```
BrianAmadori, line 92: `skipFrameIfDepletedTimeBudget` is already using the throttling counter for this, so here we are having two different ways for throttling the same thing. Maybe a better approach is to just decrease the time budget for the counter? Did you try this?

Also be wary that this could increase the GLTF loading times too much.

Kinerius: Hmm, you are right, I might have to do a broader test to find a sweet spot between performance impact and loading times.

Kinerius:
The overhead came from the profiler, this fix just increases parsing time, the hiccup fix is very very small for just 1 parse over 200+

PR #2229, opened 2022-04-28
fix: entityId events issue

BrianAmadori:
## What does this PR change?

This PR fixes the issues with the global pointer event bug.

The bug was related with entityId coming from kernel as strings, converted using `GetHashCode()`, and then sent over back to kernel with the incompatible hash code id. To fix it, a helper was created to maintain a conversion `<long, string>` dictionary to convert the longs back to the original format before being sent back to Kernel again.

## How to test the changes?

**Space office door doesn’t open**
The interaction to open the door using the card key was not working (the door should open after clicking it)

It can be reproduced quickly by entering the world at:
https://play.decentraland.zone/?renderer-branch=fix/entityid-events-issue&NETWORK=ropsten&position=64,-64

**Shooting range scene stopped instantiating shot decals**
The bullseye targets should instantiate a shot decal on every target you click, after positioning yourself in the red rectangle on the floor.

Link: https://play.decentraland.zone/?renderer-branch=fix/entityid-events-issue&position=-14,1

## Our Code Review Standards


BrianAmadori:
Tested and works for the two cases described in the PR. Can't approve it because I opened it. @Suduck maybe you can approve it for me?

AjimenezDCL (approved):

PR #2242, opened 2022-05-02
feat: CRDT deserializer

pbosio:
add code to de-serialize incoming binary messages (like CRDTs)
we are planning for web to write messages directly on wasm memory and send the pointer to unity and for ws to send a byte array


unity-renderer/Assets/Scripts/MainScripts/DCL/CRDTProtocol/Deserializer/CRDTDeserializer.cs
```diff
@@ -0,0 +1,31 @@
+using KernelCommunication;
+
+namespace DCL.CRDT
+{
+    // Deserialize CRDT binary messages (BigEndian)
```
kuruk-mm, line 5: Is the endianness at this point relevant? I think it should depend on the implementation of the IBinaryReader.

kuruk-mm (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/CRDTProtocol/Deserializer/CRDTDeserializer.cs
pbosio: 🤷‍♂️ I don't know. I imagine that if someone has to modify something regarding CRDT messages they will probably land on this scripts before reaching the `IBinaryReader` and I think that knowing the endianness of the message could be relevant. So I left that comment there just in case :P

pravusjif (approved):
LGTM

PR #2244, opened 2022-05-03
fix: change to decode instead of GetHashCode for entity ids

Suduck:
## What does this PR change?

This PR changes the way we get the long number for an entity id from a string, we were using the GetHashCode approach but since that approach can make some iDs to collide I have changed it to decode the base36 string

I also has done some perfomance benchmark to ensure that we are not losing too much perfomance with this changes.

For the benchmark I just stay at the casino while everthing is loading. Since a lots of messages are sent in the casino scene, we can meassure if there is an impact when a lot of entityIds are converted

Testing in the casino , we lose 0.01 seconds (in total) when you are standing in the casino for 30 seconds with the conversion . The GetHashCode approach lose less than 0.00001 seconds

## How to test the changes?

1. Go to: https://play.decentraland.org/?renderer-branch=fix/entity-collision&position=-75,78
2. Everything work as expected

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Interfaces/EntityIdHelper.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Interfaces/EntityIdHelper.cs
@@ -64,7 +64,7 @@ namespace DCL
                 {
                     entityIdLong = negativeCounter;
                     negativeValues.Add(entityId,negativeCounter);
-                    negativeCounter++;
+                    negativeCounter--;
                 }
             }
             else
@@ -78,7 +78,7 @@ namespace DCL
             return entityIdLong;
         }
         
-        public Int64 DecodeBase36(string input)
+        public long DecodeBase36(string input)
         {
             const string charList = "0123456789abcdefghijklmnopqrstuvwxyz";
             var reversed = input;
@@ -89,37 +89,6 @@ namespace DCL
                 result += charList.IndexOf(reversed[i]) * (long)Math.Pow(36, pos);
                 pos++;
             }
-            return result;
-        }
-        
-        public string ToBase36(long decimalNumber, int radix)
-        {
-            const int bitsInLong = 64;
-            const string digits = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ";
-
-            if (radix < 2 || radix > digits.Length)
-                throw new ArgumentException("The radix must be >= 2 and <= " + digits.Length);
-
-            if (decimalNumber == 0)
-                return "0";
-
-            int index = bitsInLong - 1;
-            long currentNumber = Math.Abs(decimalNumber);
-            char[] charArray = new char[bitsInLong];
-
-            while (currentNumber != 0)
-            {
-                int remainder = (int)(currentNumber % radix);
-                charArray[index--] = digits[remainder];
-                currentNumber = currentNumber / radix;
-            }
-
-            string result = new String(charArray, index + 1, bitsInLong - index - 1);
-            if (decimalNumber < 0)
-            {
-                result = "-" + result;
-            }
-
             return result;
         }
     }
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Interfaces/EntityIdHelper.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Interfaces/EntityIdHelper.cs
@@ -81,12 +81,11 @@ namespace DCL
         public long DecodeBase36(string input)
         {
             const string charList = "0123456789abcdefghijklmnopqrstuvwxyz";
-            var reversed = input;
             long result = 0;
             int pos = 0;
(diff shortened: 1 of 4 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Interfaces/EntityIdHelper.cs
```diff
@@ -35,12 +39,57 @@ public long EntityFromLegacyEntityString(string entityId)
...
+            
             if (!entityIdToLegacyId.ContainsKey(entityIdLong))
                 entityIdToLegacyId[entityIdLong] = entityId;
 
             return entityIdLong;
         }
+        
+        public Int64 DecodeBase36(string input)
+        {
+            string charList = "0123456789abcdefghijklmnopqrstuvwxyz";
```
AjimenezDCL, line 52: Make this a `const` to avoid allocation. You can do it here, it doesn't need to go on top of the class.

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Interfaces/EntityIdHelper.cs
```diff
@@ -35,12 +39,57 @@ public long EntityFromLegacyEntityString(string entityId)
...
+            var reversed = input.ToLower().Reverse();
```
AjimenezDCL, line 53: We can boost performance by avoiding the `Reverse` using a loop from end to start.

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Interfaces/EntityIdHelper.cs
```diff
@@ -35,12 +39,57 @@ public long EntityFromLegacyEntityString(string entityId)
...
+            }
+
+            string result = new String(charArray, index + 1, bitsInLong - index - 1);
+            if (decimalNumber < 0)
+            {
+                result = "-" + result;
+            }
+
+            return result;
+        }
```
AjimenezDCL, line 93: This is not being used by anobyd. Is it needed?

AjimenezDCL (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Interfaces/EntityIdHelper.cs
```diff
@@ -35,12 +39,57 @@ public long EntityFromLegacyEntityString(string entityId)
...
-
+            long entityIdLong = DecodeBase36(entityId) << 9;
+            
             if (!entityIdToLegacyId.ContainsKey(entityIdLong))
                 entityIdToLegacyId[entityIdLong] = entityId;
 
             return entityIdLong;
         }
+        
+        public Int64 DecodeBase36(string input)
```
kuruk-mm, line 50: Why are we using `Int64` here instead of `long`?

I know that they are almost the same, but we should have consistency with the type name.

kuruk-mm (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Interfaces/EntityIdHelper.cs
kuruk-mm: x2

menduz: +10

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Interfaces/EntityIdHelper.cs
Suduck: Yes for Unit testing

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Interfaces/EntityIdHelper.cs
```diff
@@ -49,13 +80,13 @@ public long EntityFromLegacyEntityString(string entityId)
         
         public Int64 DecodeBase36(string input)
         {
-            string charList = "0123456789abcdefghijklmnopqrstuvwxyz";
-            var reversed = input.ToLower().Reverse();
+            const string charList = "0123456789abcdefghijklmnopqrstuvwxyz";
+            var reversed = input;
```
AjimenezDCL, line 84: We don't need this. We can remove this variable and just use `input`

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Interfaces/EntityIdHelper.cs
Suduck: Just a copy paste mistake!

PR #2253, opened 2022-05-04
feat: added some debug settings

Kinerius:
## What does this PR change?

Adds some debug options to tweak how the renderer behaves on Unity editor, this is mostly used when profiling:

1.  Disable Asset Bundles
2.  Disable Texture Compression
3.  Disable GLTF Download Throttling

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/DataStore/DataStore_Performance.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/DataStore/DataStore_Performance.cs
@@ -4,6 +4,5 @@ namespace DCL
     {
         public readonly BaseVariable<bool> multithreading = new BaseVariable<bool>(false);
         public readonly BaseVariable<int> maxDownloads = new BaseVariable<int>(10);
-        public readonly BaseVariable<bool> disableTextureCompression = new BaseVariable<bool>(false);
     }
 }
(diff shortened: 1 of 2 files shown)


kuruk-mm (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Textures/DCLTexture.cs
```diff
@@ -121,7 +121,12 @@ public override IEnumerator ApplyChanges(BaseModel newModel)
                     {
                         texture.wrapMode = unityWrap;
                         texture.filterMode = unitySamplingMode;
-                        texture.Compress(false);
+
+                        if (!DataStore.i.performance.disableTextureCompression.Get())
+                        {                       
+                            texture.Compress(false);
+                        }
```
pbosio, line 128: shouldn't we use preprocessor directive `#if UNITY_EDITOR` here and everywhere else in this PR?

Kinerius: Yes and no, I don't want to lose the ability to compress textures in editor since we need to run the game as close as possible with production settings. Hence the toggle

pbosio: what about
```
#if UNITY_EDITOR
                        if (!DataStore.i.performance.disableTextureCompression.Get())
                        {                       
                            texture.Compress(false);
                        }
#else
                        texture.Compress(false);
#endif               
```         
so the check if avoided in prod

Kinerius: Yeah I can live with that

AjimenezDCL:
@Kinerius what's the status of this? how is this affected by #2450 ?

PR #2268, opened 2022-05-06
Feat: UI loading wheel

JuanSalice:
## What does this PR change?

This PR removes the loading spinner based on sprites, and replaces it with a spinner made with shader graph. New materials where added for the different uses of the spinners.

...

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=feat/UI-Loading-Wheel
2. Open the main menu
3. Verify the correct behaviour of the loading spinners

## Our Code Review Standards


unity-renderer/Assets/Rendering/UI/ChargeWheel/UILoadingSpinner.cs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    void GetMaterial()
+    {
+        _img = GetComponent<Image>();
+
+        if (_img.maskable)
+        {
+            _material = new Material(_img.material);
+            _material = _img.materialForRendering;
```
sandrade-dcl, line 27: Why are you doing here the `new Material` assignation? As I can see, the next instruction is already setting the `_material` variable.

```suggestion
            _material = _img.materialForRendering;
```

sandrade-dcl (changes requested):
I left some chenage requests.

Apart of them, I would like to add this: Could we move all the loading spinner assets (scripts, materials, animations, shaders, etc.) to `Assets/UIComponents/LoadingSpinnerAssets`? I think it would make more sense to have them in the UI Components folder.

sandrade-dcl (approved):

RominaMarchetti (approved):

PR #2288, opened 2022-05-10
fix: PBRMaterial ref count

pbosio:
In some scenarios `PBRMaterial` was not being dettach from entities when the scene was unloaded, leaving an inaccurate metric count
closes https://github.com/decentraland/sdk/issues/296


kuruk-mm (approved):

kuruk-mm:
Tested and it's working! Thanks!

sandrade-dcl (approved):

PR #2289, opened 2022-05-10
Fix: allow user to play emotes on moving platforms

### LINKED ISSUE #2251: Can't do emotes on platforms
When a player is standing still on a moving platform, they can't perform any emotes

You can see this clearly in genesis plaza
If you get on the floating train or the balloon that take you around the plaza, you can't do any emotes while the platforms move. My avatar is still doing only an idle animation. Once the platforms finish their path and stay still, then you can do emotes.

This would be especially good to have for parades, where players on moving mobiles can dance or wave at the crowds

davidejensen:
## What does this PR change?

Fix #2251
Users can now play emotes on moving platforms

...

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=fix/platform-anim
2. In genesis plaza go on the air balloon
3. Try to play emotes on it and verify it works
4. Play emotes in other contexts and verify a regular behavior

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
@@ -49,7 +49,7 @@ public class DCLCharacterController : MonoBehaviour
     Vector3 velocity = Vector3.zero;
 
     public bool isWalking { get; private set; } = false;
-    public bool isMoving { get; private set; } = false;
+    public bool isMovingByUserInput { get; private set; } = false;
     public bool isJumping { get; private set; } = false;
     public bool isGrounded { get; private set; }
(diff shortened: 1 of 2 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
```diff
@@ -49,6 +49,7 @@ public class DCLCharacterController : MonoBehaviour
     Vector3 velocity = Vector3.zero;
 
     public bool isWalking { get; private set; } = false;
+    public bool isMoving { get; private set; } = false;
```
pravusjif, line 52: Can we rename this to `isMovingByUserInput` or similar?

pravusjif (approved):
Nice job, I also think the solution is to bind that emotes escaping to character movement by user input instead of just character movement.

PR #2302, opened 2022-05-12
refactor: loading state of scene for new ECS

Suduck:
## What does this PR change?

This adds a new way to load the scenes for ECS 7. Basically it will take into account if the components are ready or not to start the scene

## Our Code Review Standards


AjimenezDCL (commented):
What are we going to do about this?

Suduck:
> What are we going to do about this?

I am going to refactor it with the solution that we agree

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/BoxShape/ECSBoxShapeComponentHandler.cs
```diff
@@ -39,6 +46,7 @@ public void OnComponentModelUpdated(IParcelScene scene, IDCLEntity entity, ECSBo
                 DisposeMesh(scene);
                 generatedMesh = shape.mesh;
                 GenerateRenderer(generatedMesh, scene, entity, model);
+                OnResourceReady?.Invoke(this);
```
pbosio, line 49: why can't we have the tracking just in the DataStore and call here something like `DataStore.i.ecs7.SetReady(...)`

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ECSRuntime/Components/Interfaces/IECSComponent.cs
```diff
@@ -1,3 +1,4 @@
+using System;
```
pbosio, line 1: ?

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ECSRuntime/ComponentsManager/ECSComponentsManager.cs
```diff
@@ -1,11 +1,14 @@
...
 using System.Collections.Generic;
 using DCL.Controllers;
 using DCL.Models;
 
 namespace DCL.ECSRuntime
 {
-    public class ECSComponentsManager
+    public class ECSComponentsManager : IECSComponentsManager
     {
+        public event Action<IECSComponent> OnComponentAdded;
```
pbosio, line 10: why do we need this callback?

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneResourcesLoadTracker/ResourcesLoadTrackerLegacyECS.cs
```diff
@@ -29,7 +29,7 @@ public float loadingProgress
         public ResourcesLoadTrackerLegacyECS(IECSComponentsManagerLegacy componentsManager, IWorldState worldState)
         {
             this.componentsManager = componentsManager;
-            this.worldState = worldState;
+            this.worldState = worldState;
```
pbosio, line 32: ?

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneResourcesLoadTracker/Tests/MockedSharedComponentHelper.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 using DCL.Components;
+using DCL.ECSRuntime;
```
pbosio, line 3: ?

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneResourcesLoadTracker/Tests/SceneResourcesLoadTrackerShould.cs
```diff
@@ -1,100 +1,117 @@
...
 using DCL.WorldRuntime;
 using NSubstitute;
+using NSubstitute.Extensions;
 using NUnit.Framework;
+using UnityEngine;
+using UnityEngine.TestTools;
 
 namespace Tests
 {
     public class SceneResourcesLoadTrackerShould
```
pbosio, line 16: this test ideally should be agnostic from the ecs version, so the modification must be according to that... mocking the behavior of both system and testing if we got the expected result.
Maybe wrapping both in a class to mock their behavior and using TestCase for each one of them.
please add a new test for the ecs7 specific loading

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneResourcesLoadTracker/Tests/SceneResourcesLoadTrackerShould.cs
```diff
@@ -1,100 +1,96 @@
...
+using DCL.ECSComponents;
+using DCL.Models;
 using DCL.WorldRuntime;
 using NSubstitute;
+using NSubstitute.Extensions;
 using NUnit.Framework;
+using UnityEngine;
 
-namespace Tests
+namespace DCL.Test
```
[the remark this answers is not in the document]
Suduck: We already have those tests in 3 separated files

`SceneResourcesLoadTrackerShould ` is completely agnostic to the system, it doesn't use one or the other
`SceneResourcesLoadTrackerForECS7 ` that is the scene resource loader for the new ECS. It check the system with the new system
`ResourcesLoadTrackerLegacyECSShould` Check the resource loader system with the old system.

This way we can then delete the Legacy system without the need to touch anything.

Why we would only want to have the two last cases in one file?

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/BoxShape/ECSBoxShapeComponentHandler.cs
```diff
@@ -57,6 +57,8 @@ internal void DisposeMesh(IParcelScene scene)
                 ECSComponentsUtils.DisposeMeshInfo(meshesInfo);
             if(rendereable != null)
                 ECSComponentsUtils.RemoveRendereableFromDataStore( scene.sceneData.id,rendereable);
+            if(model != null)
+                DataStore.i.ecs7.pendingSceneResources.DecreaseRefCount((scene.sceneData.id, model));
```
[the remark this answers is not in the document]
Suduck: If a component fails I m not sure if we should load the scene anyway. 

I will add it because right now the initial loading is waiting for other scenes to load, but if a component of the scene fail loading we should handle that appart

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneLifecycleHandler.cs
```diff
@@ -53,6 +53,19 @@ public SceneLifecycleHandler(ParcelScene ownerScene)
...
+            // This should disappear when the old ecs is removed from the project. This should be the default track 
+            DataStore.i.ecs7.pendingSceneResources.OnRefCountUpdated += ( sceneId,  amount) =>
+            {
+                if (sceneId.sceneId != ownerScene.sceneData.id)
+                    return;
+                
+                sceneResourcesLoadTracker.Dispose();
+                sceneResourcesLoadTracker.Track(sceneId.sceneId);
+                sceneResourcesLoadTracker.OnResourcesStatusUpdate += OnResourcesStatusUpdated;
+            };
```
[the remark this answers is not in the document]
Suduck: If we want to load a scene without pending resources, the scene will load normally with the old tracker because the `sceneResourceLoadTracker.ShouldWaitForPendingResources()`  will return false since there is no pending resources.

If no resources are never added, then we won't change the system loading tracker because it will work the same

It has more performance this way than doing everytime that a CRDT messages comes.

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/BoxShape/ECSBoxShapeComponentHandler.cs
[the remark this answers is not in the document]
pbosio: If component fail, we should load the scene anyway.
Handling component failing to load should be responsibility of other system

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ECSRuntime/Components/Interfaces/IECSComponentsManager.cs
```diff
@@ -0,0 +1,46 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using DCL.ECSRuntime;
+using DCL.Models;
+using UnityEngine;
+
+public interface IECSComponentsManager
```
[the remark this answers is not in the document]
pbosio: I can think of any reason for mocking that, but if the we need to, then we might add the interface at that moment

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneLifecycleHandler.cs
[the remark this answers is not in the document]
pbosio: yeah, but remember we want to delete the old ecs code in the future, so we can't depend on the other tracker.
you could use the event raised by `DataStore.i.ecs7.componentsManagers` since it would work just like this one but it won't depend on resources needed to be loaded

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneResourcesLoadTracker/Tests/SceneResourcesLoadTrackerShould.cs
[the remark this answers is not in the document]
pbosio: `ResourcesLoadTrackerLegacyECSShould` should test that ecs tracker
`SceneResourcesLoadTrackerForECS7` should test that ecs tracker
`SceneResourcesLoadTrackerShould` should test the tracker integration with `SceneResourcesLoadTracker` so we need to test that integration with both since both exists

unity-renderer/Assets/Scripts/MainScripts/DCL/DataStore/DataStore_ECS7.cs
```diff
@@ -1,11 +1,13 @@
...
 using DCL.Controllers;
 using DCL.ECSRuntime;
 
 namespace DCL
 {
     public class DataStore_ECS7
     {
-        public BaseDictionary<IParcelScene, ECSComponentsManager> componentsManagers = new BaseDictionary<IParcelScene, ECSComponentsManager>();
+        public BaseDictionary<IParcelScene, IECSComponentsManager> componentsManagers = new BaseDictionary<IParcelScene, IECSComponentsManager>();
+        public readonly BaseRefCountedCollection<(string sceneId, object model)> pendingSceneResources = new BaseRefCountedCollection<(string sceneId, object model)>();
```
AjimenezDCL, line 10: This makes really hard to find all the blockers for a sceneId.

It would be better to use:
```csharp
public readonly Dictionary<string, BaseRefCountedCollection<object>> pendingSceneResources = new Dictionary<string, BaseRefCountedCollection<object>>();
```

Also I'm not a big fan of having an `object` there, maybe would it be better to have a componentId or something?

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneLifecycleHandler.cs
```diff
@@ -53,6 +53,21 @@ public SceneLifecycleHandler(ParcelScene ownerScene)
             sceneResourcesLoadTracker = new SceneResourcesLoadTracker();
             sceneResourcesLoadTracker.Track(owner.componentsManagerLegacy, Environment.i.world.state);
             sceneResourcesLoadTracker.OnResourcesStatusUpdate += OnResourcesStatusUpdated;
+   
+            // This is done while the two ECS are living together, if we detect that a component from the new ECS has incremented a 
+            // resource for the scene, we changed the track since that means that this scene is from the new ECS.
+            // This should disappear when the old ecs is removed from the project. This should be the default track 
+            DataStore.i.ecs7.pendingSceneResources.OnRefCountUpdated += ChangeTrackingSystem;
```
AjimenezDCL, line 60: This looks like a really weird way of solving an issue.

Cannot we detect the version of the SDK before adding it to the tracker? Shouldnt the scenes bring a sdk version on its metadata?

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneResourcesLoadTracker/ResourcesLoadTrackerECS.cs
```diff
@@ -0,0 +1,90 @@
...
+            int resourcesReadyCount = resourcesReady.Count;
+            return resourcesReadyCount > 0 ? (resourcesReadyCount - pendingResourcesCount) * 100f / resourcesReadyCount : 100f;
+        }
+    }
+    
+    public event Action OnResourcesLoaded;
+    public event Action OnStatusUpdate;
+    
+    private readonly List<object> resourcesNotReady = new List<object>();
+    private readonly List<object> resourcesReady = new List<object>();
```
AjimenezDCL, line 24: Do we need to know the ready resources? what's the need?

This class might even dissapear if we just use a Dictionary as I mentioned above to track all the resources per scene.

AjimenezDCL (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneLifecycleHandler.cs
Suduck: Right now we don't identify scenes by versions or any means. Since the ECS7 will be the default in the future, I talked with pato and this is an easy way to implement it until we can remove all the code.

However, I will change this event for this one `DataStore.i.ecs7.componentsManagers.OnAdded`

unity-renderer/Assets/Scripts/MainScripts/DCL/DataStore/DataStore_ECS7.cs
Suduck: We need the model here so we can know exactly which resource is pending (imagine having different images, we need to know which one is loaded or pending) . If we want to remove the object here, we need to implement something like `IResource ` and implement it to every resource that is in the project so we can track the model of the resource. I thought it was an overengineer for the purpose

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneResourcesLoadTracker/ResourcesLoadTrackerECS.cs
Suduck: I added it for debugging reasons, since you can always check which models are loading and which one not, but I will change it for a int

AjimenezDCL (approved):

pbosio (approved):

PR #2319, opened 2022-05-17
feat: bitmap loading

Kinerius:
This PR is a monster, don't look at it.

TLDR: Changed `.LoadImage` for C# Bitmap class to load JPG/PNG images to get their RGBA data, its also faster resizing them before converting it into a unity texture and then compressing it.

Test:  https://play.decentraland.zone/?renderer-branch=feat/bitmap-loading


AjimenezDCL:
@Kinerius We can close this, right?

PR #2326, opened 2022-05-18
feat: add audio stream component for new ECS

Suduck:
## What does this PR change?

This PR creates the new audio stream component for the new ECS.

In order to pass it to the new ECS, I have refactored how it worked and fixed a couple of corner cases bugs 

## How to test the changes?

You can't

## Our Code Review Standards


unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/AudioSource/ECSAudioSourceComponentHandler.cs
```diff
@@ -55,23 +60,27 @@ public void OnComponentRemoved(IParcelScene scene, IDCLEntity entity)
...
 
-            keeperAudioClip.Keep(promiseAudioClip);
-            
-            sceneID.OnChange -= OnCurrentSceneChanged;
-            sceneID.OnChange += OnCurrentSceneChanged;
+            if (!isAudioClipReady && !isSameClip)
+            {
+                promiseAudioClip = new AssetPromise_AudioClip(model.audioClipUrl, scene.contentProvider);
+                promiseAudioClip.OnSuccessEvent += OnAudioClipLoadComplete;
+                promiseAudioClip.OnFailEvent += OnAudioClipLoadFail;
```
davidejensen, line 80: This two events are not unsubbed anywhere, you should even if it's an AssetProsime, right?

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/AudioSource/ECSAudioSourceComponentHandler.cs
```diff
@@ -124,10 +135,16 @@ private void ApplyCurrentModel()
                     audioSource.Stop();
                 }
             }
+            else if(isAudioClipReady)
+            {
+                // PlayOneShot is faster ( about twice faster) than their Play counterpart
+                audioSource.PlayOneShot(audioClip);
```
davidejensen, line 141: If I remember correctly when using PlayOneShot there is no possibility of Stopping the clip, it just stops when the clip is over, when applying the model there is a call to audioSource.Stop(), could this be problematic?

sandrade-dcl (approved):
LGTM

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/AudioSource/ECSAudioSourceComponentHandler.cs
Suduck: There is no need since we are just recreating the asset promise and when you forgot an asset promise it gets removed, anyway I will change just in case

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/AudioSource/ECSAudioSourceComponentHandler.cs
Suduck: You are right. I didn't take into account that we may want to stop the audio

davidejensen (approved):

PR #2330, opened 2022-05-19
feat: implement the box shape correctly

Suduck:
## What does this PR change?

This PR makes the box shape available for the new ECS so it can work automatically.

This way if a new runtime ECS send a component with the same component ID it will create the box shape

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/ECS7ComponentsPlugin.cs
+++ b/unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/ECS7ComponentsPlugin.cs
@@ -1,17 +1,26 @@
 using System.Collections;
 using System.Collections.Generic;
+using System.IO;
 using DCL;
+using DCL.ECSComponents;
 using UnityEngine;
 
 public class ECS7ComponentsPlugin : IPlugin
 {
     public ECS7ComponentsPlugin()
     {
-        // When we have clear how do we want to create components in the scenes, we should register here the components
+        RegisterComponents();
+    }
+
+    public void RegisterComponents()
+    {
+        DataStore.i.ecs7.componentsFactory.AddOrReplaceComponent(ComponentID.BOX_SHAPE,
+            data => PBBoxShape.Parser.ParseFrom((byte[])data),
+            () =>  new ECSBoxShapeComponentHandler());
     }
     
     public void Dispose()
     {
-        
+        DataStore.i.ecs7.componentsFactory.RemoveComponent(ComponentID.BOX_SHAPE);
     }
 }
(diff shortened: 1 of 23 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Utils.cs
```diff
@@ -58,6 +59,23 @@ public static void CleanMaterials(Renderer r)
...
+            int uvsResultIndex = 0;
+
+            for (int i = 0; i < uvs.Count;)
+            {
+                Vector2 tmpUv = Vector2.zero;
+                tmpUv.x = uvs[i++];
+                tmpUv.y = uvs[i++];
+
+                uvsResult[uvsResultIndex++] = tmpUv;
+            }
```
pbosio, line 74: Not sure if this won't create some unnecessary defensive copies of the `Vector2` struct.
maybe just in case it might be better just to create the vector when assigning it to the result ` uvsResult[uvsResultIndex++] = new Vector2(uv1,uv2);`?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/PluginSystem/PluginSystemFactory.cs
```diff
@@ -34,6 +34,7 @@ public static PluginSystem Create()
             pluginSystem.Register<UUIDEventsPlugin>(() => new UUIDEventsPlugin());
             pluginSystem.Register<UIComponentsPlugin>(() => new UIComponentsPlugin());
             pluginSystem.Register<CoreComponentsPlugin>(() => new CoreComponentsPlugin());
+            pluginSystem.Register<ECS7ComponentsPlugin>(() => new ECS7ComponentsPlugin());
```
pbosio, line 37: just in case, let's not add the plugin to the factory yet :)
since we don't need to do it right now, I would prefer to avoid executing any ecs7 code in prod for the moment

pbosio (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Utils.cs
Suduck: This should be fixed from the Unity from 2 or 3 years If I remember correctly, but I will change it just in case

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/ECS7ComponentsPlugin.cs
```diff
@@ -1,17 +1,26 @@
...
+    {
+        DataStore.i.ecs7.componentsFactory.AddOrReplaceComponent(ComponentID.BOX_SHAPE,
+            data => PBBoxShape.Parser.ParseFrom((byte[])data),
+            () =>  new ECSBoxShapeComponentHandler());
     }
     
     public void Dispose()
     {
-        
+        DataStore.i.ecs7.componentsFactory.RemoveComponent(ComponentID.BOX_SHAPE);
```
AjimenezDCL, line 24: The DataStore can be provided throught Dependency Injection so we can mock it and avoid singletons.

AjimenezDCL (approved):
Approving with small suggestions

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/ECS7ComponentsPlugin.cs
AjimenezDCL: I talked about this with Pato on Friday, I forgot that we have the non-written rule about having parameter-less constructors for plugins. So in this case this usage is correct.

AjimenezDCL: In the near future DataStore will be part of the ServiceLocator

PR #2332, opened 2022-05-19
fix: collider bug

Suduck:
## What does this PR change?

...

## How to test the changes?

2. ...

## Our Code Review Standards


pravusjif (approved):
We can't remove that collisions hack because of the scenes having a part of the SDK used to build them when they are deployed, a lot of old scenes have a wrong default value in their `withCollisions`.

The best possible solution would be to create a new `withCollisions` property but called differently, with the correct default value (as it wouldn't be in the SDK deployed with the scene, it would use the correct one), but that requires more changes from the SDK side.

Tested with golfcraft collisions: https://play.decentraland.zone/?renderer-branch=fix%2Fcollider-bug&position=48%2C-45&realm=dg&island=Iif5

PR #2335, opened 2022-05-19
fix: CJK fonts are not being rendered

sandrade-dcl:
## What does this PR change?
When you write Chinese, Japanese or Korean symbols, they are not being rendered in the UI.

## How to test the changes?
1. Go to: https://play.decentraland.zone/?renderer-branch=fix/cjk-fonts-issues
2. Try to write different phrases in chinese, japanese and korean (copy-paste from the translator for that if needed) in the chat.
3. Notice the text is being showed correctly.

## Our Code Review Standards


pravusjif (approved):
Tested copypasting japanese hiragana & katakana + simplified chineses with no problems.

FYI: the Unity.data file size changes from 56mb to 71mb with this change.

PR #2338, opened 2022-05-19
chore: moved chat entries prefabs to resources folder

lorux0:
## What does this PR change?

Fixes `DefaultChatEntryShould` on desktop project.
We cannot set a fixed path for a prefab since desktop project changes the path cause the project is loaded as a package dependency.

## Our Code Review Standards


Kinerius (approved):
LGTM

PR #2341, opened 2022-05-20
feat: add ecs7 transform component

pbosio:


Suduck (approved):
Good Job! A bit concerned about the transform using the entity ID for parenting since we are coupling the entity and the transform too much

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Transform/ECSTransformPlugin.cs
```diff
@@ -0,0 +1,24 @@
...
+        private readonly ECSComponentsFactory factory;
+        private readonly int componentId;
+
+        public ECSTransformPlugin(int componentId, ECSComponentsFactory factory)
+        {
+            ECSTransformHandler handler = new ECSTransformHandler();
+            factory.AddOrReplaceComponent(componentId, ECSTransformSerialization.Deserialize, () => handler);
+            this.factory = factory;
+            this.componentId = componentId;
+        }
```
AjimenezDCL, line 17: Not sure if it's a good idea, but this starts to look like we could implement a `ECSComponentPlugin<THandler>` class.

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Transform/ECSTransformPlugin.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+using DCL.ECSRuntime;
+
+namespace DCL.ECSComponents
+{
+    public class ECSTransformPlugin : IDisposable
```
AjimenezDCL, line 6: Shouldn't this be a `IPlugin`?

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Transform/Serialization/ECSTransformSerialization.cs
```diff
@@ -0,0 +1,57 @@
+using System.IO;
+using BinaryWriter = KernelCommunication.BinaryWriter;
+
+namespace DCL.ECSComponents
+{
+    public static class ECSTransformSerialization
```
AjimenezDCL, line 6: I'm a bit confused, is this not going to be a protobuff in the future?

AjimenezDCL (approved):
Looks good, left some small requests

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Transform/ECSTransformPlugin.cs
pbosio: yeah, that might not be a bad idea, we would have to analyze it to make sure that won't constrain new components, since not every component may need to be created in the same standardized way

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Transform/ECSTransformPlugin.cs
pbosio: this class will be instantiated inside the ecs7 plugin, it won't be added directly to the plugin factory

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Transform/Serialization/ECSTransformSerialization.cs
pbosio: we are not going to use protobuf for transform component because it faster this way and it is the component that is updated more frequently

PR #2345, opened 2022-05-23
feat: update Unity to 2020.3.34f1

Kinerius:
## What does this PR change?

Features: 
- Unity version is now 2020.3.34f1
- TextMesh Pro is now 3.20-pre.3
- TextMesh Pro essentials were re-imported
- Universal Renderer Pipeline is now 10.6.0

Fixes https://github.com/decentraland/explorer-desktop/issues/287

Desktop PR: https://github.com/decentraland/explorer-desktop/pull/307

## How to test the changes?

macOS Desktop: World and UI texts should be visible when they share the same materials. Those that weren't working before was the bartender's text bubble and your Account name when opening your profile menu.

## Our Code Review Standards


sandrade-dcl (changes requested):
The app freezes completely just when I try to teleport to WONDERMINE from the PLACES section:
1. Enter the world to 0,0 coords.
2. Open the Start Menu.
3. Go to the PLACES tab.
4. Try to teleport to WONDERMINE place from its card [JUMP IN] button.
5. Notice the app freezes!

Kinerius:
@sandrade-dcl Thank you for testing! 

Right now if you teleport right after loading genesis plaza, there are 100% chances it will crash, something changed with the Loading/Unloading of AssetBundles in this version of unity and there's an operation that's making everything fall apart.

pravusjif:
> @sandrade-dcl Thank you for testing!
> 
> Right now if you teleport right after loading genesis plaza, there are 100% chances it will crash, something changed with the Loading/Unloading of AssetBundles in this version of unity and there's an operation that's making everything fall apart.

For me the build is only getting frozen when I use the "Explore/Places" teleportation as Santi mentioned, so there must be some using that UI.

- Tested teleporting several times with `/goto random` with no issues
- Tested teleporting using the map (to Wonder mine and other places) with no issues 
- Tested teleporting with `/goto -27, 58` to wonder mine with no issues
- Tested teleporting to wonder mine from the Genesis Plaza "Crowded" column with no issues
- Tested walking from the Genesis plaza to The Salmonomicon and finished the game with no issues

Kinerius:
> > @sandrade-dcl Thank you for testing!
> > Right now if you teleport right after loading genesis plaza, there are 100% chances it will crash, something changed with the Loading/Unloading of AssetBundles in this version of unity and there's an operation that's making everything fall apart.
> 
> For me the build is only getting frozen when I use the "Explore/Places" teleportation as Santi mentioned, so there must be something going on with using that UI.
> 
> * Tested teleporting several times with `/goto random` with no issues
> * Tested teleporting using the map (to Wonder mine and other places) with no issues
> * Tested teleporting with `/goto -27, 58` to wonder mine with no issues
> * Tested teleporting to wonder mine from the Genesis Plaza "Crowded" column with no issues
> * Tested walking from the Genesis plaza to The Salmonomicon and finished the game with no issues
> * Did a test run in Ropsten 64,-64 without issues
> 
> I think the ABs loading is working fine in the Web build since all the builder scenes, empty scenes and most of genesis plaza are asset bundles and it loads fine

That's quite strange because some "Jump In" buttons from the Explorer menu are working but the Wondermine one instantly freezes

pravusjif:
- Apparently after we are teleported, Kernel sends a couple of messages to the client, to be printed in the chat window, that state to which coords and which realm the client has changed, in the case of Wonder Mine for example those are `Jumping to dg at -29,55...` and `Jumped to -29,55 in realm dg!`.
- At that moment when [`DecaultChatEntry.Populate()` sets its `body.text`](https://github.com/decentraland/unity-renderer/blob/0f8ca6c72b5fe425119afb663376654d9addd13b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/DefaultChatEntry.cs#L83), if the text string has the following character configuration, the build freezes completely:
1. blank space
2. `-`
3. any number
- This super strange breaking bug has to be a Unity (or TMPro) bug.
- After several tests, the only thing that apparently makes the client not crash is waiting for at least 2 frames before setting that text.

sandrade-dcl:
> * Apparently after we are teleported, Kernel sends a couple of messages to the client, to be printed in the chat window, that state to which coords and which realm the client has changed to, in the case of Wonder Mine for example those are `Jumping to dg at -29,55...` and `Jumped to -29,55 in realm dg!`.
> * At that moment when [`DecaultChatEntry.Populate()` sets its `body.text`](https://github.com/decentraland/unity-renderer/blob/0f8ca6c72b5fe425119afb663376654d9addd13b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/DefaultChatEntry.cs#L83), if the text string has the following characters configuration, the build freezes completely:
> 
> 1. blank space
> 2. `-`
> 3. any number
> 
> for example " -2" (as found inside `Jumped to -29,55 in realm dg!`) makes the client freeze forever.
> 
> * This super strange breaking bug has to be a Unity (or TMPro) bug.
> * After several tests, the only thing that apparently makes the client not crash is waiting for at least 2 frames before setting that text.

WTF... Super strange!! And do we know why it is not happening in another situations? for instance, when we put `/goto -29,55` in the chat? What is the difference?

sandrade-dcl (approved):
I have tested the teleport from the Start Menu cards and it is effectively working now 👍  Good work with the investigation of this issue, Pravus!

Anyway, apart from if we decide to go now with this solution, I'm wondering if we should try to find another one a bit "less ugly". For me, to make the `Populate()` method of ALL chat entries wait for X frames each time we receive a message is too strange and it forces us to take it into account for any possible related implementation we need to do in the chat side.

I'm thinking that maybe we could rethink if we really need to receive a chat message from kernel each time we do a teleport... or we could at least make that delay happens in the kernel side. I don't know, for me, making a view (in this case the `DefaultChatEntry`) to wait for a time for populate its info doesn't make sense.

Maybe it would be something to discuss together with the team.

For now, I guess we need to merge this PR asap (because of the fonts issue in Mac), this fix solves the problem and we can create an issue for the future in order to find another possible solution.

pravusjif:
> I'm wondering if we should try to find another one a bit "less ugly". For me, to make the `Populate()` method of ALL chat entries wait for X frames each time we receive a message is too strange and it forces us to take it into account for any possible related implementation we need to do in the chat side.
> 
> I'm thinking that maybe we could rethink if we really need to receive a chat message from kernel each time we do a teleport... or we could at least make that delay happens in the kernel side. I don't know, for me, making a view (in this case the `DefaultChatEntry`) to wait for a time for populate its info doesn't make sense.

I thought about that: filtering and only applying the 3-frames wait for specific messages or something like that, but I fear there could be other cases that we are missing, that could trigger this crash... but if the team wants we could try filtering the fix to only be applied to certain messages anyway.

Thanks for reviewing!

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/DefaultChatEntry.cs
```diff
@@ -57,6 +58,17 @@ private void Awake()
...
+        PopulateTask(chatEntryModel);
+    }
+
+    public async UniTask PopulateTask(ChatEntryModel chatEntryModel)
+    {
+        // Due to a TMPro bug in Unity 2020 LTS we have to wait several frames before setting the body.text to avoid a
+        // client crash. More info at https://github.com/decentraland/unity-renderer/pull/2345#issuecomment-1155753538
+        await UniTask.NextFrame();
+        await UniTask.NextFrame();
+        await UniTask.NextFrame();
```
AjimenezDCL, line 70: I'm really concerned on how reliable this is. We can merbe but we should test it intensively.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/DefaultChatEntry.cs
```diff
@@ -57,6 +58,17 @@ private void Awake()
 
     public override void Populate(ChatEntryModel chatEntryModel)
     {
+        PopulateTask(chatEntryModel);
+    }
+
+    public async UniTask PopulateTask(ChatEntryModel chatEntryModel)
```
AjimenezDCL, line 64: We should implement this with a CancellationToken

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/DefaultChatEntry.cs
pravusjif: Yes, actually with a 2-frame wait the crash is avoided, I put a 3-frame wait just in case, although in all my tests 2 was enough to avoid the crash.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/DefaultChatEntry.cs
pravusjif: I ended up correcting the `PopulateTask()` to be a `UniTaskVoid` instead, since no other method has to wait for it, and since it's usage is a "call and forget" I also added the `Forget()` call.

With all that I think the cancellation token is not needed, since no one will be cancelling that task, same as we do in 2 other places ([here](https://github.com/decentraland/unity-renderer/blob/a222a7a0ac66b438e178419035b505305435aeef/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WorldChatWindowHUD/PublicChatChannelController.cs#L168) and [here](https://github.com/decentraland/unity-renderer/blob/a222a7a0ac66b438e178419035b505305435aeef/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/PrivateChatWindow/PrivateChatWindowController.cs#L160)) using `UniTaskVoid` to await for `UniTask.NextFrame()` inside.

What do you think?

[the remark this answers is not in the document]
pravusjif: totally, done.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/Tests/DefaultChatEntryShould.cs
```diff
@@ -22,8 +24,8 @@ public void TearDown()
         Object.Destroy(canvas.gameObject);
     }
     
-    [Test]
-    public void PopulateSystemChat()
+    [UnityTest]
+    public IEnumerator PopulateSystemChat()
```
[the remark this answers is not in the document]
pravusjif: Done

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/DefaultChatEntry.cs
pravusjif: In the end I had to keep the `PopulateTask()` as an awaitable `UniTask` to be able to await it in the tests.

Should I add the `CancellationToken` even though no one will ever cancel that "fire and forget" method?

AjimenezDCL: yes, we should do the cancellation in the `OnDestroy()` at the very least

unity-renderer/Assets/Scripts/MainScripts/DCL/WebInterface/Interface.cs
```diff
@@ -691,6 +691,7 @@ public class TimeReportPayload
     [DllImport("__Internal")] public static extern void MessageFromEngine(string type, string message);
     [DllImport("__Internal")] public static extern string GetGraphicCard();
     [DllImport("__Internal")] public static extern bool CheckURLParam(string targetParam);
+    [DllImport("__Internal")] public static extern string GetURLParam(string targetParam);
```
AjimenezDCL, line 694: Also, why is this needed.

AjimenezDCL (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/WebInterface/Interface.cs
AjimenezDCL: Cool, maybe we should state this information somehwere in the code itself. To make it clear than in normal production environment we dont care about URLs, we just do what kernel says.

unity-renderer/Assets/Plugins/JSFunctions.jslib
```diff
@@ -36,6 +36,21 @@ mergeInto(LibraryManager.library, {
   },
   CheckURLParam: function(targetParam) {    
     return window.DCL.queryString.has(Pointer_stringify(targetParam))
+  },
+  GetURLParam: function(targetParam) {
+    const stringifiedTargetParam = Pointer_stringify(targetParam)
+    
+    if(window.DCL.queryString.has(stringifiedTargetParam)) {
```
menduz, line 43: what is DCL.queryString?

Can we use `new URL(document.location)` instead?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/ExploreV2/Tests/EventCardComponentViewTests.cs
```diff
@@ -342,15 +344,19 @@ public void CloseModalCorrectly()
...
     {
         // Arrange
         eventCardModalComponent.Show();
 
         // Act
         eventCardModalComponent.OnCloseActionTriggered(new DCLAction_Trigger());
 
+        yield return null;
+        yield return null;
+        yield return null;
```
menduz, line 358: Why is this necessary? please add a comment or delete

unity-renderer/Assets/Scripts/MainScripts/DCL/WebInterface/Interface.cs
menduz: May I suggest to send this information via KernelConfig instead?

unity-renderer/Assets/Plugins/JSFunctions.jslib
menduz: May I suggest to send this information via KernelConfig instead? This file should not be touched actually

menduz (changes requested):
Please address kernel integration comments.

Also, there is a `TMP_SDF SSD.shader` file, should that be commited? If yes, can we change the name to anything but "tmp"?

unity-renderer/Assets/Scripts/MainScripts/DCL/WebInterface/Interface.cs
pravusjif: I removed them in [this commit](https://github.com/decentraland/unity-renderer/pull/2345/commits/da6dc59b49b8fdbeb15cf789a097d3df67f4eef2) and [created this issue](https://github.com/decentraland/unity-renderer/issues/2519) for the future

unity-renderer/Assets/Plugins/JSFunctions.jslib
pravusjif: I removed them in [this commit](https://github.com/decentraland/unity-renderer/pull/2345/commits/da6dc59b49b8fdbeb15cf789a097d3df67f4eef2) and [created this issue](https://github.com/decentraland/unity-renderer/issues/2519) for the future

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/ExploreV2/Tests/EventCardComponentViewTests.cs
pravusjif: Added a comment in [this commit](https://github.com/decentraland/unity-renderer/pull/2345/commits/19b1c45cc2a793728c2587b6c6774ad28f16dcdc)

pravusjif:
> Please address kernel integration comments.
> 
> Also, there is a `TMP_SDF SSD.shader` file, should that be commited? If yes, can we change the name to anything but "tmp"?

@menduz In this case TMP is from TextMeshPro, not from "temp"

menduz (commented):
Thanks for the changes, please review the TMP shader.

menduz (approved):

PR #2370, opened 2022-05-30
Fix: removed ff for new skybox

### LINKED ISSUE #2309: New skybox only works if feature flags domain is up
New skybox only works if feature flags domain is up. I'd recommend to stop using the feature flag, delete it and convert the new skybox to the default setting.
Otherwise it simply looks like this:
Screen Shot 2022-05-13 at 18 16 51

davidejensen:
## What does this PR change?

Fix #2309 
Removes feature flag for new skybox

...

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=fix/ff-skybox
2. Verify that the new skybox is visible
3. Verify that the skybox settings work as expected

## Our Code Review Standards


unity-renderer/Assets/Rendering/ProceduralSkybox/ToolProceduralSkybox/Scripts/SkyboxCamera.cs
```diff
@@ -33,6 +33,9 @@ public SkyboxCamera()
 
         public void AssignTargetCamera(Transform mainCam)
         {
+            if (mainCam == null)
```
AjimenezDCL, line 36: Is this needed?

AjimenezDCL (changes requested):
Isnt this just removing the whole SkyBox? the point of the task is to make it default.

davidejensen:
> Isnt this just removing the whole SkyBox? the point of the task is to make it default.

This removes the ff, we still have the plugin of the skybox subbed in a previous line   
`pluginSystem.Register<SkyboxController>(() => new SkyboxController());`
and it's already set as default.

The null check is needed due to the initialization of the plugin being called before the presence of the actual camera and with a sequent getcomponent on it.

PR #2371, opened 2022-05-30
feat: proto automatic generation

Suduck:
## What does this PR change?

This PR introduces the ability to download and update the protocol buffers for every SO.

## How to test the changes?

We should test this in mac, linux and windows

1. Download the branch feat/proto-generation
2. Compile the project
3. Test that everything work as expected 

## Our Code Review Standards


unity-renderer/Assets/DCLPlugins/ECS7/ProtocolBuffers/Editor/compiledVersion.gen.txt
```diff
@@ -0,0 +1 @@
+1.0.0-20220525130536.commit-7f8e347
```
kuruk-mm, line 1: This should not be ignored or in `Generated` folder?

unity-renderer/Assets/DCLPlugins/ECS7/ProtocolBuffers/Editor/ProtobufEditor.cs
```diff
@@ -0,0 +1,337 @@
...
+        public static void DownloadProtobuffExecutable()
+        {
+            // Download package
+            string machine  = null;
+            string executableName = "protoc";
+#if UNITY_EDITOR_WIN
+            machine = "win64";
+            executableName = "protoc.exe";
+#endif
+#if UNITY_EDITOR_OSX
```
kuruk-mm, line 194: use `#elif`

unity-renderer/Assets/DCLPlugins/ECS7/ProtocolBuffers/Editor/ProtobufEditor.cs
```diff
@@ -0,0 +1,337 @@
...
+            finally
+            {
+                // We removed everything has has been created and it is not usefull anymore
+                File.Delete(zipProtoFileName);
+                if (Directory.Exists(destPackage))
+                    Directory.Delete(destPackage, true);
+            }
+        }
+
+        private static void Unzip(string name, string path)
```
kuruk-mm, line 243: Can we use something like https://docs.microsoft.com/en-us/dotnet/api/system.io.compression.zipfile.extracttodirectory?view=net-6.0

instead of a process execution?
The file that are you downloading is a `zip`, not a `tar`.

When you download the package in `.tar.gz`, the function should be `Untar`.

Here is an example of using `Untar` with Pure C#

https://gist.github.com/ForeverZer0/a2cd292bd2f3b5e114956c00bb6e872b

Try to avoid using process execution because is difficult to make it work in different OS

kuruk-mm (changes requested):

unity-renderer/Assets/DCLPlugins/ECS7/ProtocolBuffers/Editor/compiledVersion.gen.txt
Suduck: No, this is the compiled version, we need to push this to the repo. This will be the the file that the system will look to compile new protos

unity-renderer/Assets/DCLPlugins/ECS7/ProtocolBuffers/Editor/ProtobufEditor.cs
Suduck: No, we can't use the zip file since the library (zip file) is not included in Unity and we should avoid including libraries that won't be included in the build.

The untar function hasn't work

.gitignore
```diff
@@ -20,6 +20,11 @@ Assets/InitTestScene*.unity.meta
 ehthumbs.db
 Thumbs.db
 
+# ===================================== #
+# Proto generation system               #
+# ===================================== #
+#Generated/
+bin/
```
AjimenezDCL, line 27: why is this in the .meta file of the test scene?

AjimenezDCL (approved):
Looks fine! leaving a small comment

.gitignore
Suduck: I don't understand, this is in the `.gitignore`

kuruk-mm (approved):

pbosio (approved):

PR #2377, opened 2022-05-31
feat: add cylinder shape to the new ECSComponent

Suduck:
## What does this PR change?

This PR adds the new component cylinder shape to the new ECS

## How to test the changes?

Right now there is no way to test changes, we are releasing soon an scene where we will be able to see changes

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Models/DCLResources/PrimitiveMeshModel.cs
```diff
@@ -21,6 +22,8 @@ public PrimitiveMeshModel(Type type)
     }
 
     public Type type;
+    public float radiusTop;
+    public float radiusBottom;
```
sandrade-dcl, line 26: Only a doubt: Does it make sense that ALL primitive meshes have a top/bottom radius? I'm thinking, for example, in a sphere... it would be a primitive shape but a top/bottom radius wouldn't have sense, no?

sandrade-dcl (approved):
LGTM! I only left you a little doubt :)

unity-renderer/Assets/Scripts/MainScripts/DCL/Models/DCLResources/PrimitiveMeshModel.cs
Suduck: I has changed it to be more general. I don't like neither of both approaches , but I think they will do the job

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/CylinderShape/ECSCylinderShapeComponentHandler.cs
```diff
@@ -0,0 +1,64 @@
...
+        public void OnComponentModelUpdated(IParcelScene scene, IDCLEntity entity, PBCylinderShape model)
+        {
+            Mesh generatedMesh = null;
+            if (primitiveMeshPromisePrimitive != null)
+                AssetPromiseKeeper_PrimitiveMesh.i.Forget(primitiveMeshPromisePrimitive);
+
+            PrimitiveMeshModel primitiveMeshModelModel = new PrimitiveMeshModel(PrimitiveMeshModel.Type.Cylinder);
+            primitiveMeshPromisePrimitive = new AssetPromise_PrimitiveMesh(primitiveMeshModelModel);
+            primitiveMeshPromisePrimitive.model.radiusBottom = model.RadiusBottom;
+            primitiveMeshPromisePrimitive.model.radiusTop = model.RadiusTop;
```
pbosio, line 34: why the modification of the inner `model` field of the promise?
shouldn't those modification be applied to `primitiveMeshModelModel` instead?

pbosio (approved):

PR #2380, opened 2022-05-31
Fix/nearby visible

lorux0:
## What does this PR change?

Fixes an issue that happened whenever you locked the mouse while a private chat window was opened, the public channel window was getting on top of it.
Private & public chat windows only gets focus when clicking over them.
Public chat window is opened when experiences window is closed.

## How to test the changes?

1. Open a private chat window
2. Lock the mouse
3. Navigate through different chat windows, chat list and nearby channel.
4. Toggle on & off all the taskbar buttons

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/PrivateChatWindow/PrivateChatWindowComponentView.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/PrivateChatWindow/PrivateChatWindowComponentView.cs
@@ -3,9 +3,10 @@ using System.Collections;
 using SocialBar.UserThumbnail;
 using TMPro;
 using UnityEngine;
+using UnityEngine.EventSystems;
 using UnityEngine.UI;
 
-public class PrivateChatWindowComponentView : BaseComponentView, IPrivateChatComponentView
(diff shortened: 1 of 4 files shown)


lorux0:
> The code looks good.
> 
> Testing the app I think there are a couple of strange behaviours:
> 
> * The chat taskbar toggle button is only working for minimizing (actually making chat background transparent) and not for maximizing the chat window (if it's already minimized I click the orange chat button and nothing changes). Using the Enter key works fine for toggling the chat window transparency, probably clicking the taskbar chat button should work as the enter key right?

Fixes should be ready. The toggle is now on/off depending on the transparent state of the window.

> * If I opened a private chat, minimized the chat (the background turns transparent but the chat messages are still there) and then I click in the chat taskbar button: it changes the chat window messages to the #nearby messages

Whenever you close the private chat, the #nearby channel should be opened. So if you click the chat toggle it will opaque the #nearby channel

lorux0:
> The behaviour changed now:
> 
> * The chat is minimized randomly (when it happens in this video I'm not clicking nor doing anything, it just closes):
> 
>  Screen.Recording.2022-06-01.at.13.24.01.mp4 
> * The enter key toggling stopped working for me

this is an expected behaviour. The window transparents when the mouse is off the area for a couple of seconds, only when the input field is not selected. We suppose at that point that you are not using the chat so we avoid adding visual noise to the world

PR #2382, opened 2022-05-31
Chore/social tests

lorux0:
## What does this PR change?

`Links to #2321 `

Friends HUD Controller refactor into dependency injection and static usage removal.
Friends HUD Controller test coverage.

## How to test the changes?

Check that all features related to friends are working.

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/HUDController.cs
```diff
@@ -269,7 +269,7 @@ private void AllUIHiddenOnOnChange(bool current, bool previous)
 
                     if (friendsHud != null)
                     {
-                        friendsHud.Initialize(FriendsController.i, UserProfile.GetOwnUserProfile(), socialAnalytics);
+                        friendsHud.Initialize(FriendsHUDComponentView.Create());
```
AjimenezDCL, line 272: We usually solve the mocking issue here instead of using DI, just having a virtual method `CreateView` in the Controller that can be changed using `Substitute.ForParts<>()`.

Not saying I dont like this approach, but we should keep a common approach.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/HUDFactory.cs
```diff
@@ -51,7 +51,12 @@ public virtual IHUD CreateHUD(HUDElementID hudElementId)
                 hudElement = new TermsOfServiceHUDController();
                 break;
             case HUDElementID.FRIENDS:
-                hudElement = new FriendsHUDController(DataStore.i);
+                hudElement = new FriendsHUDController(DataStore.i, new WebInterfaceFriendsController(FriendsController.i),
+                    new UserProfileWebInterfaceBridge(),
+                    new SocialAnalytics(
+                        Environment.i.platform.serviceProviders.analytics,
+                        new UserProfileWebInterfaceBridge()),
+                    Environment.i.serviceLocator.Get<IFriendsNotificationService>());
```
AjimenezDCL, line 59: I wouldnt make this a service, we should leave the services to really low level stuff.
I would implement this as a plugin and make it DataDriven.

You could have a DataStore with notifications by friend that is filled by a bridge (living inside this new plugin). 
The `Mark as Read` process would be implemented that way as well. We could have another collection with the `Seen notifications` (maybe carrying the timestamp of the last seen message) and let the plugin do the reconciliation once kernel recives the update of the seen status.

In fact the current architecture with the NotificationScriptableObjects is already pointing to the above design. I would also take the opportunity to remove the usage of SO since we are deprecating them.

AjimenezDCL (changes requested):
Good job! I just have some concerns with the new service created.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/HUDFactory.cs
lorux0: I think its correct to call it an application service which provides a solution for the application domain. I dont like the idea of moving it to a plugin since its part of the core of the application. We always need to display the notifications.
Moving the data to the DataStore will break the functionality of `NotificationBadge` since is the one that renders the friends notifications. I think we should move it, but it gets out of the scope of this refactor.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/HUDController.cs
AjimenezDCL: Null checking is not needed (or is not a need raised from having a CreateView method) you could pass a null view through dependency injection as well.

I agree I rather have DI though.

lorux0:
@AjimenezDCL i also added the test coverage for `FriendsHUDComponentView`

AjimenezDCL (approved):
looks great!

PR #2389, opened 2022-06-02
fix: discard pixels below Y plane

AjimenezDCL:
Due to some camera tricks for the new floor, the objects hidden in the negative Y of the world were being displayed. This fixes it by discarding the pixels directly.

Test below the floor is rendered. Also take a look at the world, GLTF transition effect and such.

## Our Code Review Standards


pravusjif (approved):
tested in ropsten and works great!

PR #2391, opened 2022-06-02
fix: friend request notification count, direct messages spinner for guests & chat window resizing on preview mode

lorux0:
## What does this PR change?

- private & public chat windows are resized to avoid overlapping with scene UIs
- hide direct chats loading spinner when the user is a guest
- received friend request notifications count at taskbar

## How to test the changes?

1. Authenticate as a guest, open the chat list window, search for a friend. No UI components should be overlapped.
2. Go to ice poker scenes, check that the chat in transparent mode does not overlap with their UI.
3. Receive a friend request, the notification count should be updated correctly in the taskbar.

## Our Code Review Standards


sandrade-dcl (approved):
Good work Nico! It's working fine for me 👍

PR #2419, opened 2022-06-06
fix: performance analytics id instead of position

Kinerius:
## What does this PR change?

Changed the way the scene memory scoring is presented to the analytics.
Also changed the name of the feature flag.

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Debugging/Performance/PerformanceMetricsController.cs
```diff
@@ -99,7 +99,7 @@ private void Report(string encodedSamples)
                 if (coords.x == 0 && coords.y == 0)
```
pravusjif, line 99: Instead of the coordiantes, let's check with `parcelScene.isPersistent` to know if it's the global scene (as we do in [this other place](https://github.com/decentraland/unity-renderer/blob/7415d9f6526d1afa46c266b07af667a072f27b22/unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/Utils/PositionUtils/WorldStateUtils.cs#L16)

pravusjif (approved):

PR #2423, opened 2022-06-07
feat: ecs components serialization implementation

Suduck:
## What does this PR change?

This PR implements the implementation of the serialization for the components. It also organizes better the current structure of the ECS components.

Right now we have a ComponentRegister that will register the handler and the serialization of the component 

## How to test the changes?

This can't be tested

## Our Code Review Standards


unity-renderer/Assets/DCLPlugins/ECS7/ECS7Plugin.cs
```diff
@@ -1,24 +1,28 @@
...
+        {
+            crdtWriteSystem = new ComponentCrdtWriteSystem(Environment.i.platform.updateEventHandler, Environment.i.world.state);
+            componentWriter = new ECSComponentWriter(crdtWriteSystem.WriteMessage);
 
-    public void Dispose()
-    {
-        componentsPlugin.Dispose();
-        crdtWriteSystem.Dispose();
-        componentWriter.Dispose();
+            componentsPlugin = new ECS7ComponentsPlugin(DataStore.i.ecs7.componentsFactory, componentWriter);
```
AjimenezDCL, line 18: One of the forbidden rules in our Plugins System is that a plugin cannot contain/reference other plugin.

I know this is different because `ECS7ComponentsPlugin`  is not an `IPlugin` but we should find a better name so it's not confusing.

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/AudioSource/AudioSourceRegister.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using DCL.ECSRuntime;
+using DCL.SettingsCommon;
+
+namespace DCL.ECSComponents
+{
+    public class AudioSourceRegister : IDisposable
```
AjimenezDCL, line 7: I see all the registers are basically identical.

Maybe we can have a `DefaultRegister<THandler>` to avoid boilerplate?
We could still create a custom register if any of the components need specific stuff.

AjimenezDCL (approved):
Approving with 2 suggestions. Good job!

unity-renderer/Assets/DCLPlugins/ECS7/ECS7Plugin.cs
Suduck: I will call it ECS7ComponentComposer

AjimenezDCL: Love it!

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/AudioSource/AudioSourceRegister.cs
Suduck: I thought to make a single code to do it, but this way we can encapsulate assembly. Since each handler has different parameters in the constructor.

I can go for that if you want!

pbosio (approved):

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/SphereShape/Handler/ECSSphereShapeHandler.cs
```diff
@@ -33,6 +33,8 @@ public void OnComponentModelUpdated(IParcelScene scene, IDCLEntity entity, PBSph
                 DisposeMesh(scene);
                 generatedMesh = shape.mesh;
                 GenerateRenderer(generatedMesh, scene, entity, model);
+                entity.OnShapeLoaded?.Invoke(entity);
+                entity.OnShapeUpdated?.Invoke(entity);
```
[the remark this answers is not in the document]
Suduck: I will remove them for now and I will open a discussion about them

PR #2440, opened 2022-06-09
chore: reduce allocations from avatar system

Kinerius:
## Why?

When a lot of avatars spawn, the GarbageCollector causes hiccups very frequently, this is caused by very large allocations every time an avatar is spawned. 

## What changed?

Reduced the allocations of copying arrays by using Unity's NativeArray which provides allocation-free management of arrays.

## Results

Performance Score incremented from `68,78` to `73,87`
Allocations over 30 seconds lowered from `544.585,61` to `509.632,536`

(results might differ because genesis plaza is chaotic and people move, the test started when the first avatar was about to spawn)

For more information, read the comments below.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Rendering/Utils/GPUSkinning/SimpleGPUSkinning.cs
+++ b/unity-renderer/Assets/Rendering/Utils/GPUSkinning/SimpleGPUSkinning.cs
@@ -1,5 +1,7 @@
 using System.Linq;
+using Unity.Collections;
 using UnityEngine;
+using UnityEngine.Profiling;
 using Object = UnityEngine.Object;
 
 namespace GPUSkinning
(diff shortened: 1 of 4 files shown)


Kinerius:
Results of the first commit: 

image

Kinerius:
Regarding the second commit, allocations were reduced by 100% but the time is much higher. After some research, reading and writing from `NativeArray` in the editor seem slow because of thread-safe checks that are completely gone from compiled builds. I'll do a performance test once I finish reducing allocations from the avatar system,

image

Kinerius:
The third commit removed allocations of new arrays by using NativeArrays of UninitializedMemory, I couldn't get rid of the allocation of a list of Materials because they have different usages.

For getting these screenshots I always try to get the most underperforming frames to get the biggest numbers, in this case, both frames are the worst of both profile sessions.

Old:
Screenshot 2022-06-10 152852

New:
Screenshot 2022-06-10 153342

Kinerius:
In editor performance:

Old
```
PerformanceMeter - Data report step 2 - Processed values:
 * PERFORMANCE SCORE (0-100) -> 68,78
 * average FPS -> 32,75605
 * highest FPS -> 48
 * lowest FPS -> 14
 * 50 percentile (median) FPS -> 33
 * 95 percentile FPS -> 45
 * total hiccups (>0,05ms frames) -> 162 (17,03% of frames were hiccups)
 * total hiccups time (seconds) -> 11,4532
 * total frames -> 951
 * total frames time (seconds) -> 30,00033
 * lowest allocations (kb) -> 23,573
 * highest allocations (kb) -> 15786,35
 * average allocations (kb) -> 572,645
 * total allocations (kb) -> 544585,61
 ```

New
``` 
PerformanceMeter - Data report step 2 - Processed values:
 * PERFORMANCE SCORE (0-100) -> 73,87
 * average FPS -> 37,76638
 * highest FPS -> 54
 * lowest FPS -> 14
 * 50 percentile (median) FPS -> 37
 * 95 percentile FPS -> 53
 * total hiccups (>0,05ms frames) -> 160 (15,19% of frames were hiccups)
 * total hiccups time (seconds) -> 10,78473
 * total frames -> 1053
 * total frames time (seconds) -> 30,02126
 * lowest allocations (kb) -> 50,342
 * highest allocations (kb) -> 16852,97
 * average allocations (kb) -> 483,981
 * total allocations (kb) -> 509632,536
```

TLDR: 50MB less allocations per 1000 frames and better performance score 🚀 

I expected performance to drop a bit since we are using NativeArrays that are slower on editor but it seems it's not having much impact.

File: unity-renderer/Assets/Rendering/Utils/GPUSkinning/SimpleGPUSkinning.cs [hunk 1]
```diff
@@ -21,28 +23,59 @@ public static void EncodeBindPosesIntoMesh(SkinnedMeshRenderer skr)
         {
             Mesh sharedMesh = skr.sharedMesh;
             int vertexCount = sharedMesh.vertexCount;
-            Vector4[] bone01data = new Vector4[vertexCount];
-            Vector4[] bone23data = new Vector4[vertexCount];
-
-            BoneWeight[] boneWeights = sharedMesh.boneWeights;
+            
+            NativeArray<Vector4> bone01data = new NativeArray<Vector4>(vertexCount, Allocator.Temp, NativeArrayOptions.UninitializedMemory);
```
lorux0, line 27: you are not disposing the arrays.. could this be a reason of memory leaks?

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarMeshCombiner/AvatarMeshCombinerUtils.cs
```diff
@@ -59,22 +60,42 @@ public static BoneWeight[] ComputeBoneWeights( List<CombineLayer> layers )
...
                     boneWeightArrays.Add(boneWeights);
-                    resultSize += boneWeights.Length;
+
+                    totalVertexes += bonesPerVertex.Length;
+                    totalBones += boneWeights.Length;
                 }
             }
 
-            BoneWeight[] result = new BoneWeight[resultSize];
+            NativeArray<byte> finalBpV = new NativeArray<byte>(totalVertexes, Allocator.Temp, NativeArrayOptions.UninitializedMemory);
```
lorux0, line 74: you are not disposing the arrays.. could this be a reason of memory leaks?

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarMeshCombiner/Tests/AvatarMeshCombinerUtilsCan.cs
```diff
@@ -121,18 +121,18 @@ SkinnedMeshRenderer CreateRendererWithWeights()
             SkinnedMeshRenderer renderer = DCL.Helpers.SkinnedMeshRenderer.Create(mat);
 
             int vertexCount = renderer.sharedMesh.vertexCount;
-            var boneWeights = new BoneWeight[vertexCount];
 
-            for (int i = 0; i < boneWeights.Length; i++)
+            var nativeBonesPerVertex = new NativeArray<byte>(vertexCount, Allocator.Temp, NativeArrayOptions.UninitializedMemory);
```
lorux0, line 125: you are not disposing the arrays.. could this be a reason of memory leaks?

unity-renderer/Assets/Rendering/Utils/GPUSkinning/SimpleGPUSkinning.cs
```diff
@@ -21,28 +23,59 @@ public static void EncodeBindPosesIntoMesh(SkinnedMeshRenderer skr)
...
+                byte boneCount = bonesPerVertex[i];
+                if (boneCount == 0)
+                {
+                    bone01data[i] = new Vector4(0, 0, 0, 0);
+                    bone23data[i] = new Vector4(0, 0, 0, 0);
+                } 
+                else if (boneCount == 1)
+                {
+                    var bw0 = boneWeightsNative[boneIndex];
+                    bone01data[i] = new Vector4(bw0.boneIndex, bw0.weight, 0, 0);
```
menduz, line 44: Isn't `new Vector` allocating memory? and the assignment copying it? If you are already initializing the memory with zeros, then something like this may be better:

```suggestion
var vec = bone01data[i];
vec.x = bw0.boneIndex;
vec.y = bw0.weight;
```

It is not the prettiest, but in search of speed some stylish things can be put in second place to reduce operations

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarMeshCombiner/AvatarMeshCombiner.cs
```diff
@@ -124,17 +127,14 @@ public static Output CombineSkinnedMeshes(Matrix4x4[] bindPoses, Transform[] bon
 
             finalMesh.bindposes = bindPoses;
 
-            var boneWeights = AvatarMeshCombinerUtils.ComputeBoneWeights( layers );
-            finalMesh.boneWeights = boneWeights;
+            (NativeArray<byte> bonesPerVertex, NativeArray<BoneWeight1> weights) = AvatarMeshCombinerUtils.ComputeBoneWeights(layers);
+            finalMesh.SetBoneWeights(bonesPerVertex, weights);
```
menduz, line 131: Wouldn't be better to pass `finalMesh` to `AvatarMeshCombinerUtils.ComputeBoneWeights` instead of returning the couple?

unity-renderer/Assets/Rendering/Utils/GPUSkinning/SimpleGPUSkinning.cs
menduz: Doesn't it require a `using` to perform the disposal properly?

Kinerius: I did a test for the sake of it, I've changed the allocation of the native array to Persistent and didn't dispose them, after a few frames Unity actually tells you that is leaking 
image

So calling dispose or not with Allocation.Temp is the same, unity disposes it when the reference count is zero after a few frames.

Kinerius: I will add the Disposes to avoid future confusion.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarMeshCombiner/AvatarMeshCombiner.cs
Kinerius: It's intended for testing that this method doesn't have a side effect (assigning the bone weights to a mesh), but I've separated this method into 2 to avoid the usage of a tuple.

lorux0 (approved):

PR #2453, opened 2022-06-13
Fix: render all empty parcels with same color

### LINKED ISSUE #2416: Floor tiles change color when far away
The floor tiles have a darker color when they are far away, and then changes to their correct color when the user moves 
closer to them.

.

davidejensen:
## What does this PR change?

Fix #2416 
Increases the range of the reflection probe to include all nearby parcels in the dynamic baking 

...

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=fix/floor-color
2. Go to an isolated area like -66,-86
3. Verify that the colour of the ground in the nearby parcels is always the same

## Our Code Review Standards


sandrade-dcl (commented):
I'm noticing these black horizontal lines... is it normal?

image

sandrade-dcl (approved):
Anyway, the color is now always the same 👍

davidejensen:
> I'm noticing these black horizontal lines... is it normal?
> 
> image

Yeah this is happening in prod as well, it's not related to this changes

PR #2463, opened 2022-06-13
fix: sign up UI

RominaMarchetti:
## What does this PR change?

This PR fixes the missing elliptical mask of the avatar snapshot and adds a texture to the scrollbar sliding area in order to make it stand out.

Mask before vs. after
Screen Shot 2022-04-22 at 16 55 54
Screen Shot 2022-06-14 at 16 21 20

Scrollbar sliding area before vs. after
Screen Shot 2022-06-14 at 16 21 02
Screen Shot 2022-06-14 at 16 19 53

## How to test the changes?

1. Create a new account from the following link ( to make it easier, enter in incognito mode with Chrome and select "Play as Guest"): https://play.decentraland.zone/?renderer-branch=fix/SignupHUD-UI
2. After creating your avatar you should see that your profile picture is masked inside an ellipse.
3. In the step Terms of Service (...) you should see the extension of the sliding area, not just the handle.


davidejensen (approved):
Works as expected!

PR #2468, opened 2022-06-14
style: close button

RominaMarchetti:
## What does this PR change?

The close button UI was not aligned with the UI guidelines not in the teleport/goto panels nor in the realm selector.
Animation on mouse interactions are now added too.

Teleport & Go to panel before vs. after
Screen Shot 2022-06-14 at 10 51 05

Realm selector panel before vs. after
Screen Shot 2022-06-14 at 10 50 43

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=style/CloseButton
2. Go to the scenes column in the genesis plaza and click any of the scene's card jump in button. You should see the new close button (squared + animation on mouse hover & click).
3. Then go to the main menu and click the realm selector. You should see the new close button in the realm selector panel  (squared + animation on mouse hover & click).


davidejensen (approved):
nice! :)

PR #2473, opened 2022-06-14
 fix: mac (12.3+) local web building

pravusjif:
### WHY
- Emscripten build toolchain for WebGL builds on Unity 2020.3 depends on Python 2.7 or later. [WebGL builds on Unity 2021.2 and later require Python 3].
- MacOS doesn't include Python anymore since macOS 12.3 (and removes the previously-existent version). So making a WebGL build locally in Mac 12.3+ always fails because Python can't be found.

#### Solution for building locally for WebGL on macOS:
1. Download Python 3+ at https://www.python.org/downloads/release/python-3105/
2. Have a PreBuildProcessing script in place (added in this PR) to re-map Python path when building. 

Resources:
- https://forum.unity.com/threads/case-1412113-builderror-osx-12-3-and-unity-2020-3-constant-build-errors.1255419/#post-7993017
- https://answers.unity.com/questions/1893841/unity-2020328f1-webgl-build-failed-on-macos-monter.html

### WHAT
Added PreBuildProcessing script needed for building in MacOS 12.3 or later

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,13 +97,27 @@ When the steps are followed, you will be able to run the local Unity build by go
 
 1. Make sure you have the proper Unity version up and running
 2. Make sure you have [Kernel repository](https://github.com/decentraland/kernel) cloned.
-3. Make sure you are running kernel through `make watch` command in the cloned repo directory.
-4. Make sure you have the [explorer website repository](https://github.com/decentraland/explorer-website) cloned.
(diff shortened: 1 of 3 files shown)


pravusjif:
> It looks good! CI is not using Mac but do you think we need to do anything else?

Not really since the CI uses Linux, this fix only affects people building the WebGL locally using MacOS.

PR #2476, opened 2022-06-15
feat: add CRDT RPC Service

pbosio:
https://github.com/decentraland/sdk/issues/321
implement RPC for CRDT messages.

context:
* CRDT === ECS7 scenes messages.
* kernel <-> transport (websocket or nativeBridge for web) <-> rpc server <-> rpc service
* CRDT service consist in 2 functions:
** `sendCrdt` which will receive incoming CRDT messages from scene
** `crdtNotificationStream` which is an endless stream that consumes a queue of batched messages to send back to the scenes


unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/KernelCommunication/RPC/GeneratedCode/RendererProtocolService.gen.cs
```diff
@@ -0,0 +1,44 @@
...
+        var result = new ServerModuleDefinition<Context>();
+      
+        result.definition.Add("SendCrdt", async (payload, context) => { var res = await sendCrdt(CRDTManyMessages.Parser.ParseFrom(payload), context); return res?.ToByteString(); });
+        result.streamDefinition.Add("CrdtNotificationStream", (payload, context) => { return new ProtocolHelpers.StreamEnumerator<CRDTManyMessages>(crdtNotificationStream(CRDTStreamRequest.Parser.ParseFrom(payload), context)); });
+
+        port.RegisterModule(ServiceName, (port) => UniTask.FromResult(result));
+    }
+}
+
+public abstract class PingPongService<Context>
```
kuruk-mm, line 30: Do you want to keep with this test Service? 🤔

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/KernelCommunication/RPC/Transports/WebSocketTransport/WebSocketTransport.cs
```diff
@@ -0,0 +1,80 @@
+using System;
+using rpc_csharp.transport;
+
+namespace RPC.Transports
+{
+    // TODO: fix issues when websocket reconnect!
```
kuruk-mm, line 6: Which issues are we having? The Unity-side doesn't care about reconnecting. The Desktop connecting flow it's like:

- Kernel tries to connect to Unity
- Kernel tries to connect to Unity
- Kernel tries to connect to Unity
- Unity starts listening
- Kernel connects

kuruk-mm (approved):
Great work! I will test this on Desktop Client

kuruk-mm:
explorer-desktop version: https://github.com/decentraland/explorer-desktop/pull/347

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/KernelCommunication/RPC/Transports/WebSocketTransport/WebSocketTransport.cs
pbosio: I can't recall... I think there were actually no issue there... but I do recall you telling me that you were going to refactor the websocket service 🙄😝

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/KernelCommunication/RPC/GeneratedCode/RendererProtocolService.gen.cs
pbosio: it should be removed after merging this PR, right? https://github.com/decentraland/unity-renderer/pull/2577

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/KernelCommunication/RPC/GeneratedCode/RendererProtocolService.gen.cs
```diff
@@ -0,0 +1,44 @@
...
+using Cysharp.Threading.Tasks;
+using Google.Protobuf;
+using rpc_csharp.protocol;
+using rpc_csharp;
+
+public abstract class CRDTService<Context>
+{
+    public const string ServiceName = "CRDTService";
+
+    public delegate UniTask<CRDTResponse> SendCrdt(CRDTManyMessages request, Context context);
```
AjimenezDCL, line 15: I'd assume we need a Cancellation Token for this.

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/KernelCommunication/RPC/GeneratedCode/RendererProtocolService.gen.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    public delegate IEnumerator<CRDTManyMessages> CrdtNotificationStream(CRDTStreamRequest request, Context context);
```
AjimenezDCL, line 17: Why some are tasks and other corroutines?

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/KernelCommunication/RPC/ServerBuilder/RPCServerBuilder.cs
```diff
@@ -0,0 +1,76 @@
...
+                   .Build();
+        }
+
+        public static RPCServerBuilder<RPCContext> BuildServer(RPCContext context, ITransport transport)
+        {
+            return new RPCServerBuilder<RPCContext>(transport, context);
+        }
+    }
+
+    public class RPCServerBuilder<T>
```
AjimenezDCL, line 42: Don't we need to narrow `T` into `RPCContext` or derived?

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/KernelCommunication/RPC/Services/CRDTService/CRDTService.cs
```diff
@@ -0,0 +1,83 @@
...
+using DCL;
+using DCL.CRDT;
+using Google.Protobuf;
+using KernelCommunication;
+using rpc_csharp;
+using UnityEngine;
+
+namespace RPC.Services
+{
+    public static class CRDTServiceImpl
```
AjimenezDCL, line 13: This shouldn't be static and it should be moved into the Service Locator

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/KernelCommunication/RPC/Services/CRDTService/CRDTService.cs
```diff
@@ -0,0 +1,83 @@
...
+        public static void RegisterService(RpcServerPort<RPCContext> port)
+        {
+            CRDTService<RPCContext>.RegisterService(
+                port,
+                sendCrdt: OnCRDTReceived,
+                crdtNotificationStream: CRDTNotificationStream
+            );
+        }
+
+        private static UniTask<CRDTResponse> OnCRDTReceived(CRDTManyMessages messages, RPCContext context)
```
AjimenezDCL, line 29: We need a cancellation token here, take a look at other usages (Avatar.cs for example) to implement it properly using `OperationCanceledException`, otherwise a cancelled task will keep runing until the end of player loop.

AjimenezDCL (changes requested):

unity-renderer/Assets/DCLPlugins/ECS7/ComponentCrdtWriteSystem/ComponentCrdtWriteSystem.cs
```diff
@@ -127,7 +127,7 @@ private bool DispatchSceneMessages(string sceneId, Queue<CRDTMessage> messages)
 
     private void DispatchBinaryMessage(string sceneId, MemoryStream stream)
     {
-        WebInterface.SendBinaryMessage(sceneId, stream.ToArray());
+        RPCGlobalContext.context.crdtContext.notifications.Enqueue((sceneId, stream.ToArray()));
```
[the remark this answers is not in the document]
pbosio: Not sure if ServiceLocator is the proper place, since it's just data.
Maybe DataStore is a more suitable place, but I don't feel very comfortable adding there, since at this point we have no idea about how many data types are we going add to that context, so it might become an asmdef nightmare moving it there

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/KernelCommunication/RPC/GeneratedCode/RendererProtocolService.gen.cs
pbosio: This is called by kernel and it won't try to cancel the task.
So any cancellation needed should be handled by the implementation of those delegates.
a quick example might be:
```
UniTask<RpcResponse> SomeProceduralCall(Request someRequest, Context context){
     try{
          var result = await SomeCancellableTask(someRequest, context);
          return responseTaskOK;
     } catch (Exception e){
          return responseTaskException;
     }
}
```

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/KernelCommunication/RPC/GeneratedCode/RendererProtocolService.gen.cs
pbosio: We had lot of back and forth thinking about that.
The idea why procedure calls (`SendCrdt`) are tasks is because maybe in some scenarios we want to process something "async" there, which is almost never, but we didn't feel comfortable to remove that possibility.
And streams (`CrdtNotificationStream`) are not actually running on a coroutine, it's just and old school enumerator (using iterator.MoveNext), we didn't think it was reasonable to enforce the extra work of every implementation using tasks when it is a "almost never" scenario and can have an easy workaround like `while(task_not_completed) yield return null`

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/KernelCommunication/RPC/ServerBuilder/RPCServerBuilder.cs
pbosio: Didn't add the constrain there because that class don't need it, it doesn't do any data handling, `T` might be whatever you want.
Maybe for some tests we would want to use other context type instead of `RPCContext`, so we have that flexibility by not adding that constrain

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/KernelCommunication/RPC/Services/CRDTService/CRDTService.cs
pbosio: It's static because it is stateless, we do not need an instance, it exist just to make it easier to read (because we avoid creating some lambda functions) and creating some static objects to reduce allocations.
We don't need it in service locator since no script needs no access it except for the rpc server to register the service.

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/KernelCommunication/RPC/Services/CRDTService/CRDTService.cs
pbosio: No sure if I understand.
In `Avatar.cs` we are cancelling the task if fails to load or if disposed while loading.
Here we are not cancelling the tasks since it's just deserializing and adding to a queue, so we don't have any of those kind of conditions where we should cancel the running task

pbosio:
@AjimenezDCL as discussed:
*moved Rpc Context to DataStore
*added cancellation tokens
*I also renamed the test file

AjimenezDCL (approved):

PR #2489, opened 2022-06-17
chore: textures compression feature flag

pravusjif:
### New Feature Flag: `ENABLE_PERF_TEX_COMPRESSION` / `DISABLE_PERF_TEX_COMPRESSION`

We use texture compression in these 4 places:
- https://github.com/decentraland/unity-renderer/blob/a7420c37bef56ef4aa4a3d3c382e6cf10bb7538d/unity-renderer/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs#L726
- https://github.com/decentraland/unity-renderer/blob/a7420c37bef56ef4aa4a3d3c382e6cf10bb7538d/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Texture/Asset_Texture.cs#L25
- https://github.com/decentraland/unity-renderer/blob/a7420c37bef56ef4aa4a3d3c382e6cf10bb7538d/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/TextureResource/AssetPromise_TextureResource.cs#L69
- https://github.com/decentraland/unity-renderer/blob/a7420c37bef56ef4aa4a3d3c382e6cf10bb7538d/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Textures/DCLTexture.cs#L125

### WHY
We need to have texture compression bound to a feature flag to test some scenarios with performance metrics.

### WHAT
~In this PR we are centralizing the compression toggle data in `TextureUtils`. If we don't centralize that somewhere, we end up having to have the same feature flag CONSTs and booleans for texture compression in those 4 classes, as it was in the [first implementation](https://github.com/decentraland/unity-renderer/pull/2489/commits/74fe779983b007388f1c8b526d2b106972b32087)~

- Created [new feature flag in Unleash](https://features.decentraland.systems/#/features/strategies/explorer-tex_compression) `TEX_COMPRESSION`
- Converted old `DataStore_TextureSize` to `DataStore_TextureConfig` and added a new `BaseVariable<bool> runCompression` in there as a single source of truth to determine if textures should be compressed or not when loaded.
- Bound texture compressions to a new condition that checks the new `runCompression` bool.
- Created simple `TextureCompressionTogglePlugin` to enable `runCompression` in the datastore when the pertinent `TEX_COMPRESSION` feature flag is ON.
- Removed Asset Bundles converter mentions regarding texture compression since we haven't been compressing textures in the converter for a long time ([since we introduced this `Application.isPlaying` conditional](https://github.com/decentraland/unity-renderer/blob/d515eaff5b0f6c5b7bd98f71d49b202feb876f35/unity-renderer/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs#L754))

### TEST
- This can't be tested in runtime besides checking that textures still load fine.
- Locally I tested with a `Debug.Log()` checking if textures are compressed or not based on the Feature flag and it worked fine for me.


unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/TextureHelpers/TextureHelpers.cs
```diff
@@ -76,4 +77,17 @@ public static Texture2D CopyTexture(Texture2D sourceTexture)
         
         return texture;
     }
+
+    private const string TEXTURE_COMPRESSION_FLAG_NAME = "tex_compression";
+#if UNITY_STANDALONE
+        bool compressTextures = false
+#else
+    private static bool compressTextures = DataStore.i.featureFlags.flags.Get().IsFeatureEnabled(TEXTURE_COMPRESSION_FLAG_NAME);
```
menduz, line 85: How does this work with race conditions? Because at the moment of initialization of the static bool, feature flags are not available. Thus, it may use a value that is not the one that comes from the feature flag system. It may work a better % of times for query params, but definitely not for the ffsystem.

A solution would be to always ask for the fresh `DataStore.i.featureFlags.flags.Get().IsFeatureEnabled(TEXTURE_COMPRESSION_FLAG_NAME)` in the Compress function instead of using the static bool

unity-renderer/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
```diff
@@ -717,15 +717,18 @@ protected void ConstructUnityTexture(TextureCreationSettings settings, byte[] bu
...
-
-            // We need to keep compressing in UNITY_EDITOR for the Asset Bundles Converter
-#if !UNITY_STANDALONE || UNITY_EDITOR
-            if ( Application.isPlaying )
+            
+            if (Application.isPlaying)
             {
+                
+                // We need to keep compressing in UNITY_EDITOR for the Asset Bundles Converter
+#if UNITY_EDITOR
```
menduz, line 725: Could this be a "IS_RUNNING_AB" instead of "UNITY_EDITOR"? It may be better to check that feature flags are working

menduz (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/TextureHelpers/TextureHelpers.cs
pravusjif: Yes, I just didn't want to ask `IsFeatureEnabled()` all the time but I guess I'll do something like that then

AjimenezDCL (changes requested):
I see some issues with the implementation

## 1. Not Data Driven Design friendly
The design is the root cause of needing strange stuff such as the usage of directives to make it mockable.

I would do any of the following:
1. Use `DataStore.i.featureFlags.flags.Get().flags[COMPRESSION_ENABLED_FLAG]` as a source of truth. 
- Check this data to know when to compress. 
- AB Converter flow should set that flag to `true` either by mocking the data directly or by having the environment set with the feature flag...

2. Create a new data in any of the existent data stores (create a new one if it doesnt fit in one)
- Create a plugin to manage that data that would rely on the FeatureFlags to enable/disable it.
- Check this data to know when to compress.
- AB Converter can easily set that flag to whatever it needs without worrying about pesky processes in the background since plugins wont be enabled during the conversion.

## 2. Some SOLID violations
The idea of encapsulating the behaviour in the `TextureHelpers.Compress` brings some problems.

- The usage of the `forceCompression` param is strange. If you dont want to compress the texture, just dont call the `Compress` method.
- It will lead to code that breaks the `interface segregation principle`. If you dont use the `forceCompression` param (or don't even notice it exists) you might end up doing obscure stuff like enabling/disabling the feature flags in the datastore just to workaround your problem.

I would just let every system to decide what to do. Most of them will just evaluate the DataStore to call (or not) the `Compress` method, some might offer capabilities to forceCompression, some might also never compress...

## 3. Test suite!
Basically we should do a couple of tests on the above.

unity-renderer/Assets/UnityGLTF/Scripts/GLTFSceneImporter.cs
pravusjif: In the end I discovered that we were already avoiding texture compressions with the Asset Bundle converter so I removed any comment regarding that.

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/TextureHelpers/TextureHelpers.cs
pravusjif: done, changed approach based on Alex feedback.

pravusjif:
- Changed appoach based on @AjimenezDCL [feedback](https://github.com/decentraland/unity-renderer/pull/2489#pullrequestreview-1012389252) point **1.2**.
- Updated PR description
- Still need to add tests coverage

pravusjif:
Status:
- From Desktop Client we need to find a way of disabling textures compression with this new approach
    - Unregistering the new plugin after it's been registered doesn't work
    - Forcing the datastore bool in MainDesktop Awake() doesn't work (combined with the previous step either)
- Missing tests

AjimenezDCL:
There was an issue with the Unregister of Plugins with feature flags, solved it.

Suduck (approved):

PR #2528, opened 2022-06-22
chore: performance meter update

### LINKED ISSUE #2460: Change performance meter to measure frame time instead of FPS
We should measure frame time instead of FPS, to prevent logarithmic scales which are not representative of the performance.

Today:
```
 * average FPS -> 37,76638
 * highest FPS -> 54
 * lowest FPS -> 14
 * 50 percentile (median) FPS -> 37
 * 95 percentile FPS -> 53
```

Proposed:
```
 * max frame time -> 54ms
 * avg frame time -> 30ms
 * min frame time -> 14ms
 * error percentage -> ±20% (this is new, using Alex meter)
 * 1 percentile frame time -> 10ms  (this is new, used to compare vs avg)
 * 50 percentile (median) FPS -> 30ms
 * 99 percentile frame time -> 50ms  (this is new from 95 to 99, used to compare vs avg)
```

Two new metrics are added, the P99 and P1 (worst and best 1%), and the error percentage which is being implemented by @AjimenezDCL in another PR.

The goal of stability is not frame time, it is frame pacing. It is better to have frames regularly at 30ms than 80% at 20ms and 20% at 70ms.

The metric to find out frame stability is the difference between the p1 and the average, the lower the better. Also adding the error % will provide us with a similar metric

Kinerius:
## What does this PR change?

Updated the performance meter tool to give us frame data instead of fps, also added error margins

Fixes #2460

## How to test the changes?

run `clientDebug.RunPerformanceMeterTool(30)` in the browser console an wait for the results

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Debugging/Performance/PerformanceMeterController/PerformanceMeterController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Debugging/Performance/PerformanceMeterController/PerformanceMeterController.cs
@@ -323,9 +323,9 @@ namespace DCL
         private int CalculatePerformanceScore()
         {
             double desiredFrameTime = Settings.i.qualitySettings.Data.fpsCap ? 1000/30.0 : 1000/60.0;
-            double fpsScore = Mathf.Min((float)(averageFrameTime / desiredFrameTime), 1); // from 0 to 1
+            double frameScore = Mathf.Min((float)(desiredFrameTime/ averageFrameTime), 1); // from 0 to 1
             double hiccupsScore = 1 - (float) totalHiccupFrames / samples.Count; // from 0 to 1
-            double performanceScore = (fpsScore + hiccupsScore) / 2 * 100; // scores sum / amount of scores * 100 to have a 0-100 scale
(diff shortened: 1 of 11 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Debugging/Performance/PerformanceMeterController/PerformanceMeterController.cs
```diff
@@ -163,15 +173,17 @@ private void OnMetricsChange(PerformanceMetricsData newData, PerformanceMetricsD
                 secondsConsumed = Time.timeSinceLevelLoad - lastSavedSample.currentTime;
             }
 
-            SampleData newSample = new SampleData()
+            float frameTimeMs = Time.deltaTime * 1000f;
```
aixaCode, line 176: We could use `ProfilerRecorder.StartNew(ProfilerCategory.Internal, "Main Thread")` to collect the samples for frame time.

aixaCode: I'm not sure as well, but would improve consistency.

aixaCode (approved):

PR #2531, opened 2022-06-22
feat: friends lazy loading

lorux0:
## What does this PR change?

Implementation of lazy loading of friend lists.

## How to test the changes?

1. Open the friends window
2. In case the service is down, right click over `FriendsController` component at `Main` object -> Force Initialization
3. Open the friend list tab, mocked friends should be displayed
4. Use the search bar to look for any friend, mocked friend entries should be matched
5. In case you have more friends to load, click on the show more button
6. Open the friend request tab, send/received requests should be displayed
7. In case you have more requests to load, click on the show more button

## Our Code Review Standards


sandrade-dcl:
Some things I have seen during the testing:

### Number of friends hidden
- The number of friends hidden is not being showed. A "Friends hidden" static message is being used instead.
image

### Show More button
- We need to change the SHOW MORE button approach by a direct loading dependant on the the scroll position. It is something dicussed yesterday with Romi [here](https://decentralandteam.slack.com/archives/C03HN6KUUUF/p1655915456807709). We can talk about it to be synched.

- When I open the FRIENDS window by first time, I see the SHOW MORE button. But if I close the window and open it again, the SHOW MORE button dissapears.
image
image

### Loading feedback
- I'm not seeing any loading feedback in the FRIENDS window when it is being loaded.
- I'm not seeing any loading feedback in the REQUESTS window when it is being loaded.
- When we click on any SHOW MORE button (in FRIENDS or REQUEST section), we don't see any loading feedback. We have to add little loading spinners in the corresponding places while the data is coming from kernel.

### Other
- Don't we have any SENT FRIEND REQUEST mocked? I cannot test this part.
image

PR #2539, opened 2022-06-23
chore: avatar should face same direction as spawn point after teleport

dalkia:
## What does this PR change?

When in _third person mode_, after a teleport to a new location, the avatar was facing always north and not the direction of the camera forward. With this PR, the avatar now looks to the same direction as the camera forward.

## How to test the changes?

1) Keep _third person mode_
1) Teleport to -74,53
2) Check that the avatar is looking west

## Our Code Review Standards


davidejensen (approved):
Works fine now, GJ!

PR #2545, opened 2022-06-24
feat: friend entries loaded by the scroll position & visual improvements on the loading spinners

lorux0:
## What does this PR change?

- friend entries loaded when you scroll down to the bottom on friend list and friend request list
- the "show more" option has been changed adding a loading spinner & a label modification
- the loading spinner style has been changed

## How to test the changes?

1. Open the friends window from the taskbar
2. While the friend service is not initialized, check that the loading spinner has been resized to 40px
3. Force the initialization of friends
4. Scroll down to the bottom of the friend list
5. There should be an ui component indicating that we are loading more entries
6. Load all the available entries, the "show more" component should be hidden
7. Go to friend request tab and repeat the process

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Rendering/UI/ChargeWheel/M_ChargeBarV2Thicker.mat
+++ b/unity-renderer/Assets/Rendering/UI/ChargeWheel/M_ChargeBarV2Thicker.mat
@@ -48,13 +48,13 @@ Material:
     - _StencilOp: 0
     - _StencilReadMask: 255
     - _StencilWriteMask: 255
-    - _fillHead: 0.9301209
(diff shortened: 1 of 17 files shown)


sandrade-dcl (changes requested):
Testing the build I have seen this:
- When I open the FRIENDS tab and I scroll down to load more entries, the loading spinner seems to be hidden:
image

- The same seems to be happened in the REQUESTS tab:
image

lorux0:
I can see it correctly on unity & web (windows). I think its a problem with the shader. We can fix this issue in a different pr

PR #2565, opened 2022-06-29
fix: Tutorial not starting on new user

dalkia:
## What does this PR change?

The tutorial was not starting because there was a check on `CommonScriptableObjects.rendererState.Get()` before subscribing the `OnRenderingStateChanged` method, which eventually triggered the tutorial. If the user went first to the avatar customization scene, renderState was true, and the event was not subscribed when the user got to Genesis Plaza. 

This was fixed by removing the check.

## How to test the changes?

1. Delete cookies or start Incognito Window to have a fresh start
2. Go to: https://play.decentraland.zone/?renderer-branch=fix/tutorial-not-starting
3. Check that the tutorial appears after avatar customization

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Tutorial/Scripts/TutorialController.cs
+++ b/unity-renderer/Assets/Tutorial/Scripts/TutorialController.cs
@@ -177,10 +177,8 @@ namespace DCL.Tutorial
             NotificationsController.disableWelcomeNotification = true;
 
             WebInterface.SetDelightedSurveyEnabled(false);
-            
-            IWorldState worldState = Environment.i.world.state;
-            
-            if ((!CommonScriptableObjects.rendererState.Get()) || worldState?.currentSceneId == null)
+
+            if (!CommonScriptableObjects.rendererState.Get())
                 CommonScriptableObjects.rendererState.OnChange += OnRenderingStateChanged;
             else
                 OnRenderingStateChanged(true, false);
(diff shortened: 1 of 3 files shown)


dalkia:
Is this screen necessary? Consider the current flow, the only way it seems to appear is if it comes from `openedFromDeepLink` condition.

image (1)

unity-renderer/Assets/Tutorial/Scripts/TutorialController.cs
```diff
@@ -177,11 +177,8 @@ internal void SetupTutorial(string fromDeepLink, string enableNewTutorialCamera,
             NotificationsController.disableWelcomeNotification = true;
 
             WebInterface.SetDelightedSurveyEnabled(false);
-
-            if (!CommonScriptableObjects.rendererState.Get())
-                CommonScriptableObjects.rendererState.OnChange += OnRenderingStateChanged;
-            else
-                OnRenderingStateChanged(true, false);
+            
+            CommonScriptableObjects.rendererState.OnChange += OnRenderingStateChanged;
```
sandrade-dcl, line 181: It is needed to keep the `if (!CommonScriptableObjects.rendererState.Get())` condition because there are some times when we need to start directly the tutorial while we are already in the world (with the `rendererState` as `true`). For instance clicking in the [TUTORIAL] button from the settings window:

image

sandrade-dcl (changes requested):
I think it is worth to investigate why the `rendererState` is being `true` while we are in the SignUp stage, because it shouldn't.

sandrade-dcl (approved):
Code reviewed an build tested... it is working fine now. Good work!

Try to avoid the use of the `IntegrationTestSuite_Legacy`, but if you really need it, ok 👍

PR #2575, opened 2022-07-04
feat: billboard component

Suduck:
## What does this PR change?

This PR adds the billboard component to the new ECS.

It also rotates the TextShape gameobject 180 degrees to match all the expected behaviour of text ( instead of look like a mirror).

This is done because the billboard component will make the text shape component to face the camera inverted, with this this is solved

## How to test the changes?

Can't be tested

## Our Code Review Standards


unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Billboard/Handler/BillboardComponentHandler.cs
```diff
@@ -0,0 +1,103 @@
...
+            updateEventHandler.AddListener(IUpdateEventHandler.EventType.LateUpdate, LateUpdate);
+        }
+
+        public void OnComponentCreated(IParcelScene scene, IDCLEntity entity)
+        {
+            this.entity = entity;
+            this.scene = scene;
+            
+            // Each time the camera position changes, we update the rotation of the entity
+            cameraPosition.OnChange += CameraPositionChanged;
```
pbosio, line 33: do we need this event if we are re-orienting the entity's transform in each late update?

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Billboard/Handler/BillboardComponentHandler.cs
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        private void ChangeOrientation()
+        {
+            if (entityTransform == null || model == null)
+                return;
+            UnityEngine.Vector3 lookAtVector = GetLookAtVector();
+            if (lookAtVector != UnityEngine.Vector3.zero)
+                entityTransform.forward = lookAtVector;
+        }
```
pbosio, line 99: maybe we should add a "TODO" comment here to change it later? because we probably want to write to our `Transform` component instead of modify unity's `Transform` directly

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Billboard/Handler/BillboardComponentHandler.cs
```diff
@@ -0,0 +1,103 @@
...
+        private UnityEngine.Vector3 lastPosition;
+
+        private IDCLEntity entity;
+        private IParcelScene scene;
+        private PBBillboard model;
+        
+        public BillboardComponentHandler(IUpdateEventHandler updateEventHandler)
+        {
+            this.updateEventHandler = updateEventHandler;
+            updateEventHandler.AddListener(IUpdateEventHandler.EventType.LateUpdate, LateUpdate);
```
pbosio, line 24: We should start discussing moving all components that require an update loop to a more "System" approach in the near future

pbosio (approved):

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Billboard/Handler/BillboardComponentHandler.cs
Suduck: I think we can safely remove it, since the camera is updated in the `Update `function, removed!

Suduck: I think we can safely remove it, since the camera is updated in the `Update `function, removed!

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Billboard/Handler/BillboardComponentHandler.cs
Suduck: Sure! Right now this is the option that we have but we can start discussing it

AjimenezDCL: I agree!

AjimenezDCL (approved):

PR #2360, opened 2022-05-26
fix: remove perfomance regression

Suduck:
This reverts commit e5b5942ce445c0412b2f4aaec4c44b4ba8e2368a.

## What does this PR change?

This PR reverts the update of the Protobuff library. because the new version is less perfomant.

image

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=fix/perfomance-regresion&position=60%2C-71&NETWORK=ropsten
3. This scene should work smooth

## Our Code Review Standards

https://github.com/decentraland/unity-renderer/blob/master/docs/code-review-standards.md

PR #2577, opened 2022-07-04
feat: update protoc 3.20.1

kuruk-mm:
in previous updating `protoc`, we had a performance regression which was inverted here https://github.com/decentraland/unity-renderer/pull/2360

This was because the newer version of `protoc` added a runtime reflection copy, which decreases the performance but may work better for IL2CPP (based on the README comment from @BrianAmadori here: https://github.com/decentraland/renderer-protocol)
In any case, if we add `GOOGLE_PROTOBUF_REFSTRUCT_COMPATIBILITY_MODE`, we disable the reflection, and the performance should be the same as the current version.

This PR include:
- update of protoc 3.20.1
- update regenerated code using the protoc 3.20.1
- fix component ID for CAMERA_MODE_AREA (was 1070, should be 1071)
- change the download source from `decentraland-ecs` to `@dcl/sdk`, and fix the downloader to accept packages with slashes (any way, we are using the local .proto for now)

## Our Code Review Standards


Suduck (approved):
LGTM! @pravusjif  please could you check if the performance is dropping with this PR?

Suduck (changes requested):
The perfomance has droped even with the flag.

Check this scene in main and this branch, if you take a look ad the scene of the purples cubes, and start it you will notice that it is slower in this branch. The same happened to me:

Main:
https://play.decentraland.zone/?NETWORK=ropsten&realm=artemis&position=60%2C-71&island=I4h

Branch:
https://play.decentraland.zone/?renderer-branch=feat%2Fupdate-protoc-3.20-1&NETWORK=ropsten&realm=artemis&position=60%2C-70


image

kuruk-mm:
Please test with:
Main:
https://play.decentraland.zone/?NETWORK=ropsten&realm=artemis&position=60%2C-71&island=I4h&kernel-branch=fix/legacy_transform_serialization

Branch:
https://play.decentraland.zone/?renderer-branch=feat%2Fupdate-protoc-3.20-1&NETWORK=ropsten&realm=artemis&position=60%2C-70&kernel-branch=fix/legacy_transform_serialization

PR #2579, opened 2022-07-04
fix: GLTFImporter to generate a mesh when materials are ignored.

Kinerius:
## What does this PR change?

Fixes the import of GLTF with Import Materials not checked.

**Important note**: Importing a GLTF with materials during the first import is still broken, this PR does not fix that, it just gives us an option.

## How to test the changes?

When starting Decentraland, the avatar should load correctly.

## Our Code Review Standards


davidejensen (approved):
It works!!!

PR #2583, opened 2022-07-04
fix: ECS7 CRDT remove components and default values

pbosio:
We used to assume that incoming crdt message without `data` means that the component was going to be removed.
But now, preparing for having components with default values, empty data could also mean that the component should be created with it default values.
For that reason:
*If no data exist but the crdt message type is `PUT_COMPONENT` we create an empty byte array for `data`, so protobuf can deserialize that message and create those components with it default values.
*If no data exist but the crdt message type is `DELETE_COMPONENT` we leave `data` as null as signal to delete that component


unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/KernelCommunication/BinaryMessage/KernelBinaryMessageDeserializer.cs
```diff
@@ -24,18 +24,18 @@ public static IEnumerator<object> Deserialize(IBinaryReader reader)
             while (reader.CanRead())
             {
                 int messageLength = reader.ReadInt32();
-                int messageType = reader.ReadInt32();
+                CrdtMessageType messageType = (CrdtMessageType)reader.ReadInt32();
```
menduz, line 27: Will this fail if somebody (inadvertently) adds a new value to the enum?

pbosio: It shouldn't since C# won't try to convert the value to match a name on the enum.
So: 
1 ==  CrdtMessageType.PUT_COMPONENT
2 ==  CrdtMessageType.REMOVE_COMPONENT
3 == (CrdtMessageType) 3

unity-renderer/Assets/Scripts/MainScripts/DCL/CRDTProtocol/Deserializer/Tests/CRDTDeserializerShould.cs
```diff
@@ -189,6 +189,31 @@ public void CopyUnmanagedDataCorrectly()
...
+        public void SetDataOnPutComponentWithEmptyPayload()
+        {
+            byte[] message =
+            {
+                0, 0, 0, 32, //header: length = 32
+                0, 0, 0, 1, //header: type = 1 (PUT_COMPONENT)
+                0, 0, 0, 1, // component: entityId
+                0, 0, 0, 1, // component: componentId
+                0, 0, 0, 0, 0, 0, 0, 1, // component: timestamp (int64)
+                0, 0, 0, 0, // component: data-lenght (0)
```
Kinerius, line 202: Nice comments!

Issue #2433, opened 2022-06-08
Implement scene bounds detection with Unity Physics system


pravusjif:
We should see if we can somehow easily disable our entities position and mesh bounds checking in the Scene Bounds Checker and use Unity's Physics system with trigger colliders.

Probably every entity will need at least 1 tiny collider, and when there's a mesh, 1 box collider encompassing the whole mesh bounds.

Uncertanties to resolve:
- [x] If we re-enable physics auto-simulation, do Asset Bundles loading still have the big Physics spike bug? -> Answer: NO, apparently Unity already fixed this so we don't have this problem anymore with the auto-simulation ON (couldn't get a reliable profiling session with WebGL build so the conclusion is based on editor testing only).
- [x] Otherwise, can we make use of Physics trigger system for scene bounds checking also running the physics simulation on demand? -> Answer: Yes with Alex's approach it looks like we can make use of the trigger events to track which which parcels is an entity touching
- [x] Can we avoid having rigibodies with every entity collider? -> Yes, if we put the rigidbodies in the parcel trigger colliders. Maybe we can minimize the new monobehaviours amount also by putting them in the parcels.

After getting that working we should compare performance with the [latest scene bounds check solution PR](https://github.com/decentraland/unity-renderer/pull/2164)

------------------------------------------------------

### Defined approach
 
- [x] Having 1 trigger collider on every parcel
- [x] Tracking somewhere (Entity/Scene Bounds Checker) the parcels that an entity/mesh is touching.
- [x] Use `OnTriggerEnter()` and `OnTriggerExit()` trigger events (having a monobehaviour for that on parcels or entities, probably it's better on parcels because there will always be more entities than parcels...)
- [x] To detect that an entity entered a parcel that it shouldn't, parcel colliders should be somehow bound to their scene ids.
- [x] Move out-of-bounds detection to entity/parcel based on which scenes each entity/mesh is touching (based on the parcels it's touching)
- [x] Evaluate if we can remove the Scene Bounds checker altogether after this new solution
- [x] Benchmark performance cost comparing with the previously devised solution at https://app.zenhub.com/workspace/o/decentraland/unity-renderer/issues/2132
- [x] Test the new approach solves the same issues involved in the [previous solution approach](https://github.com/decentraland/unity-renderer/pull/2164)

#### Notes
[...]

PR #2617, opened 2022-07-12
spike: physics based scene bounds check

pravusjif:
### Spike [Finished]
- [x] Created `ParcelBoundsCheckCollider` and `EntityBoundsCheckCollider` physics layers and update physics collision matrix so that they collide only with each other
- [x] Instantiate a trigger collider for every parcel (using `ParcelBoundsCheckCollider` layer, bounds encompassing parcel's size) and a trigger collider for every entity (using `EntityBoundsCheckCollider` layer, bounds encompassing entity attached mesh or having a minimum size otherwise)
- [x] Implement Entity's "trespassing scenes" counter to trigger its inside/outside bounds effect.
- [x] Disable Scene Bounds Checker
- [x] Reroute entity's feedback style effect triggering that currently runs on the Scene Bounds Checker system
- [x] Implement height checking (no physics, same as we were checking it)
- [ ] Tighten performance a bit more? (more recycling: least monobehaviours usage + less objects instantiation )
- [x] Benchmark performance compared to production and to [previous solution approach](https://github.com/decentraland/unity-renderer/pull/2164)
- [x] Test the new approach solves the same issues involved in the [previous solution approach](https://github.com/decentraland/unity-renderer/pull/2164)

------------------------------------

### Final Implementation

- [x] Fix issue with several meshes being hidden when they shouldn't
- [ ] Change approach a bit to use MeshColliders instead of BoxCollider based on merged bounds, due to [problems with irregular-shaped scenes and rotated entities merged bounds](https://github.com/decentraland/unity-renderer/issues/2433#issuecomment-1195476203) + measure performance again using MeshColliders.
- [ ] Refactor to remove the Scene Bounds Checker and move its feedback style toggling core logic to the pertinent place (Entity/Parcel/NewClass)
- [ ] Test optimization: move collision events to ParcelScene to take advantage of that existent monobehaviour (as there are always less parcels than entities, so less monobehaviours in runtime) + compare performance doing it that way since the parcel will have to detect the entity somehow (`GetComponent` or grabbing the entity GameObject and looking for it in the current scene entities or similar) and it will cost some performance.
- [ ] Make sure this works correctly in preview mode by running the editor with a local scene

------------------------------------

### Testing
- Ropsten: https://play.decentraland.zone/?renderer-branch=spike/PhysicsBasedSceneBoundsCheck&NETWORK=ropsten&position=62%2C-68
- Preview-mode locally: ???


pravusjif:
Performance comparison at https://github.com/decentraland/unity-renderer/issues/2433#issuecomment-1194156935

pravusjif:
Closing this PR as defined in https://github.com/decentraland/unity-renderer/issues/2433#issuecomment-1197485502

PR #2662, opened 2022-07-19
feat: multiple prefabs support for channel entries

lorux0:
## What does this PR change?

The channel lists (conversation & browser) now supports a configurable scriptable object that provides different types of entry instances depending on the channel id.

## How to test the changes?

1. Open the conversation list
2. All channels should be displayed correctly
3. The nearby channel prefix should be `~`
4. The nearby channel should not have the option to leave the channel
5. Go to channel browser (search button)
6. Search for channels, everything should work correctly

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WorldChatWindowHUD/ChannelEntryFactory.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WorldChatWindowHUD/ChannelEntryFactory.cs
@@ -1,0 +1,29 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using UnityEngine;
+
+namespace DCL.Chat.HUD
+{
+    [CreateAssetMenu(fileName = "ChannelEntryFactory", menuName = "DCL/Social/ChannelEntryFactory")]
+    public class ChannelEntryFactory : ScriptableObject
+    {
+        [Serializable]
+        private struct SpecialEntry
+        {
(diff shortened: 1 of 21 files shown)


aixaCode (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WorldChatWindowHUD/ChannelEntryFactory.cs
```diff
@@ -0,0 +1,29 @@
...
+        }
+        
+        [SerializeField] private PublicChatEntry defaultEntryPrefab;
+        [SerializeField] private SpecialEntry[] specialEntryPrefabs;
+
+        private Dictionary<string, PublicChatEntry> specialEntries;
+
+        public PublicChatEntry Create(string channelId)
+        {
+            specialEntries ??= specialEntryPrefabs.ToDictionary(entry => entry.channelId, entry => entry.prefab);
```
aixaCode, line 25: As `specialEntries` lifetime will persist between switching the play mode in editor, it can happen that `specialEntries` can have old data if we modify the scriptable object. Do we have any pattern for solving that? Or we are ok with that corner case?

lorux0: i wouldn't worry about that really

sandrade-dcl (approved):

PR #2663, opened 2022-07-19
feat: ECS7 send camera and player's avatar transform to scene

pbosio:
add systems to send Camera's and Avatar's Transform to the scene
issue: https://github.com/decentraland/sdk/issues/339


unity-renderer/Assets/DCLPlugins/ECS7/ComponentCrdtWriteSystem/ComponentCrdtWriteSystem.cs
```diff
@@ -21,9 +21,9 @@ private class MessageData
     private readonly ISceneController sceneController;
     private readonly IWorldState worldState;
 
-    private readonly Dictionary<string, CRDTProtocol> outgoingCrdt = new Dictionary<string, CRDTProtocol>(24);
-    private readonly Queue<MessageData> queuedMessages = new Queue<MessageData>(24);
-    private readonly Queue<MessageData> messagesPool = new Queue<MessageData>(24);
+    private readonly Dictionary<string, CRDTProtocol> outgoingCrdt = new Dictionary<string, CRDTProtocol>(60);
```
Suduck, line 24: Can we make these numbers constant to specify what they are?

Suduck (commented):
Looks good! However, I am not very sure about the static approach for the systems. The current design is going against that and I am not sure if the performance gain is worth it

unity-renderer/Assets/DCLPlugins/ECS7/Systems/Helpers/ECSSystemsReferencesContainer.cs
```diff
@@ -0,0 +1,12 @@
+using System.Collections.Generic;
+using DCL.Controllers;
+using DCL.ECSRuntime;
+
+namespace ECSSystems.Helpers
+{
+    public static class ECSSystemsReferencesContainer
+    {
+        public static IReadOnlyList<IParcelScene> loadedScenes;
+        public static IECSComponentWriter componentsWriter;
```
AjimenezDCL, line 10: This should be in the DataStore

AjimenezDCL (approved):
Good job! My only concern is that static class with a state, that information should be in the DataStore

unity-renderer/Assets/DCLPlugins/ECS7/ComponentCrdtWriteSystem/ComponentCrdtWriteSystem.cs
pbosio: It's just a number calculated on top of my head to reduce the possibility of the dictionary resizing.
The number could be a lot lower or a lot higher depending on the amount of loaded scenes.
So adding a constant would generate more confusion than clarity

unity-renderer/Assets/DCLPlugins/ECS7/Systems/Helpers/ECSSystemsReferencesContainer.cs
pbosio: That class is just for containing references that are needed for a lot of other systems.
Main intention is to avoid accessing DataStore directly, so, for example `loadedScenes` there contain a reference to a DataStore here: https://github.com/decentraland/unity-renderer/blob/badd9b2ce38a9255208771c78bfa56c69f593f4f/unity-renderer/Assets/DCLPlugins/ECS7/ECS7Plugin.cs#L37
because, when we start to add up systems, it start to become noticeable the performance loss on accessing the DataStore directly on every system, on every update loop

AjimenezDCL: The performance loss (I would love to see numbers about the actual loss) comes from the redirection between the instance, the sub data store and then the data.

You can circumvent the issue easily with dependency injection. Your system that would do `ReferencesContainer` now simply accepts on its constructor an instance to whatever it needs and altough that stuff lives in the DataStore is never accessed through it. 

You avoid the performance loss, keep the design coherent with the rest of the project and also avoid the pitfalls associated to static classes with state.

pbosio: Yes, DataStore do a lookup at a dictionary and then a cast, which is not convenient to do multiple times in every frame.
In any case, I've just removed the static class that worked as reference container here: https://github.com/decentraland/unity-renderer/pull/2663/commits/3017565965a96310df50cf55792670881fb18a68

pbosio:
> Looks good! However, I am not very sure about the static approach for the systems. The current design is going against that and I am not sure if the performance gain is worth it

what do you mean by that?
those are just function that are self-contained, what could we gain of doing it differently?

AjimenezDCL (approved):

Suduck (approved):

PR #2667, opened 2022-07-20
chore: ui camera is now always an overlay

AjimenezDCL:
Our current Camera Stack is:
```
- Skybox (Base)
   - World Camera (Overlay)
   - HUDs Camera (Overlay)
```
Going into a fullscreen HUD will disable the cameras for performance's sake. Since the HUDs are now in camera space, we were forced to change the camera mode of the HUD camera to Base (instead of Overlay) when in this mode. 

Now the skybox camera is never disabled, even if we are in a fullscreen HUD. The impact in performance is not even noticeable and simplyfies a lot the design.


sandrade-dcl (approved):
LGTM!

Suduck (approved):

PR #2672, opened 2022-07-20
spike: unity 2021 emscripten memory cap update

pravusjif:
Added needed emscripten memory cap arguments in PreBuildProcessing script


pravusjif:
- I confirm the emscripten args added in this new prebuild script are working, since I tested with the `--memoryprofiler` argument and it logs every time the memory is expanded, in the browser's console
- By forcing the Scene Load Radius to a bigger number like 16 (`LOS=16`) I got to see the memory expanding up to **2625mb** before the executaion got broken by other issues:
Screen Shot 2022-08-26 at 17 41 39

URL for testing https://play.decentraland.zone/?renderer-branch=spike%2FUnity2021EmscriptenMemoryCapUpdate&position=-70,0&LOS=16&DISABLE_ASSET_BUNDLES

PR #2675, opened 2022-07-20
fix: cast exceptions on friends hud when updating the profile

lorux0:
## What does this PR change?

It is a refactor to avoid casting different types of friend entries on the friends hud controller/view.
It fixes exceptions that were occurring when the profile was updated.

## How to test the changes?

1. Open the friends window
2. Check that your friends are correctly displayed
3. Go to friend requests and check they are correctly displayed
4. Send and receive friend requests
5. Accept them or reject them, the view should be updated correctly

## Our Code Review Standards


sandrade-dcl (approved):
Reviewed and tested... and it's workig fine. GJ!

PR #2677, opened 2022-07-21
fix: Join feature doesn't work correctly

sandrade-dcl:
## What does this PR change?
There used to be a feature that when a friend was online, their status would be reported as well as its whereabouts and it was possible to jump to the position of that user. However as of today, it seems this is not longer possible:

Screen Shot 2022-07-20 at 16 16 55

## How to test the changes?
1. Go to: https://play.decentraland.zone/?renderer-branch=fix/join-feature-not-working-as-expected
1. Check friends that are online
1. Verify that the position and realm are visible in the friends list or the chat window
1. Verify it's possible to jump in the user by clicking the join button

## Our Code Review Standards


lorux0 (approved):
nice!

PR #2680, opened 2022-07-21
fix: invisible character preview

AjimenezDCL:
When using an Avatar with a Hologram all the renderers (including the Bodyshape's ones) point its bones to the BaseAvatar (the hologram). This is done to have smooth transitions from hologram to real avatar.

The bug goes as follow:

- Spawn in Genesis Plaza with more people
- Avatars with holograms are rendered
- Bunch of bodyshapes are pointing its bones to their corresponding BaseAvatar
- You teleport to another location,  avatars from other people are disposed (renderers are returned to the library to be reused)
- Open Backpack
- CharacterPreview loads an AvatarWithoutHologram, that means using the upperbody as the bones orchestator)
- Most wearables are retrieved from the library because they are already loaded, bodyshape included.
- All wearables use the upperbody bones to animate themselves but upperbody bones are pointing to the `BaseAvatar` of its previous "life".
- Wearables are now being animated in MORDOR (where the assets in the library live) and therefore invisible

# Test
- Do multiple visits to crowded places and then teleport before opening the backpack. The Character Preview should work

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/AvatarSystem/AvatarSystemUtils.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/AvatarSystem/AvatarSystemUtils.cs
@@ -41,16 +41,24 @@ namespace AvatarSystem
             return (mainTextureUrl, maskTextureUrl);
         }
 
-        public static void CopyBones(SkinnedMeshRenderer source, IEnumerable<SkinnedMeshRenderer> targets)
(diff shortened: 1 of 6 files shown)


sandrade-dcl (approved):
Tested and owrking fine in my side. Good work!

lorux0 (approved):
fixed, thank you alex!

davidejensen (approved):
Thanks Alex the Savior!

PR #2682, opened 2022-07-21
revert: forward renderer changes for blur

AjimenezDCL:
Reverted the forward renderer changes for blur since it's not implemented yet.

# Test
Everything should look as usual.


davidejensen (approved):
Works as expected

PR #2688, opened 2022-07-21
fix: player positions in minimap is not unclamped

AjimenezDCL:
Player positions in minimap was being clamped into integers.

# Test

Go to a crowded place, player position should correspond to its representation in the minimap


Kinerius (approved):
🚀

PR #2689, opened 2022-07-21
fix: layout updating when long messages are added to the chat

lorux0:
## What does this PR change?

It fixes a layout issue that was causing the chat messages to be trimmed when they are too long.

image

## How to test the changes?

1. Open any chat window
2. Write very long messages
3. The scroll should behave correctly

## Our Code Review Standards


Kinerius (approved):
Nais

PR #2692, opened 2022-07-21
fix: Remove avatar loading animation if avatar appears inside modifier area

dalkia:
## What does this PR change?

Changes the visibility of the `AvatarWithHologram` animation depending if the player is teleported into an `AvatarHideModifierArea`. With this PR, the loading animation should not be seen.

It can happen that the animation is visible when starting. This is because there is no way to assure the order of the loading animation vs the triggering of entering the area. There is no way to predict it either. So, if the animation gets triggered, it should be turned off immediately.

## How to test the changes?

1) Open two instances of the race game: https://play.decentraland.zone/?renderer-branch=fix%2Fremove_avatar_particles_in_avatar_modifier&NETWORK=mainnet&position=-13%2C77&CATALYST=peer-testing.decentraland.org&realm=v2%7Epeer-testing.decentraland.org&DEBUG_SCENE_LOG=&island=Iv1n
2) Enter the race with one player. When the second player enters, no hologram or particle animation should be visible when moving the second player into the racing position. 

## Our Code Review Standards


dalkia:
Video reference. Compare to the [original video issue](https://app.zenhub.com/workspaces/explorer-6047bad476c3c0001942ee91/issues/decentraland/unity-renderer/2647), where particles were clearly visible

https://user-images.githubusercontent.com/1999557/180906469-cf008e1d-c7f7-4858-aa44-0433c0576db0.mp4

unity-renderer/Assets/Scripts/MainScripts/DCL/AvatarSystem/Visibility.cs
```diff
@@ -11,16 +11,18 @@ public class Visibility : IVisibility
...
         private List<Renderer> facialFeatures = null;
+        private IBaseAvatar baseAvatar = null;
 
         /// <summary>
         /// Bind a set of renderers, previous renderers enabled wont be modified
         /// </summary>
         /// <param name="combinedRenderer"></param>
         /// <param name="facialFeatures"></param>
-        public void Bind(Renderer combinedRenderer, List<Renderer> facialFeatures)
+        public void Bind(Renderer combinedRenderer, List<Renderer> facialFeatures, IBaseAvatar baseAvatar = null)
```
AjimenezDCL, line 21: This creates an unneeded coupling between `Visibility` and `IBaseAvatar`.

I'd handle the visibility for the BaseAvatar in `AvatarWithHologram.AddVisibilityConstrain` and `AvatarWithHologram.RemoveVisibilityConstrain` and remove this dependency.

AjimenezDCL (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/AvatarSystem/Visibility.cs
dalkia: Done! I had to add an `if` to check the key that is being addded since there are different ones that may affect the `AvatarWithHologram`. Maybe turning these consts strings into an enum somewhere would be nice?

unity-renderer/Assets/Scripts/MainScripts/DCL/AvatarSystem/AvatarWithHologram.cs
```diff
@@ -113,9 +111,20 @@ public async UniTask Load(List<string> wearablesIds, AvatarSettings settings, Ca
...
+        {
+            visibility.AddGlobalConstrain(key);
+            baseAvatar.ToggleHideAvatarArea(key.Equals("IN_HIDE_AREA"));
+        }
 
-        public void RemoveVisibilityConstrain(string key) { visibility.RemoveGlobalConstrain(key); }
+        public void RemoveVisibilityConstrain(string key)
+        {
+            visibility.RemoveGlobalConstrain(key);
+            if (key.Equals("IN_HIDE_AREA"))
```
AjimenezDCL, line 123: Does it matter? Particles must be disabled regarding the source that hides the avatar.

I would disable the particles if the avatar is not visible (regarding the source). You can implemente a new set of methods to consult the status on visibility.
``` 
IVisibility.IsGloballyVisible()
IVisibility.IsMainRendererVisible()
IVisibility.AreFacialFeatureVisible()
```

And then I would just disable particles if `IsGloballyVisible == false` and enable them otherwise.

unity-renderer/Assets/Scripts/MainScripts/DCL/AvatarSystem/BaseAvatar.cs
```diff
@@ -51,10 +54,21 @@ public async UniTask FadeOut(MeshRenderer targetRenderer, bool playParticles, Ca
...
-            await avatarRevealer.StartAvatarRevealAnimation(playParticles, cancellationToken);
+            await avatarRevealer.StartAvatarRevealAnimation(playParticles && !isInsideHideArea, linkedCt);
+            
+            disposeCts?.Dispose();
+            disposeCts = null;
+        }
+        
+        public void ToggleHideAvatarArea(bool entered)
+        {
+            if(entered) disposeCts?.Cancel();
```
AjimenezDCL, line 70: You should dispose the CTS here as well.

Also, if you are cancelling the animation you have to set the final state. Otherwise, if the avatar goes into a hidden area in the middle of the transition, it will stay there forever.

unity-renderer/Assets/Scripts/MainScripts/DCL/AvatarSystem/BaseAvatar.cs
```diff
@@ -51,10 +54,21 @@ public async UniTask FadeOut(MeshRenderer targetRenderer, bool playParticles, Ca
...
+            
             avatarRevealer.AddTarget(targetRenderer);
-            await avatarRevealer.StartAvatarRevealAnimation(playParticles, cancellationToken);
+            await avatarRevealer.StartAvatarRevealAnimation(playParticles && !isInsideHideArea, linkedCt);
+            
+            disposeCts?.Dispose();
+            disposeCts = null;
+        }
+        
+        public void ToggleHideAvatarArea(bool entered)
```
AjimenezDCL, line 68: I would make the `BaseAvatar` oblivious to hide areas and such to avoid coupling (at least conceptually by now).

AjimenezDCL (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/AvatarSystem/BaseAvatar.cs
```diff
@@ -54,21 +53,22 @@ public async UniTask FadeOut(MeshRenderer targetRenderer, bool playParticles, Ca
...
             
-            disposeCts ??= new CancellationTokenSource();
-            CancellationToken linkedCt = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, disposeCts.Token).Token;
+            transitionCts ??= new CancellationTokenSource();
+            CancellationToken linkedCt = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, transitionCts.Token).Token;
             linkedCt.ThrowIfCancellationRequested();
             
             avatarRevealer.AddTarget(targetRenderer);
-            await avatarRevealer.StartAvatarRevealAnimation(playParticles && !isInsideHideArea, linkedCt);
+            await avatarRevealer.StartAvatarRevealAnimation(playParticles, linkedCt);
```
AjimenezDCL, line 61: Let's rename `playParticles` to `withTransition` (or `instant` and negate it)

AjimenezDCL: Let's also add a comment to clarify that `StartAvatarRevealAnimation`  completes the transition when canceled.

AjimenezDCL (approved):
Approving with two small comments, good job!

PR #2728, opened 2022-07-26
refactor: ECS7 component manager

pbosio:
*refactor ecs7 component manager to handle only one instance instead of having an instance per scene.
-removed `ECSComponentsManager` and `ECSComponentsFactory` references from `DataStore`
+add `BaseList` class for triggering add/remove events from a list.
*`CRDTExecutor` is now instantiated by `ECS7Plugin` instead of being instantiated on `ParcelScene.Awake`
*modified tests accordingly.


pravusjif (approved):
Checked code and tested build and had no problems

PR #2731, opened 2022-07-27
spike: physics based scene bounds check - mesh colliders

pravusjif:
Ugly refactor to use MeshColliders as boundaries check triggers instead of box colliders based on merged bounds.

This alternative implementation was done due to the problems found [here](https://github.com/decentraland/unity-renderer/issues/2433#issuecomment-1195476203).

This implementation fixes those problems, but introduces other heavy isues like most Genesis Plaza meshes failing loading because of errors when trying to use their meshes for their concave mesh collider:
Screen Shot 2022-07-27 at 00 58 18

Besides that, it appears to work fine on simpler meshes like the ones we have in Ropsten 64,-64.


pravusjif:
Closing this PR as defined in https://github.com/decentraland/unity-renderer/issues/2433#issuecomment-1197485502

PR #2748, opened 2022-07-27
feat: create channels window

lorux0:
## What does this PR change?

Adds the posibility to create channels through a window.

## How to test the changes?

1. Open the conversation list
2. Click on the `+` button
3. It should open a channel creation window
4. Write nearby, it should throw an error
5. Write a channel that already exists, it should throw an error
6. Check that the max limit of the name of the channel is `17`
7. Clear the channel name text, the button should be disabled
8. Click the create button
9. The channel window should be opened
10. Check that the channel has been added in the conversation list
11. Go to the channel browser and click the create button on the top of the window, repeat process
12. Go to the channel browser, search for any channel. There should be a create option displayed at the bottom of the list. Click it.
13. Modify the `ChatChannelsControllerMock.GetFakeChannels` method to return 0 channels. Go to the channel browser and check that a new option to create a channel is added. Click it.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/ChatController/Channels/ChatChannelsControllerMock.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/ChatController/Channels/ChatChannelsControllerMock.cs
@@ -312,6 +312,26 @@ namespace DCL.Chat.Channels
         public int GetAllocatedUnseenMessages(string userId) => controller.GetAllocatedUnseenMessages(userId);
 
         public int GetAllocatedUnseenChannelMessages(string channelId) => controller.GetAllocatedUnseenChannelMessages(channelId);
+        
+        public void CreateChannel(string channelId)
+        {
+            SimulateDelayedResponseFor_CreateChannel(channelId).Forget();
+        }
+
+        private async UniTask SimulateDelayedResponseFor_CreateChannel(string channelId)
+        {
+            await UniTask.Delay(Random.Range(40, 1000));
+            
+            controller.JoinChannelConfirmation(JsonUtility.ToJson(new ChannelInfoPayload
+            {
+                description = "",
+                channelId = channelId,
+                joined = true,
+                memberCount = 1,
+                muted = false,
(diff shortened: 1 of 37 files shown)


sandrade-dcl:
The creation modal let me create different `nearby` channels using using upper and lower case letters.

image
image

I think it shouldn't be case sensitive.

sandrade-dcl:
I think we should add a semitransparent background (like we have for the rest of modals) that close the modal in case the user clicks out of the window. Now, when the user clicks on the background, the mouse is locked and the modal window keeps opened.

image

sandrade-dcl:
I think you're not using the default buttons from our UIComponents library, because the visual aspect is not the same as the resto of buttons we have and also the click sounds are missed.

image

sandrade-dcl (approved):

lorux0:
> The creation modal let me create different `nearby` channels using using upper and lower case letters.
> 
> image image
> 
> I think it shouldn't be case sensitive.

done!

lorux0:
> I think we should add a semitransparent background (like we have for the rest of modals) that close the modal in case the user clicks out of the window. Now, when the user clicks on the background, the mouse is locked and the modal window keeps opened.
> 
> image

done!

lorux0:
> I think you're not using the default buttons from our UIComponents library, because the visual aspect is not the same as the resto of buttons we have and also the click sounds are missed.
> 
> image

i added the sfx component but the button prefabs has many components that i dont need.. the background used is the same

sandrade-dcl:
> > I think you're not using the default buttons from our UIComponents library, because the visual aspect is not the same as the resto of buttons we have and also the click sounds are missed.
> > image
> 
> i added the sfx component but the button prefabs has many components that i dont need.. the background used is the same

You can deactivate all the components that you don't need, but it's important to use the prefabs we have in our UI Components library because if we change the visual aspect in the future, the idea was to modify only the prefabs in the library. In this case, imagine that in the future Romi wants to change the red background color by another one... as you have not used the prefab from the library, this "Create Channel" prefab will continue having the red color.

PR #2752, opened 2022-07-28
fix: tutorial camera canvas issues

AjimenezDCL:
Tutorial is using lot of World Position references to anchor UI, which translates to issues when having different canvas modes.

It should be fixed properly using screen position and doing a conversion instead of matching world position but let's not block the release with this.

# Test
Do the tutorial, it should work fine
https://play.decentraland.zone/?renderer-branch=fix/some-camera-canvases-issues&RESET_TUTORIAL


pravusjif (approved):
Nice job

PR #2761, opened 2022-08-01
chore: add memory metrics to ab

### LINKED ISSUE #2124: Make memory metrics work with asset bundles
Memory metrics scoring system for stuff like Animations/Mehses only work in GLTF flow, we should implement it for ABs as well

Acceptance criteria:
- [ ] Scene reports the same memory scoring when loaded with and without ABs

AjimenezDCL:
Fixes #2124 

Added metrics generation file for ABs to have memory scoring for GLTFs and Animation Clips.

# Test
Cannot really be tested, just do a smoke run with and without Asset Bundles, everything should be fine.

### DIFF
--- a/unity-renderer/Assets/ABConverter/Core.cs
+++ b/unity-renderer/Assets/ABConverter/Core.cs
@@ -15,6 +15,7 @@ using UnityGLTF;
 using UnityGLTF.Cache;
 using GLTF;
 using GLTF.Schema;
+using Newtonsoft.Json;
 using UnityGLTF.Loader;
 
 namespace DCL.ABConverter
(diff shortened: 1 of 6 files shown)


unity-renderer/Assets/ABConverter/Core.cs
```diff
@@ -156,6 +157,14 @@ void UpdateLoop()
                             
                             if (value.IsCompleted)
                             {
+                                FileInfo gltfFilePath = new FileInfo(key);
+                                var metrics = new Dictionary<string, object>()
+                                {
+                                    { "meshes_size", value.meshesEstimatedSize },
+                                    { "animation_size", value.animationsEstimatedSize }
+                                };
+                                File.WriteAllText($"{gltfFilePath.Directory.FullName}/metrics.json", JsonConvert.SerializeObject(metrics, Formatting.Indented));
```
pravusjif, line 166: Do we need this to be a Dictionary?

Wouldn't this better be a `new AssetBundleMetrics  { meshes_size = value.meshesEstimatedSize, animation_size = value.animationsEstimatedSize };` and then just serialize it with `JsonUtility.ToJson()` as we do for the metadata here?
https://github.com/decentraland/unity-renderer/blob/701f1c3b9aada74b68bfe958160bb697c7572541/unity-renderer/Assets/ABConverter/AssetBundleMetadataBuilder.cs#L26

pravusjif (approved):
Looks good, I guess we'll test this once we have the ABConverter up to date.

davidejensen (approved):

unity-renderer/Assets/ABConverter/Core.cs
AjimenezDCL: Oh my... I forgot to port that code, nice catch!

PR #2763, opened 2022-08-01
fix: emotes panel crash

dalkia:
## Our Code Review Standards


dalkia:
It served its purposes of validating the theory of the `CancelationRequest`. Closing since this should go into `dev`

PR #2764, opened 2022-08-01
refactor: `playerWorldPosition` base variable

pbosio:
-deleted `CommonScriptableObjects.playerWorldPosition`
*renamed DataStore's `playerWorldPosition` to `playerGridPosition`
+added DataStore's `playerWorldPosition`
*changing value of DataStore's `lastTeleportPosition` will trigger `DCLCharacterController.Teleport`


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
```diff
@@ -196,11 +201,11 @@ public void SetPosition(Vector3 newPosition)
         Environment.i.platform.physicsSyncController?.MarkDirty();
 
         CommonScriptableObjects.playerUnityPosition.Set(characterPosition.unityPosition);
-        CommonScriptableObjects.playerWorldPosition.Set(characterPosition.worldPosition);
+        dataStorePlayer.playerWorldPosition.Set(characterPosition.worldPosition, notifyEvent: false);
```
pbosio, line 204: @AjimenezDCL do you have a strong opinion in not notifying event on this change?

AjimenezDCL: It sounds good but then we don't need a `BaseVariable<Vector3>`, just a `Vector3`

AjimenezDCL (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs
pbosio: But I'm subscribing to the `OnChange` here: https://github.com/decentraland/unity-renderer/blob/20b1eb3e2b948bbfae72db084578bbf1e6a193a9/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs#L147
to force character to teleport when the variable is changed from outside `DCLCharacterContoller` which I'm not sure cause it's kind of odd for the event to be "sometimes" triggered.
Other approach I though that it might be better is using `lastTeleportPosition` variable instead
and here: https://github.com/decentraland/unity-renderer/blob/20b1eb3e2b948bbfae72db084578bbf1e6a193a9/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/DCLCharacterController.cs#L243
doing something like this
```
dataStorePlayer.lastTeleportPosition.OnChange -= Teleport;
dataStorePlayer.lastTeleportPosition.Set(newPosition, true);
dataStorePlayer.lastTeleportPosition.OnChange += Teleport;
```

PR #2765, opened 2022-08-01
feat: load animation on asset reuse

dalkia:
## What does this PR change?

Adds the MaterialTransitionController to all objects on reuse. It applies the change both to AB as well as GLTF.

## How to test the changes?

AB
1. Open https://play.decentraland.zone/?renderer-branch=feat/load-animation-on-reuse
2. Go running to -50,0. See along the way if objects are loaded with a transition effect instead of popping in.

GLTF
1. Open https://play.decentraland.zone/?renderer-branch=feat/load-animation-on-reuse&DISABLE_ASSET_BUNDLES&DISABLE_WEARABLE_ASSET_BUNDLES
2. Go running to -50,0. See along the way if objects are loaded with a transition effect instead of popping in.

## Our Code Review Standards


dalkia:
Performance test. Running from Genesis Plaza to Salomonicon for 60 seconds

**AB with no animation on**
image

**AB with animation on**
image

dalkia:
@AjimenezDCL Having a delay on the reuse of the AB and GLTF generated discrepancies with the unloading process. 

There were scenarios (especially running from 0,0 through the tram scene to -50,0) in which unloading of objects was called before the MaterialTransitionController had ended. MaterialTransitionController continues the loading process after it has finished; but if an unload was called before it had finished, an error occurred and no other elements were loaded. That why I had to add certain checks that allowed me to identify and stop the continuation of the pipeline if unload was called.

What do you think? Is this good enough? If not, I can recheck the loading and unloading pipeline and see if I can regulate it there.

PR #2774, opened 2022-08-03
fix: Remove "Look rotation" warning

dalkia:
## What does this PR change?

Changes the default vector for the avatar rotation after teleport. It was set to `Vector3.zero`, which caused Unity to spam a warning into the Console. Now it's set to a default value of `Vector.forward`; so the default-looking direction for every scene will be North.

## How to test the changes?

1. Go to 102,-32. Check that your avatar started looking **North**, and no _Look rotation viewing vector is zero_ warning is logged into the console. 
2.  Go to -74,53. Check that your avatar started looking **West**, and no _Look rotation viewing vector is zero_ warning is logged into the console. Move around a little bit to change the character starting looking direction
1. Go back to 102,-32. Check that your avatar started looking **North**, and no _Look rotation viewing vector is zero_ warning is logged into the console. 

## Our Code Review Standards


sandrade-dcl (approved):
Reviewed and tested... it is working fine! 👍

kwssbocian (approved):
Approving PR as per comment above

PR #2776, opened 2022-08-03
fix: invisible text shapes not being enabled due a race condition

### LINKED ISSUE #2777: TextShapes are hidden on certain conditions
This issue was reported by DG.

Upon investigation we discovered that this was caused by a race condition.

Kinerius:
## What does this PR change?

Fixes a race condition by polling the text shape properties to see if the component should be enabled.
The polling itself is pretty cheap, it's actually faster than before because we already had an Update method with billboard logic, and now the shape will register itself to the UpdateEventHandler to avoid the MonoBehaviour Update method.

image

Fixes #2777

## How to test the changes?

- [Open this link](https://play.decentraland.org/?renderer-branch=fix%2Finvisible-textshape&realm=dg&position=-120%2C135)
- Go upstairs and wait for the poker tables to load
- Get close to a poker table and interact with a floating card, this should make a popup appear, all text should be visible (the green button used to have no text before this fix, but it was caused by a race condition so its possible for you to not be able to reproduce the old bug)

## Our Code Review Standards


sandrade-dcl (approved):
LGTM! I have tested in the casino scene and everything seems to be working fine with the texts 👍

PR #2756, opened 2022-07-29
feat: UI optimizations

Kinerius:
## What does this PR change?

- Removes the refresh polling that degrades the performance when there are a lot of UIShapes, in its place now there's a unique coroutine that checks if there's a dirty UIShape on its queue.
- Removed a constant Canvas Rebuild caused by a TextShape being updated every frame with no changes on its data.

Fixes #2711

### Old ( ICE Poker - The Stronghold ) [No Avatars]
```
 * PERFORMANCE SCORE (0-100) -> 64
 * lowest frame time -> 38.9ms
 * average frame time -> 45.8ms
 * highest frame time -> 100.0ms
 * 1 percentile frame time -> 39.3ms
 * 50 percentile frame time -> 44.2ms
 * 99 percentile frame time -> 100.0ms
 * error percentage -> ±1.5%
 * total hiccups (>0.05ms frames) -> 44 (6.72% of frames were hiccups)
 * total hiccups time (seconds) -> 3.076782
 * total frames -> 655
 * total frames time (seconds) -> 30.00663
```
### New

```
 * PERFORMANCE SCORE (0-100) -> 69
 * lowest frame time -> 35.3ms
 * average frame time -> 40.6ms
 * highest frame time -> 100.0ms
 * 1 percentile frame time -> 36.2ms
 * 50 percentile frame time -> 40.0ms
 * 99 percentile frame time -> 62.8ms
 * error percentage -> ±1.0%
 * total hiccups (>0.05ms frames) -> 14 (1.89% of frames were hiccups)
 * total hiccups time (seconds) -> 0.9778976
 * total frames -> 740
 * total frames time (seconds) -> 30.02143
```

## How to test

- Go to a scene with lots of UI elements like ICE Poker (-100,126) and make sure that they are correctly aligned when they get updated.
- Go to the same scene and before the UI loads, run outside the building and wait for all the tables to load, then enter the scene, the UI should appear without a massive hiccup

## Our Code Review Standards

https://github.com/decentraland/unity-renderer/blob/master/docs/code-review-standards.md

PR #2786, opened 2022-08-03
re-revert: uishape polling, with extra fix

Kinerius:
## What does this PR change?

This PR re-reverts #2756

But also contains #2788 once its merged

# How to test

Go to [ICE Poker](https://play.decentraland.org/?renderer-branch=fix/uishape-regression&realm=dg&position=-100%2C126) be sure to connect with a wallet.

The loading of the scene should be as smooth as it is in production, when everything is loaded finally, it should be faster ( a bit )

## Our Code Review Standards


kwspgoliasz (approved):
Issue appears to be fixed.
The objects in the casino and the UI are loading noticeably faster.
The overall FPS near the tables when everything is loaded also seems to reach slightly higher values.

PR #2793, opened 2022-08-04
fix: insuficient resources crash when creating avatar for the first time

### LINKED ISSUE #2773: INSUFICIENT_RESOURCES error when entering as a guest for the first time
After you create your avatar and assign a name, the loading screen is frozen. The console throws hundreds of errors, probably caused by a looping issue produced by updating the profile.
Part of the issue is caused by kernel tracking exceptions thrown inside the renderer. When kernel sends the update profile message, the renderer starts a flow that make use of Tasks and CancellationTokens. Whenever we intentionally cancel one, an OperationCancelledException is thrown.

Repro steps:
1. Clear all cookies
2. Enter as a guest
3. Setup your avatar and name
4. The loading freezes at 0%

lorux0:
## What does this PR change?

`Fixes #2773 `

Fixes a loading freeze after you create your own avatar for the first time (guest or wallet).
This was caused by the emote list. Every emote was triggering the AvatarPreviewController to update many times in a row. Every update was being cancelled by throwing a `CanceledOperationException` catch by kernel and then flooding the analytics with it.

## How to test the changes?

1. Clear cookies so you enter as a new user
2. Open the developer console (F12 on chrome)
3. Enter as a guest
4. Start the avatar creation flow
5. Modify your avatar
6. Apply the changes
7. There should not be a flooding message related to `LoadProfile`
8. Set a name and accept the terms and conditions
9. The loading screen should not freeze at 0% and you should be able to enter the world
10. Inside the world go to the avatar editor, modify it and apply the changes. Everything should work as expected

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/DCLPlugins/BuilderInWorld/Scripts/NFTs/BIWNFTController.cs
+++ b/unity-renderer/Assets/DCLPlugins/BuilderInWorld/Scripts/NFTs/BIWNFTController.cs
@@ -45,7 +45,7 @@ public class BIWNFTController
         if (!isInit)
         {
             userProfile = UserProfile.GetOwnUserProfile();
-            // userProfile.OnUpdate += UserProfileUpdated;
+            userProfile.OnUpdate += UserProfileUpdated;
             FetchNftsFromOwner();
             isInit = true;
         }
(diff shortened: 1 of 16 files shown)


menduz:
Tested and working on my end!

pravusjif (approved):
Awsome job, it looks good to me.

I tested comparing with production and this fix effectively removes the crash we can trigger when entering the world for the first time.

However, something to take into account, the errors that crash my console in production during the world initialization (that are not being triggered with this fix) are still appearing during the avatar creation and after saving ANY change for any wearable after having the avatar in-world.
That also happens in production so I guess it's a problem for another day.

Screen Shot 2022-08-04 at 17 16 03

menduz:
I removed the one-by-one basis to report these setThrew to use an aggregated metric here: https://github.com/decentraland/kernel/pull/400 to prevent abusing the network resources. The errors are still being reported but in an aggregated way

PR #2802, opened 2022-08-05
fix: changes noted from review 06082022

lorux0:
## What does this PR change?

- The notification count from friend requests tab is showing the received friend request count.
- GetFriends & GetFriendRequests are only called once
- Replaced general channel id for nearby

## How to test the changes?

1. Set the custom test url: https://play.decentraland.zone/?kernel-branch=feat/add-friends-lazy-load&
2. Open friends and check the number on the requests tab, it should be the same as the taskbar
3. Check that nearby channel keeps working as expected and is listed correctly
4. Check that friends are loaded correctly

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/FriendsHUD/Scripts/FriendsHUDController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/FriendsHUD/Scripts/FriendsHUDController.cs
@@ -21,6 +21,8 @@ public class FriendsHUDController : IHUD
 
     private UserProfile ownUserProfile;
     private bool searchingFriends;
+    private bool isFriendsLoadedForFirstTime;
+    private bool isFriendsRequestsLoadedForFirstTime;
 
     public IFriendsHUDComponentView View { get; private set; }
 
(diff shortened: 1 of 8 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/FriendsHUD/Scripts/FriendsHUDController.cs
```diff
@@ -402,6 +405,7 @@ private void DisplayMoreFriends()
         if (!friendsController.IsInitialized) return;
         ShowOrHideMoreFriendsToLoadHint();
         friendsController.GetFriends(LOAD_FRIENDS_ON_DEMAND_COUNT, View.FriendCount);
+        isFriendsLoadedForFirstTime = true;
```
sandrade-dcl, line 408: Wouldn't be cleaner to put the `isFriendsLoadedForFirstTime` condition at the beginning of the `DisplayMoreFriends()` function instead of doing it before all the calls to `DisplayMoreFriends()`? Something like this:

```
DisplayMoreFriends()
{
    if (isFriendsLoadedForFirstTime) return

    ...

    isFriendsLoadedForFirstTime = true;
}
```

sandrade-dcl (approved):
LGTM! I only left a suggestion 👍

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/FriendsHUD/Scripts/FriendsHUDController.cs
lorux0: not all the flows aligns with this condition.. for example when you are loading more friends while scrolling down

PR #2804, opened 2022-08-06
fix: ECS7 parenting and Transform removal

pbosio:
TL;DR
```
context:
*there is no entity removal on ECS7, just component removal.
*removing all component of a parent entity does not mean removing it children entities
*removing a transform from an entity does not mean the entity won't exist anymore
therefor:
*when a transform of an entity is removed it will implicitly have a transform (Transform.identity)
*when a transform of a parent entity is removed it children will consider parent transform as Transform.identity
```
More in-depth explanation here: https://github.com/decentraland/sdk/issues/394


unity-renderer/Assets/DCLPlugins/ECS7/TestsUtils/ECS7TestUtilsScenesAndEntities.cs
```diff
@@ -0,0 +1,129 @@
...
+        go.transform.ResetLocalTRS();
+
+        // properties
+        entity.entityId = entityId;
+        entity.childrenId = new List<long>();
+        entity.gameObject = go;
+        entity.parentId = 0;
+
+        // entity.OnRemoved
+        // entity.parentId
```
Suduck, line 106: remove commented code

Suduck (approved):

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Transform/Handler/ECSTransformHandler.cs
```diff
@@ -22,12 +22,43 @@ public ECSTransformHandler(IWorldState worldState, BaseVariable<Vector3> playerT
...
-            ECSTransformUtils.orphanEntities.Remove(entity);
+
+            // if entity has any children
+            int childrenCount = entity.childrenId.Count;
+            if (childrenCount > 0)
+            {
+                for (int i = childrenCount - 1; i >= 0; i--)
+                {
+                    long childId = entity.childrenId[i];
+                    entity.childrenId.RemoveAt(i);
```
pravusjif, line 50: Here we are affecting the same collection that is being traversed, doesn't this bring any problem?

Maybe it's better to avoid doing the `RemoveAt()` here and just do `entity.childrenId.Clean()` after the `for`?

pravusjif (approved):

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Transform/Handler/ECSTransformHandler.cs
pbosio: affecting the collection shouldn't matter since the list is being iterated backwards.
but you are totally right regarding performance, since it will be better just to clear the list after the for.
changed here :https://github.com/decentraland/unity-renderer/pull/2804/commits/6631a1cdfb212e14d4c5c688493442f8f3a3f715

PR #2811, opened 2022-08-09
fix: collider bounds in scene bounds check

pravusjif:
### WHY
There's a scene at [90,-143](https://play.decentraland.org/?position=90%2C-143) that has an entity with huge mesh outside the scene boundaries and that mesh doesn't have any renderers, it only has colliders.

We weren't taking colliders into consideration for the entities merged bounds calculation, and when there were no renderers in the entity we weren't affecting its colliders (since the entity was treated as an entity without a mesh).

### WHAT
- Added collider bounds to entity merged bounds calculation and scene bounds checks
- Added Scene boundaries check optimization: escape unneeded checks after outer bounds checks on entity addition
- Changed `MeshesInfo.colliders` to be a `HashSet<Collider>` instead of a `List<Collider>`
- Improved debug method in parcel scene
- Added scene boundaries checker test for GLTFshapes with only colliders and no renderers
- Added `DCLTransform` values capping
- Updated `ParametrizedShape` visibility configuration to take into account entitie's inside-scene-boundaries state 
- Updated `TextShape` visibility configuration to take into account entitie's inside-scene-boundaries state
- Added red wireframe effect to colliders outside scene boundaries for preview mode + scene boundaries checker debug mode test

### TESTING
1. Check that the reported issue got solved:
In [production](https://play.decentraland.org/?position=90%2C-143) there are invisible colliders on the empty parcels at the left of the affected scene (90, -143).
In [this PR's build](https://play.decentraland.zone/?renderer-branch=fix/ColliderBoundsInSceneBoundsCheck&position=90%2C-143) those colliders are disabled as they are detected as part of the mesh of the entity that contains them and that is detected as outside the scene boundaries.

2. Check that the already-solved https://github.com/decentraland/unity-renderer/issues/2132 issue hasn't reappeared:
    1. Spawn [at -12, -10](https://play.decentraland.zone/?renderer-branch=fix/ColliderBoundsInSceneBoundsCheck&position=-12%2C-10) and wait until Genesis Plaza finishes loading (make sure the Scene Load Radius value in the settings is 4) 
    2. Open Settings, change the Scene Load Radius to 1 and wait 5~ seconds
    3. Change Scene Load Radius to 4 and close the settings panel, wait for 5~10 seconds after the Genesis Plaza finishes loading and notice that he "black and white" bug doesn't happen anymore
    4. Steps 2 and 3 can be reproduced many times and the black and white bug won't happen again 
[Doing those steps from master's build will trigger the black and white bug]

3. Check that the already-solved https://github.com/decentraland/unity-renderer/issues/844 issue hasn't reappeared:
    1. Spawn [at -12, -10](https://play.decentraland.zone/?renderer-branch=fix/ColliderBoundsInSceneBoundsCheck&position=-12%2C-10)
    2. Walk towards the tramline at -12, 0
    3. Walk through the tramline to the west, and notice that no mesh is instantiated int the middle of the street


kwspgoliasz (changes requested):
The issue with the collision on empty parcels is fixed, however, two out of the three older issues mentioned in the description seem to have returned.
The aforementioned issues are: 

1. **The black and white bug when changing the scene load radius:**  
issue can be reproduced by following the steps from the ticket's Testing point 2 section. Sometimes it may take two or three attempts for the issue to occur - https://user-images.githubusercontent.com/110464313/184110948-fbc88555-f293-489b-b7ea-0293c2341fbb.mp4 .

2. **The issue with the Rainbow Road bucket appearing temporarily on the player's head:** 
for the issue to occur, the player needs to spawn at -40,1 and reach the Rainbow Road scene. When the area loads, the player needs to turn back until the area disappears. After the area reloads for the second time, a bucket will spawn on the player's head - https://user-images.githubusercontent.com/110464313/184110988-79c383fb-c3b1-4300-b85b-7c1da28ddb72.mp4 .

pravusjif:
@kwspgoliasz can you test again but in this other test PR I just made with a minor change done to this branch please?
https://github.com/decentraland/unity-renderer/pull/2830

Once the CI finishes

kwspgoliasz:
@pravusjif No problem, I'll test on the other branch once the CL finishes. I was testing it on Windows. I had all of my graphical setting set to ultra the entire time. Attaching a screenshot of the settings: Screenshot 2022-08-11 143346.

pravusjif:
@kwspgoliasz those 2 problems are already happening in production, so it looks like the [previously released PR](https://github.com/decentraland/unity-renderer/issues/2132) didn't actually fix them =(, and this current PR doesn't introduce those problems.

#### Black and white bug happening in prod
I had forgotten that his only happens with ULTRA graphics settings.

https://user-images.githubusercontent.com/1031741/184587291-ef352222-7215-472a-9353-ceca537693a0.mp4

#### Object attached to player happening outside scene boundaries
This is not happening the first time the culprit scene is loaded, it happens when the scene is re-loaded during the same sessions, so it must have something to do with the asset library asset being reused.

https://user-images.githubusercontent.com/1031741/184587542-c0074d41-9f5f-493d-a5d1-ec33b9ab0a55.mp4

So based on that, I'll see if I can find a way of fixing those in this same PR, otherwise we can merge this collider-out-of-scene-boundaries fix PR and tackle those issues in another one.

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Transform/DCLTransform.cs
```diff
@@ -67,5 +70,61 @@ public void UpdateFromModel(BaseModel model)
         public BaseModel GetModel() => DCLTransform.model;
         public int GetClassId() => (int) CLASS_ID_COMPONENT.TRANSFORM;
         public void UpdateOutOfBoundariesState(bool enable) { }
+        
+        public static void CapTransformGlobalValuesToMax(Transform transform)
```
lorux0, line 74: is it necessary to be static? could be an instance method? on the other hand it could be refactor into an extension method: `public static void ClampToMaxGlobalValues(this Transform transform)` so the call would cleaner: `transform.ClampToMaxGlobalValues()`

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Transform/DCLTransform.cs
```diff
@@ -67,5 +70,61 @@ public void UpdateFromModel(BaseModel model)
...
+
+                    if (Mathf.Abs(newScale.z) > MAX_TRANSFORM_VALUE)
+                        newScale.z = MAX_TRANSFORM_VALUE * Mathf.Sign(newScale.z);
+                }
+                
+                SetTransformGlobalValues(transform, newPosition, transform.rotation, newScale, scaleOutsideBoundaries);
+            }
+        }
+        
+        public static void SetTransformGlobalValues(Transform transform, Vector3 newPos, Quaternion newRot, Vector3 newScale, bool setScale = true)
```
lorux0, line 111: is it necessary to be static? could be an instance method? in the other hand it could be an extension method so the call is cleaner

unity-renderer/Assets/Scripts/Tests/TestUtils.cs
```diff
@@ -858,9 +858,9 @@ public static IEnumerator TestShapeCollision(BaseShape shapeComponent, BaseShape
             yield return shapeComponent.routine;
 
             // check colliders correct behaviour
-            for (int i = 0; i < entity.meshesInfo.colliders.Count; i++)
+            foreach (Collider collider in entity.meshesInfo.colliders)
```
lorux0, line 861: i would consider the fact that `foreach` introduces a GC issue compared to standard `for`.. if this method is not frequently called, then ignore the comment

lorux0 (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Transform/DCLTransform.cs
pravusjif: Good idea! I've moved that method [into Utils](https://github.com/decentraland/unity-renderer/blob/f84254771fc59f64c5aca9791a4b2e032e203b85/unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Utils.cs#L93) next to the commonly used [`ResetLocalTRS()`](https://github.com/decentraland/unity-renderer/blob/f84254771fc59f64c5aca9791a4b2e032e203b85/unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Utils.cs#L149)

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Transform/DCLTransform.cs
pravusjif: I moved that one into Utils as well, since it's needed by `CapGlobalValuesToMax`

unity-renderer/Assets/Scripts/Tests/TestUtils.cs
pravusjif: The GC problems that the older MONO compiler used in Unity had, have been solved since long time ago, some sources [here](https://answers.unity.com/questions/1359179/does-unity-fixed-foreach-problem-gc-boxing.html) and [here](https://www.jacksondunstan.com/articles/3805). 

Regardless of that, since that collection traversing is happening in a Test environment I think we'd be safe anyway

pravusjif:
- I re-opened [the issue](https://github.com/decentraland/unity-renderer/issues/978) regarding assets attached to the player being rendered outside their scene boundaries
- I created [this other issue](https://github.com/decentraland/unity-renderer/issues/2886) to some day in the future tackle the needed refactor to fix that, after discovering the problem

@kwspgoliasz I updated the description, can you please re-test to give the needed approve to merge? thanks!

pravusjif:
@kwspgoliasz can you re-test please? I updated the PR description and test steps.

kwspgoliasz (approved):
The main issue appears to be fixed, as there are no obsolete collisions present in the specified scene. The other two issues mentioned in the ticket could not be reproduced after the new fixes.

lorux0 (approved):

PR #2817, opened 2022-08-09
fix: conversation list empty state when searching

lorux0:
## What does this PR change?

Correctly updates the empty state container when searching conversations.

## How to test the changes?

1. Open the conversation list (click on chat button, from nearby click back)
2. Search for any text
3. If no results are retrieved then a label should be shown indicating you have no results
4. If results are found, the label should not be shown, only the result entries (direct conversations or nearby channel)

## Our Code Review Standards


sandrade-dcl (approved):

kwspgoliasz (approved):
When no chats are found during the search, a 'No results found' message is displayed below the search bar.

PR #2826, opened 2022-08-10
fix: added payloads for reporting scenes and players

lorux0:
## What does this PR change?

Adds a data structure for the payloads when reporting players and scenes.
The reported scene id has been replaced for coordinates: `x,y`

## How to test the changes?

1. You cannot report a scene with the current ui
2. Click over other player, then open the passport and click the report button

## Our Code Review Standards


aixaCode (approved):

kwspgoliasz (approved):
Feature appears to function accordingly with the design.
Reporting a player redirects the reporter to a form in which the player's name section is already filled in with an ID number.

PR #2828, opened 2022-08-10
fix: avatar animator legacy overhead

### LINKED ISSUE #2714: AvatarAnimatorLegacy Update is slow
We have to rethink the way we are handling the logic of that update.

- [ ] Implement an update manager that handles the update of all the AvatarAnimatorLegacy components to avoid having an Update method on all of them.

- [ ] Avoid getting the name of the animation each frame, it allocates memory and we should avoid it.

- [ ] Check if we can avoid doing a Physics.Raycast , if not, cache the usage of a Ray struct to avoid constructing one since almost all of the raycast is overhead

- [ ] Avoid Null checks every frame

Kinerius:
# What does this PR change?

Fixes the performance overhead of the AvatarAnimatorLegacy, it has zero allocations now.

Fixes #2714 

The performance improvement is 0.7 ms per frame so it's not that much noticeable. But it adds up.

## How to test

Check that running, walking, jumping and falling animate like before.

## Our Code Review Standards


sandrade-dcl (approved):
LGTM!

kwssbocian (approved):
Verified the run, walk, jump and fall animations in various places against those in prod and they appear to be working without issues.

PR #2837, opened 2022-08-11
fix: ECS7 cyclic parenting

pbosio:
Added check to make sure we don't end up with a circular parenting like:
```
entity1-> parent: entity2
entity2-> parent: entity1
```
So we'll ignore the parenting in this scenario
issue: https://github.com/decentraland/sdk/issues/399


unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Transform/Handler/ECSTransformHandler.cs
```diff
@@ -93,6 +93,14 @@ private static void ProcessNewParent(IParcelScene scene, IDCLEntity entity, long
                 }
             }
 
+            //check for cyclic parenting
+            if (ECSTransformUtils.IsCircularParenting(scene, entity, parentId))
+            {
+                Debug.LogError($"cyclic parenting found for entity {entity.entityId} " +
+                               $"parenting to {parentId} at scene {scene.sceneData.id} ({scene.sceneData.basePosition})");
+                return;
+            }
```
pravusjif, line 102: this should be called at the beginning of `private static void ProcessNewParent(IParcelScene scene, IDCLEntity entity, long parentId)` right? otherwise in the recursive parenting situation the entity would lose its previous parent and then be unparented

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Transform/Utils/ECSTransformUtils.cs
```diff
@@ -48,6 +48,33 @@ public static bool TrySetParent(IParcelScene scene, IDCLEntity child, long paren
...
+        {
+            if (!scene.entities.TryGetValue(parentId, out IDCLEntity parent))
+                break;
+
+            if (entity.entityId == parentId)
+            {
+                return true;
+            }
+            parentId = parent.parentId;
+        } while (parentId != SpecialEntityId.SCENE_ROOT_ENTITY);
```
pravusjif, line 73: I think all this logic can probably be removed

pravusjif: In the end I think we need something to cover the case of the target entity trying to be parented to a child entity of its child hierarchy, but this `do while` solution looks more like it's checking the parents instead of the children

pravusjif: OK I just understood how this works: it checks the target parent entity and it traverses from that entiti upwards in the hierarchy checking if the current entity o be parented is in those.

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Transform/Utils/ECSTransformUtils.cs
```diff
@@ -48,6 +48,33 @@ public static bool TrySetParent(IParcelScene scene, IDCLEntity child, long paren
         return false;
     }
 
+    public static bool IsCircularParenting(IParcelScene scene, IDCLEntity entity, long parentId)
```
pravusjif, line 51: We can maybe remove the `scene` parameter since `IDCLEntity` has a `scene` public property

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Transform/Handler/ECSTransformHandler.cs
pbosio: good catch, fixed here: https://github.com/decentraland/unity-renderer/pull/2837/commits/e09ce0c5c3140ae9c74852f70ec4ab1cfd71ff0b

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Transform/Utils/ECSTransformUtils.cs
pbosio: We are keeping scene and entities apart on the ecs7 development, so we can simplify and reduce dependencies for `IParcelScene` and `IDCLEntity` structures

pravusjif (approved):

Issue #2595, opened 2022-07-07
Unexpected backpack behaviours


davidejensen:
The backpack section is presenting different bugs related both to the listing of nft collectibles and also in the character preview, following a video of some of the most encountered issues:

https://user-images.githubusercontent.com/4254522/177723404-925a2af6-2644-43f0-abe9-f03882f63917.mp4

https://user-images.githubusercontent.com/4254522/177723424-6a0308e5-e453-4824-ae0f-651d6c030aa9.mov

To reproduce them just use multiple times the backpack and equip new wearables, close the backpack and repeat

Repro steps:

1. Open the backpack
2. Change your shoes
3. Save
4. Notice your avatar has been corretly modified
5. Open the backpack again (do this step quickly after the previous one)
6. Change tour shoes to another ones
7. Save
8. Notice your avatar has been corretly modified
9. Open the backpack again (do this step quickly after the previous one)
10. Notice your avatar preview has lost all your custom wearables (if you go to the COLLECTIBLES tab, you won't see any wearable loadad there)

![backpack-broken](https://user-images.githubusercontent.com/1031741/181349711-eaf42c7a-86a9-4008-b9cd-bc10e06dd67d.png)

davidejensen:
I managed to break the personal avatar after equipping 1 new wearable in master on unity:
![Screenshot 2022-07-07 at 10 14 51](https://user-images.githubusercontent.com/4254522/177725551-d96c8950-f8ef-474b-b7a7-c2073dd600ca.png)

```
MissingReferenceException: The object of type 'SkinnedMeshRenderer' has been destroyed but you are still trying to access it.
Your script should either check if it is null or you should not destroy the object.
AvatarSystem.AvatarSystemUtils.CopyBones (UnityEngine.SkinnedMeshRenderer source, System.Collections.Generic.IEnumerable`1[T] targets) (at Assets/Scripts/MainScripts/DCL/AvatarSystem/AvatarSystemUtils.cs:51)
AvatarSystem.Loader+<Load>d__21.MoveNext () (at Assets/Scripts/MainScripts/DCL/AvatarSystem/Loader/Loader.cs:90)
--- End of stack trace from previous location where exception was thrown ---
System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw () (at <695d1cc93cca45069c528c15c9fdd749>:0)
[...]

PR #2838, opened 2022-08-11
fix: Unexpected backpack behaviours

sandrade-dcl:
## What does this PR change?
This PR fixes this issue: https://github.com/decentraland/unity-renderer/issues/2595

The backpack was loosing all the COLLECTIBLES wearables each time we did this steps:

1. Open the backpack
2. Change your shoes
3. Save
4. Notice your avatar has been corretly modified
5. Open the backpack again (do this step quickly after the previous one)
6. Change tour shoes to another ones
7. Save
8. Notice your avatar has been corretly modified
9. Open the backpack again (do this step quickly after the previous one)
10. Notice your avatar preview has lost all your custom wearables (if you go to the COLLECTIBLES tab, you won't see any wearable loadad there)

backpack-broken

## How to test the changes?
1. Go to: https://play.decentraland.zone/?renderer-branch=fix/unexpected-backpack-behaviours
2. Reproduce the steps described above.
3. Check the backpack works as expected and our COLLECTIONS are always being loaded correctly.
4. Check also the avatar preview is always being loaded with the correct equipped wearables.

## Our Code Review Standards


pravusjif (approved):
I guess we can merge it since it works for fixing the problem but it looks like a patch and not the fix to the real problem, do we really need to create these 2 lists on demand and traverse them checking if the equipped wearables are the same ones that were the last owned wearables requested?

If we are correctly saving the profile locally and correctly updating the equipped wearables only once when we save the backpack changes then we shouldn't be needing this patch that re-checks everything later...

sandrade-dcl:
> I guess we can merge it since it works for fixing the problem but it looks like a patch and not the fix to the real problem, do we really need to create these 2 lists on demand and traverse them checking if the equipped wearables are the same ones that were the last owned wearables requested?
> 
> If we are correctly saving the profile locally and correctly updating the equipped wearables only once when we save the backpack changes then we shouldn't be needing this patch that re-checks everything later...

Yeah, I agree. We're going to investigate this issue more deeply because it seems we have some inconsistencies with the profile inventory and we shouldn't have to update it manually. Meanwhile, this patch could be useful if we don't find a definitive solution before the next release 👍

lorux0 (approved):
looks good to me!

lorux0:
> I guess we can merge it since it works for fixing the problem but it looks like a patch and not the fix to the real problem, do we really need to create these 2 lists on demand and traverse them checking if the equipped wearables are the same ones that were the last owned wearables requested?
> 
> If we are correctly saving the profile locally and correctly updating the equipped wearables only once when we save the backpack changes then we shouldn't be needing this patch that re-checks everything later...

yeah we decided to remove the `UserProfileModel.inventory` field because kernel was always sending an empty array, and that was triggering the update of an empty inventory which was the root of the problem.. i think this approach removes confusion and the complexity of the current code state

kwssbocian (changes requested):
On my end issue does not appear to be fixed, at least not entirely. Launching the client with the fix/unexpected-backpack-behaviours branch with the wallet login and going to the Collectibles section in the backpack causes the interface to dissapear almost immediately. Issue seems to be fixed for the guest login though as I have observed no issues there.

image

kwssbocian (approved):

sandrade-dcl:
> On my end issue does not appear to be fixed, at least not entirely. Launching the client with the fix/unexpected-backpack-behaviours branch with the wallet login and going to the Collectibles section in the backpack causes the interface to dissapear almost immediately. Issue seems to be fixed for the guest login though as I have observed no issues there.

This is another issue (already happening in prod) that will be managed in a separate ticket 👍 Thanks!

PR #2839, opened 2022-08-11
chore: update debug script

dalkia:
## What does this PR change?

Cleans up and changes the way the debug script is being used.

1. Change URL mode. Now the options are `zone`, `org`, `localhost`, or `custom`. if custom is chosen, a field appears which needs to be filled. This was achieved by adding the DrawIf attribute class.
2. Added Network dropdown. You can choose which network to connect to. For now, we only have `MAINNET` and `ROPSTEN`. `GOERLI` is going to be added in the short term when the testing phase is finished.
3. Removed `ALL_WEARABLES `and `TEST_WEARABLES` since nobody was using them.
4. Added DEBUG_MODE to boolean options. 
5. General code cleanup and organization.

## How to test the changes?

Open the script in Unity, and check that the correct URL is being addressed

## Our Code Review Standards


pravusjif (changes requested):
- Can we have this custom URL as the default one in the fielt that appears after setting the "Base URL Mode" as "CUSTOM"? `https://play.decentraland.zone/?` as we did before this change
- Can we correct the "Open Browser When Start" to "Open Browser On Start"?
- In MacOS the editor never triggers the browser tab and we get this super strange error that I've never seen before:

Screen Shot 2022-08-12 at 03 15 54

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Debugging/DebugParameters/DrawIfAttribute.cs
```diff
@@ -0,0 +1,40 @@
+using UnityEngine;
+using System;
+ 
+/// <summary>
+/// Draws the field/property ONLY if the compared property compared by the comparison type with the value of comparedValue returns true.
+/// Based on: https://forum.unity.com/threads/draw-a-field-only-if-a-condition-is-met.448855/
+/// </summary>
+[AttributeUsage(AttributeTargets.Property | AttributeTargets.Field, AllowMultiple = true)]
+public class DrawIfAttribute : PropertyAttribute
```
aixaCode, line 9: Perhaps we could move this class to a more general location. We can use it in other scripts, as more editor utils. We have already `ExpandableAttribute.cs`, however, it seems that it's not used, we could make a shared assembly for those.

dalkia: Makes sense. I'll make a shared assembly between this class and `ExpandableAttributes.cs`, so all the editor utils are centralized in one place.

aixaCode (approved):

pravusjif (approved):
Latest updates fixed the issue, nice job!

PR #2870, opened 2022-08-17
feat: ECS7 add Material component

pbosio:
Add Material Component for ECS7

overall, applying a material should work as following:
* when creating/updating`BoxShape`, `SphereShape`, `PlaneShape` or `CylinderShape` internal component `InternalTexturizable` is created/updated, which contain a reference to the shapes' renderers
* when creating/updating `Material` component and material is loaded, internal component `InternalMaterial` is created/updated which contains a reference for the loaded material
* `ECSMaterialSystem` will match every entity containing both `InternalTexturizable` and `InternalMaterial` components and apply it changes if any of those components have changed

some relevant changelog:
*refactor `AssetPromise_Material` to handle both pbr and basic material, fixed ref count issues, changed model to readonly struct to avoid hashing issues and others
-deleted `AssetPromise_TextureResource`
+add internal components `InternalTexturizable` and `InternalMaterial`
+add component for `Material`
+add `ECSMaterialSystem` system to apply material to renderers
*modified `BoxShape`, `SphereShape`, `PlaneShape` and `CylinderShape` to create `InternalTexturizable` component on loaded


unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Defaults/PBMaterial_Defaults.cs
```diff
@@ -0,0 +1,84 @@
...
+    public static class PBMaterial_Defaults
+    {
+        private static readonly Color3 colorWhite = new Color3()
+        {
+            R = 0, G = 0, B = 0
+        };
+        private static readonly Color3 colorBlack = new Color3()
+        {
+            R = 1, G = 1, B = 1
+        };
```
menduz, line 12:
```suggestion
        private static readonly Color3 colorWhite = new Color3()
        {
            R = 1, G = 1, B = 1
        };
        private static readonly Color3 colorBlack = new Color3()
        {
            R = 0, G = 0, B = 0
        };
```

pbosio: 🤦‍♀️ thanks

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/PlaneShape/Tests/ECSPlaneShapeShould.cs
```diff
@@ -17,14 +19,16 @@ public class ECSPlaneShapeShould
...
 
         [SetUp]
         protected void SetUp()
         {
             gameObject = new GameObject();
             entity = Substitute.For<IDCLEntity>();
             scene = Substitute.For<IParcelScene>();
-            planeShapeComponentHandler = new ECSPlaneShapeComponentHandler(DataStore.i.ecs7);
+            texurizableInternalComponent = Substitute.For<IInternalECSComponent<InternalTexturizable>>();
+            planeShapeComponentHandler = new ECSPlaneShapeComponentHandler(DataStore.i.ecs7, texurizableInternalComponent);
```
kuruk-mm, line 31:
```suggestion
            planeShapeComponentHandler = new ECSPlaneShapeComponentHandler(DataStore.i.ecs7, texturizableInternalComponent);
```

unity-renderer/Assets/DCLPlugins/ECS7/InternalECSComponents/Interfaces/InternalECSComponentsId.cs
```diff
@@ -0,0 +1,5 @@
+public enum InternalECSComponentsId
+{
+    TEXTURIZABLE = 2,
```
kuruk-mm, line 3: Any reason that is starting with 2 instead of 1?

unity-renderer/Assets/DCLPlugins/ECS7/ProtocolBuffers/Definitions/Material.proto
```diff
@@ -0,0 +1,60 @@
+syntax = "proto3";
+
+import "common/id.proto";
+option (ecs_component_id) = 1017;
```
kuruk-mm, line 4: This can be on `https://github.com/decentraland/protocol`

To test from a PR from `https://github.com/decentraland/protocol`, go to `ProtobufEditor.cs` and look for this line:

```
private const string FIXED_NPM_PACKAGE_LINK = "";
```

you must apply the npm package URL (which the bot from @dcl/protocol provides you when you open a PR)

kuruk-mm (approved):
Looks great! Can't follow the whole logic, but it looks very good isolated in the ECS7 Plugin.

So, move on :)

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/PlaneShape/Tests/ECSPlaneShapeShould.cs
pbosio: thnks! https://github.com/decentraland/unity-renderer/pull/2870/commits/0abbaffa3d064113a0cc99b84dc95523f0e39f01

unity-renderer/Assets/DCLPlugins/ECS7/InternalECSComponents/Interfaces/InternalECSComponentsId.cs
pbosio: we currently have Transform as id 1. maybe we could change transform's id, we should talk about this with lean

unity-renderer/Assets/DCLPlugins/ECS7/ProtocolBuffers/Definitions/Material.proto
pbosio: yeap, I will be adding it to the protocol repo

unity-renderer/Assets/DCLPlugins/ECS7/InternalECSComponents/Interfaces/InternalECSComponentModels.cs
```diff
@@ -0,0 +1,19 @@
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace DCL.ECS7.InternalComponents
+{
+    public class InternalTexturizable
+    {
+        public IList<Renderer> renderers = new List<Renderer>();
```
pravusjif, line 8: If the collection won't ever hold more than 1 instance of the same renderers then it could be a Hashset instead, that is faster

unity-renderer/Assets/DCLPlugins/ECS7/InternalECSComponents/Interfaces/InternalECSComponentsId.cs
pravusjif: Maybe just by adding the `TRANSFORM = 1` in this same enum would be enough to clarify

pravusjif (approved):
LGTM, I also tested the build witout issues

unity-renderer/Assets/DCLPlugins/ECS7/InternalECSComponents/Interfaces/InternalECSComponentModels.cs
pbosio: Hashset is faster for looking and removing elements, but painfully slow when iterating through it elements and it instantiate an enumerator every time is iterated.
Since we can expect more elements iterations than element removals I've decided to go for a List which, while iterating, it's as fast and allocation free as an array

unity-renderer/Assets/DCLPlugins/ECS7/InternalECSComponents/Interfaces/InternalECSComponentsId.cs
pbosio: I could add a comment there if it helps clarifying, but in any case, there is a test here: https://github.com/decentraland/unity-renderer/blob/e48b52edf9fc89059fa6039a0cb2a224613cdce9/unity-renderer/Assets/DCLPlugins/ECS7/InternalECSComponents/Tests/InternalComponentsIdShould.cs
that will prevent ids for colliding

PR #2872, opened 2022-08-17
feat: add first UI integration test for ECS7 

Suduck:
## Context

We don’t have any way to test how different systems interact with each other. We need a way to test it. 
Many systems in the explorer need to interact together ( for example, the OnPointerEvent and GLTFShapes, or UITransform and the CanvasPainter). 

We should be able to test the interaction of different systems with each other to create a consistent behavior between different versions of ECS.

The first approach was to do Visual testing. The problem is that the Unity package UI Toolkit has a bug that is not able to render the UI if you run Unity in batch mode and the visual test fails. We delay the implementation of visual testing for the future when it gets fixed.

The current approach is to send CRDT messages to the scene and test that the visual elements of the UI have the same properties as the model that we send in the CRDT. If the visual element from UI Toolkit has the property set as expected we can trust that the rendering will be as expected. Otherwise, it will probably be a bug in their API

## What does this PR change?

This PR adds integration tests for the `CanvasPainter `class with the component UITransform.

CanvasPainter class is responsible for painting the UI of the ECS7 scenes and we wanted to add some integration tests to ensure that everything is working as expected with the first UI component.

This class already has DI so we can avoid having the full environment up. This way we are detaching the integration test from the runtime

Approach: Create a scene of ECS7 type and send CRDT messages with the components that we want to be in the scene. Then after the messages have been processed. We tell the canvas painter to draw the UI and create the corresponding visual elements. Then we test that the visual elements has the same properties specify in the model of the CRDT messages.

I also have created a helper to create ECS7 components. For the time being,  a new class `ECS7TestUIUtils` is created that will create a default UITransform component for serialization.

## How to test the changes?

Since this is an integration test, the changes can't be tested beside check that the CI passes

## Our Code Review Standards


unity-renderer/Assets/DCLPlugins/ECS7/CanvasPainter/Tests/CanvasPainterUITransformIntegrationTest.cs
```diff
@@ -0,0 +1,333 @@
...
+            canvasPainter.Dispose();
+            DataStore.i.ecs7.scenes.Clear();
+            DataStore.i.ecs7.uiDataContainer = new UIDataContainer();
+        }
+
+        [Test]
+        public void DrawListOfItemCorrectly()
+        {
+            int parentId = 512;
+            int childIdStartIndex = 513;
```
sandrade-dcl, line 60: Curiosity: Have these numbers been choosen randomly? or have they these specific values for any reason?

sandrade-dcl (approved):

lorux0 (approved):

PR #2889, opened 2022-08-19
fix: direct message listing fixes

lorux0:
## What does this PR change?

- removes direct messages entries when the friendship updates
- always allocates the last message received no matter if the profile exists or not
- removed deprecated filters for showing direct messages
- added test coverage

## How to test the changes?

1. Go to the conversation list and check everything is listed correctly
2. Remove any friend that has a direct message with you
3. The friend should be removed from the list
4. Check that the last message is correctly display in the entry, when you receive one

## Our Code Review Standards


sandrade-dcl (approved):
LGTM!

PR #2890, opened 2022-08-22
Feat: added top chat notification

### LINKED ISSUE #2401: Implement top chat notifications
The chat, friends and voice chat panels compete for the same space as the notifications. When these are open, the notification area moves up and narrows in size showing only one entry.

[Documentation Link](https://www.notion.so/decentraland/PRD-Chat-Notifications-open-for-comments-5ab89b9ab5794e1aa8e7759ad3163c1c#1a9a4fa60b7a4de89967eb24a544f84f)

**Acceptance Criteria**
- These notifications have a fixed width, unlike the others. 
- They will also include the number of unread messages from each conversation.
- If the entry comes from a private chat the avatar snapshot, now larger in size, is centered to the left with respect to the name and message.
- You will not receive notifications about currently open conversation (if any).

**Notes**
There is a pending decision on what is going to happen when several notificacions are popping at the same time, we need to iterate on development

davidejensen:
## What does this PR change?

Fix #2401 
Adds top notification

...

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=feat/top-notification&ENABLE_NOTIFICATION_CHAT
2. Play with chat and notification
3. Verify that everything works as expected
4. Verify that when chat or channels are opened a top chat appears
5. Verify that it updates when new notifications arrive
6. Verify that by clicking the notification it opens the correct chat

## Our Code Review Standards


sandrade-dcl:
I have found this visual issue: When I have entered the world, I have received several notifications about old conversations and when I have open the chat window, the topn notification seems to be visually broken:

image

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/NotificationMessagesHUD/NotificationComponent/ChatNotificationMessageComponentView.cs
```diff
@@ -177,7 +190,7 @@ public void SetNotificationTargetId(string notificationTargetId)
         this.notificationTargetId = notificationTargetId;
     }
 
-    private void ForceUIRefresh()
+    public void ForceUIRefresh()
     {
         Utils.ForceRebuildLayoutImmediate(backgroundTransform);
```
sandrade-dcl, line 195: I'm seeing you need to use this `ForceRebuildLayoutImmediate` in several places as each time we show the view or each time a new notification enters... this could happen a lot of times. Have you measured, more or less, the performance impact it has? Be careful with that because I remember the `ForceRebuildLayoutImmediate` could be a bit heavy in terms of performance. Please check if you really need to use it in all the cases where you currently have it!

sandrade-dcl (approved):
The code looks good to me! I only left a concern 👍

sandrade-dcl:
I'm seeing the images are not being well encapsulated into the top notifications:

image

sandrade-dcl:
I cannot click on any top notification notif. My mouse is always locked when I do it.

image

sandrade-dcl:
Have we changed the behavior of the chat window when we click on the world?
I open the chat window (or any DM window).
I click on the world.
Notice the window is not passing to transparent as before.
If I click on the close button in the window, it closes BUT the status in the taskbar is wrong.

image

sandrade-dcl (changes requested):
I detected some visual issues. I have put them as comments in the PR 👍

lorux0:
the scroll is not correctly staying at the bottom when new messages are added

lorux0 (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/NotificationMessagesHUD/NotificationComponent/ChatNotificationMessageComponentView.cs
davidejensen: I managed to remove an useless one, the others are needed but not used much, only at the creation

davidejensen:
> I'm seeing the images are not being well encapsulated into the top notifications. Even knowing that the real ones are going to come rounded, we should always mask them. We don't know how they can be in the future and the UI should be compatible with any type of shape of image.
> 
> image

This is a problem only with the fake messages, this happens because the fake images are squared but real profile pictures are transparent and never get out of the borders, so there is no need of masking that consumes resources

davidejensen:
> I cannot click on any top notification notif. My mouse always become locked when I do it.
> 
> image

Doesn't happen to me, could you share a video of how it happens?

davidejensen:
> the scroll is not correctly staying at the bottom when new messages are added

What do you mean? If you are hovering on the notification list it is supposed to be like that, otherwise if you are not hovering it will go to the bottom

sandrade-dcl:
> > I cannot click on any top notification notif. My mouse always become locked when I do it.
> > image
> 
> Doesn't happen to me, could you share a video of how it happens?

I'm noticing that it is only happenning when we try to click on a notif having the CONVERSATIONS LIST opened. When I do it having a private message opened, it works.

image

sandrade-dcl:
> > I'm seeing the images are not being well encapsulated into the top notifications. Even knowing that the real ones are going to come rounded, we should always mask them. We don't know how they can be in the future and the UI should be compatible with any type of shape of image.
> > image
> 
> This is a problem only with the fake messages, this happens because the fake images are squared but real profile pictures are transparent and never get out of the borders, so there is no need of masking that consumes resources

Still, I think we should make this UI compatible for any kind of shape of image (at the same way we are already doing in the private messages header). We don't know how the pictures could come in the future.

sandrade-dcl (approved):

kwssbocian (approved):
✔️ Whenever a chat or voice chat windows is opened a single notification window above it is present
✔️ Whenever there is a new message that arrives the notification on top appears and is updated with the new message
✔️ Clicking on the notification opens up the correct chat

PR #2897, opened 2022-08-23
fix: ECS7 gltf initial visibility

pbosio:
remove usage of `entity.isInsideSceneBoundaries` for initial visibility in ecs7 gltf component.
we are going to implement scene boundaries checks for ecs7 in the near future after all components needed to achieve feature parity are completed


AjimenezDCL (approved):
Looks good! can we add a TODO near the call?

PR #2900, opened 2022-08-23
feat: ECS7 MeshRenderer component

pbosio:
Add MeshRenderer component which will replace components: BoxShape, PlaneShape, CylinderShape and SphereShape
More info here: https://github.com/decentraland/sdk/issues/366
As with the primitive "Shapes" components we are using a asset promises to create, share and destroy the meshes.
A difference with those components is that we are not adding colliders or visibility properties since those functionalities are going to be added as individual components.

other relevant changelog:
*modify `AssetPromise_PrimitiveMesh` to use a readonly struct to avoid any hashing issue
*added `PrimitiveMeshBuilder.BuildPlaneV2` function so plane is not created upside down as it's currently happening


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/PrimitiveMesh/AssetPromise_PrimitiveMesh_Model.cs
```diff
@@ -0,0 +1,48 @@
...
+        public readonly float radiusTop;
+        public readonly float radiusBottom;
+        public readonly IList<float> uvs;
+
+        public AssetPromise_PrimitiveMesh_Model(PrimitiveType type, IList<float> uvs, float radiusTop, float radiusBottom)
+        {
+            this.type = type;
+            this.uvs = uvs;
+            this.radiusTop = radiusTop;
+            this.radiusBottom = radiusBottom;
```
AjimenezDCL, line 25: Isnt this too specific of the Cylinder?

AjimenezDCL (changes requested):
It might be really confusing to have MeshRenderer and also GLTFShape, I also see violations of the Dependency Inversion Principle, that MeshRenderer is too coupled to a cylinder and Single Responsibility Principle (Mesh Renderer containing also the Mesh Definition).

In my opinion we should increase the abstraction one level, make the MeshRenderer agnostic to the mesh itself and develop that concept properly (in the same way UE or even Unity work or something along these lines).

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/PrimitiveMesh/AssetPromise_PrimitiveMesh_Model.cs
pbosio: that's true. would you prefer we split that model into different structs and extend from a base model struct for those two properties?

> It might be really confusing to have MeshRenderer and also GLTFShape, I also see violations of the Dependency Inversion Principle, that MeshRenderer is too coupled to a cylinder and Single Responsibility Principle (Mesh Renderer containing also the Mesh Definition).
> 
> In my opinion we should increase the abstraction one level, make the MeshRenderer agnostic to the mesh itself and develop that concept properly (in the same way UE or even Unity work or something along these lines).

I look confusing because of the protobuf structure.
In practice is defining a mesh and adding it to a MeshRenderer so it can be renderer.
In this PR we are just enabling meshes for those primitives (box, sphere, planes, cylinder) but adding custom meshes, where the dev define vertices, uvs, etc it's also a possibility in the future.
We are currently discussing the proper way expose that at the end user api level for a better dev-exp while coding scenes.
Regarding GLTShape, we are going to change and rename that component, the new paradigm will be that GLTF wont be a "shape" any more but something closer to a "scene", meaning, a collection of meshes, materials, colliders, etc

PR #2904, opened 2022-08-24
Feat: add home point functionality

### LINKED ISSUE #2021: Allow players to set scene as Home
All players enter decentraland through the same coordinates (0,0) on genesis plaza. As a player, I would like to set the current coordinates as my home, so everything I come back to Decentraland, I start at this place.

**Acceptance criteria**
- [x] Home coordinates are stored locally in ~~player prefs~~ local storage (per computer)
- [x] This action can be executed from the scene actions below the minimap
- [x] Unsetting the scene as home, reset it to 0,0
- [x] This only happens when the url does not contain any coordinate params

**Pending design**
- [ ] HUD interaction to set scene as home (action button, house icon etc)
- [ ] Mark in the minimap displaying home scene
- [ ] Alternative field in the settings to input or reset the home coordinates

davidejensen:
## What does this PR change?

Fix #2021 
Allows player to decide which scene to place as home

...

with the following link: https://play.decentraland.zone/?renderer-branch=feat%2Fhome-point&kernel-branch=feat%2Fsend-home-point&realm=artemis
* Enter decentraland
* Verify that the home point in the map is 0,0
* Go to a random parcel
* Set home there with the three dots menu below the minimap
* Verify that untoggling the set home the home point is set again at 0,0
* Set the home point again
* Open again decentraland after closing it without the position query parameter
* Verify that you spawn in the home point

## Our Code Review Standards


kwssbocian (approved):
✔️ Default home point is set to 0,0 and the marker on the map shows up on that tile.
✔️ Going to any random parcel and setting it as the home point causes it to be designated as the spawn point. Map marker shows up on that tile.
Minor issue with the map marker visibility upon zooming in found and reported as [#2971 ](https://github.com/decentraland/unity-renderer/issues/2971)

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Minimap/MinimapHUDController.cs
```diff
@@ -75,6 +92,11 @@ public void UpdatePlayerPosition(Vector2 position)
         UpdatePlayerPosition(string.Format(format, position.x, position.y));
     }
 
+    public void UpdateSetHomePanel()
+    {
+        view?.UpdateSetHomePanel(currentCoords == homeCoords);
```
lorux0, line 97: can it happen that the view is null?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Minimap/MinimapHUDController.cs
```diff
@@ -102,6 +124,21 @@ public void ReportScene()
         WebInterface.SendReportScene($"{coords.x},{coords.y}");
     }
 
+    public void SetHomeScene(bool isOn)
+    {
+        var coords = playerCoords.Get();
+        if (playerCoords == homeCoords)
+        {
+            if (!isOn)
+                WebInterface.SetHomeScene($"0,0");
```
lorux0, line 133: please instead of referencing `WebInterface` directly, we should inject an interface that sets the home scene, so follow the dependency inversion principle

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Minimap/MinimapHUDController.cs
```diff
@@ -102,6 +124,21 @@ public void ReportScene()
...
+        }
+        else
+        { 
+            if(isOn)
+                WebInterface.SetHomeScene($"{coords.x},{coords.y}");
```
lorux0, line 138: please instead of referencing `WebInterface` directly, we should inject an interface that sets the home scene, so follow the dependency inversion principle

lorux0 (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Minimap/MinimapHUDController.cs
davidejensen: Nope, fixed

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Minimap/MinimapHUDController.cs
davidejensen: I don't get this, didn't we always use the WebInterface static class in this way like I did?

lorux0: you can create a new interface then inject it in `MinimapHUDController`, for example:
```
interface IHomeLocationController
{
    void SetHomeScene(Vector2 location);
}
```

you can have a single implementation:
```
class WebInterfaceHomeLocationController : IHomeLocationController
{
    public void SetHomeScene(Vector2 location)
    {
        WebInterface.SetHomeScene($"{location.x},{location.y}");
    }
}
```

this way we remove the coupling with `WebInterface`, its easier to mock/test and opened to change the way it works in the future without modifying `MinimapHUDController`

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HomePoint/IHomeLocationController.cs
```diff
@@ -0,0 +1,7 @@
+using UnityEngine;
+
+public interface IHomeLocationController
+{
+    void SetHomeScene(string location);
```
lorux0, line 5: I think the interface should not support this overload, because it could lead to bugs since a string could have a wrong format or anything you want. I think we should only support `Vector2` as a parameter to get a more controlled usage of the feature.

davidejensen: alright, will change in a bit :)

lorux0 (approved):
nice work!

PR #2929, opened 2022-08-26
feat: map renderer shader

AjimenezDCL:
Shader based map renderer

https://play.decentraland.zone/?renderer-branch=feat/map-renderer-shader


AjimenezDCL:
Closing this, the shader is outdated and not compatible with our current unity version.

PR #2942, opened 2022-08-30
fix: passport not opening on click in Unity

dalkia:
## What does this PR change?

Changes the way the method ApplyChanges toggles the passport click functionality. Its now separated from the logic of the avatar modifier area, so it can call it without involving the counter relative to the modifier area

## How to test the changes?

1. Try and click on any avatar. The passport should open
2. Teleport to any scene (like Ice Poker). Click on any avatar and see that passport opens.

## Our Code Review Standards


davidejensen (approved):
Awesome, it works!!!

sandrade-dcl (approved):

kwspgoliasz (approved):
The feature appears to be implemented correctly. No issues with opening the passport were observed.

PR #2944, opened 2022-08-31
Fix: fixed broken scene text shape

### LINKED ISSUE #2893: Missing or corrupted letters of ICE Poker's interface buttons
Description:
Letters for the floating action buttons of the ICE Poker's interface are either missing or appear corrupted. Issue appears to occur only for the leftmost button of the top row as well as for the red and blue buttons of the lower row.

Steps to reproduce:
1. Log in to the game using a wallet account with an ICE Enabled wearable.
2. Go to the ICE Poker - Stronghold at -102,129.
3. From the popup that appears choose "Find me a table" and move over to the table designated by the green arrow.
4. Enter the poker game.
5. Observe the poker game's interface buttons.

Results:
Letters on the poker game's buttons are missing or corrupted.

Expected results:
Letters on the poker game's buttons are present and forming words.

Notes:
- Platforms: Windows and Mac web and desktop, production and dev branch.

davidejensen:
## What does this PR change?

Fix #2893 
Removes text shape minor optimization in order to fix the broken texts

...

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=fix/broken-text-shape
2. Go to an ice poker scene
3. Play (ask Kyllian if you need a wearable to test the scene)
4. Check that all the texts appear correctly

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
```diff
@@ -82,14 +82,6 @@ private void OnUpdate()
             {
                 transform.forward = mainCamera.transform.forward;
             }
-            
-            var opacity = cachedModel.visible ? cachedModel.opacity : 0;
-            var totalSize = cachedModel.fontSize;
-            var totalScale = text.transform.lossyScale.sqrMagnitude;
-            bool isVisible = opacity > 0 && totalSize > 0 && totalScale > 0 && entity.isInsideSceneBoundaries;
```
Antivortex, line 89: I think we should be aware that just a little value of 0.0001 or -0.0001 in any of scale components of any parent of text will make totalScale > 0 equal true. Maybe there should be a threshold a little different from 0.

dalkia: Agreed with Anton! If we bring this optimization back, we should add some delta threshold to determine the visibility of the TextShape.

davidejensen: don't think we need to bring the optimization back, if you look at the issue I've posted the benchmark and in ice poker (one of the places with most intense usage of text shapes) there is actually not that big of an impact

PR #2954, opened 2022-09-01
fix: onpointerevent colliders reset on scene bounds check

pravusjif:
### WHY
When an entity with a shape is outside the scene boundaries, its renderers are disabled. If at that moment an OnPointerEvent component is attached to the entity, it will have its onpointer event colliders disabled and they'll never be enabled when the entity re-enters the scene boundaries and gets re-evaluated.

That happens because during onpointer vent colliders initialization, the colliders are initialized as enabled/disabled based on the entity renderers enabled value.

### WHAT
- Implemented 'IOutOfSceneBoundariesHandler' interface in OnPointerEvent to be called when the pertinent scene changes its 'inside scene boundaries' state.
- Added pointer event colliders update based on renderers state after scene boundaries state change event
- Implemented `UnloadAll()` method in old `CommonScriptableObjects` classes
- Added `UnloadAll()` call in `IntegrationTestSuite_Legacy.TearDown()` to clean `CommonScriptableObjects` classes after every test to avoid many tests failing only when run all together (but passing when run as isolated).

### TESTING
1. Enter the Ice Poker Stronghold scene (`.org` url has to be used otherwise the scene doesn't load the poker tables)
2. Approach a poker table and confirm that every floating card has the onpointer event hover feedback and is clickable

#### Confirm **problem** using `dev` build
https://play.decentraland.org/?renderer-branch=dev&position=-100,129&LOS=0

https://user-images.githubusercontent.com/1031741/187985556-41c70df6-5833-460b-9233-ef564bce4292.mp4

#### Confirm **fix** using `fix/onPointerEventCollidersResetOnSceneBoundsCheck` build
https://play.decentraland.org/?renderer-branch=fix/onPointerEventCollidersResetOnSceneBoundsCheck&position=-100,129&LOS=0

https://user-images.githubusercontent.com/1031741/187985650-185884e2-aea2-4aa8-a54a-ec340cfc064f.mp4

### DIFF
--- a/unity-renderer/Assets/DCLPlugins/UUIDEventComponentsPlugin/UUIDComponent/OnPointerEvent.cs
+++ b/unity-renderer/Assets/DCLPlugins/UUIDEventComponentsPlugin/UUIDComponent/OnPointerEvent.cs
@@ -27,6 +27,12 @@ namespace DCL.Components
             this.entity = entity;
             eventColliders.Initialize(entity);
         }
+        
+        public void UpdateCollidersEnabledBasedOnRenderers(IDCLEntity entity)
+        {
+            this.entity = entity;
+            eventColliders.UpdateCollidersEnabledBasedOnRenderers(entity);
+        }
 
         public void SetFeedbackState(bool showFeedback, bool hoverState, string button, string hoverText)
         {
@@ -62,7 +68,7 @@ namespace DCL.Components
         }
     }
 
-    public class OnPointerEvent : UUIDComponent, IPointerInputEvent
+    public class OnPointerEvent : UUIDComponent, IPointerInputEvent, IOutOfSceneBoundariesHandler
     {
         public static bool enableInteractionHoverFeedback = true;
 
@@ -109,6 +115,8 @@ namespace DCL.Components
 
             entity.OnShapeUpdated -= SetEventColliders;
             entity.OnShapeUpdated += SetEventColliders;
+            
+            DataStore.i.sceneBoundariesChecker.Add(entity,this);
         }
 
         public WebInterface.ACTION_BUTTON GetActionButton()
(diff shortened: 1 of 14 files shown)


unity-renderer/Assets/DCLPlugins/UUIDEventComponentsPlugin/UUIDComponent/OnPointerEvent.cs
```diff
@@ -109,6 +115,8 @@ public override void Initialize(IParcelScene scene, IDCLEntity entity)
 
             entity.OnShapeUpdated -= SetEventColliders;
             entity.OnShapeUpdated += SetEventColliders;
+            
+            DataStore.i.sceneBoundariesChecker.Add(entity,this);
```
sandrade-dcl, line 119: Maybe it would be nice to inject the `DataStore` as a dependency in this `Initialize()` method.

sandrade-dcl (approved):
LGTM! Good work!

unity-renderer/Assets/DCLPlugins/UUIDEventComponentsPlugin/UUIDComponent/OnPointerEvent.cs
pravusjif: You are right that sounds good, but that would require a big refactor as we should change the `Initialize()` method from [`IEntityComponent`](https://github.com/decentraland/unity-renderer/blob/dev/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Interfaces/IEntityComponent.cs), as the `Initialize()` of the `OnPointerEvent` is called in [`ECSComponentsManagerLegacy.EntityComponentCreateOrUpdate()`](https://github.com/decentraland/unity-renderer/blob/dev/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ECSComponentManagerLegacy/ECSComponentsManagerLegacy.cs#L421) along with the `Initialize()` of every other component implementing the `IEntityComponent` interface.

kwssbocian (approved):
Issue appears to be fixed.
✔️Upon approaching a poker table every floating card is interactable and clicking it causes a prompt to pop up.

PR #2960, opened 2022-09-01
fix: username resetting after modified in modal window

dalkia:
## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/UIComponents/Scripts/Components/ProfileCard/ProfileCardComponentView.cs
+++ b/unity-renderer/Assets/UIComponents/Scripts/Components/ProfileCard/ProfileCardComponentView.cs
@@ -172,8 +172,6 @@ public class ProfileCardComponentView : BaseComponentView, IProfileCardComponent
 
         if (profileAddress != null)
             profileAddress.gameObject.SetActive(!isClaimedName);
-
-        SetProfileName(model.profileName);
     }
 
     public void SetLoadingIndicatorVisible(bool isVisible) { profileImage.SetLoadingIndicatorVisible(isVisible); }


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/AvatarEditorHUDController.cs
```diff
@@ -359,6 +359,9 @@ public void LoadUserProfile(UserProfile userProfile, bool forceLoading)
         if (userProfile.avatar == null || string.IsNullOrEmpty(userProfile.avatar.bodyShape))
             return;
 
+        if (DataStore.i.exploreV2.nameOrDescriptionJustSaved.Get())
```
lorux0, line 362: im not really convinced on the solution since we are patching a kernel issue.. is it necessary after the kernel changes?

dalkia: That solves an issue that kernel is not handling. If you do the following steps, you will get some incosistencies>

1) Open avatar editor hud and change wearabales
2) Open the name tab while avatar editor hud is open with unsaved changes
3) Edit and save name/description
4) Since the save name triggers a profile update (which I think makes sense, to confirm that saving has been done correctly) the unsaved changes of the wearables get overridden.

This is the reason why I added that check. Another option would be to ask kernel just to send back the data that has been saved, but that may get messy on both sides.

kwssbocian (approved):
Both issues appear to be fixed.
✔️ Opening and changing contents of the Profile HUD causes the changes to be correctly retained in game as well as in https://account.decentraland.org/. Changes persist after closing and reopening the backpack, profile menu and after changing the worn wearables.
✔️ Changes made to the currently worn wearables are retained after accessing the Profile HUD including after selecting "Done" then reopening the backpack and accessing the Profile HUD or changing the username and/or about description.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/AvatarEditorHUD/Scripts/AvatarEditorHUDController.cs
lorux0: so probably instead of having a variable referring to saving the name or description of the profile, we could have an internal variable in `AvatarEditorHUDController` to check if the user has unsaved changes on the wearables. Then avoid replacing them if the profile gets updated in the middle? What do you think?

dalkia: Agreed; actually there was already a bool that had the dirty state of the avatar. So I just reused it.

lorux0 (approved):
great work!

PR #2965, opened 2022-09-02
nft emotes different approach to new structure

Antivortex:
## WHAT AND WHY

There were issues in changing nft emotes structure so after discussion with Mendez we decided to use different approach: make only one version for the Renderer emote structure but convert structures for old users in runtime.

## HOW TO TEST

To test this change it is very important to test in both old users and new users. For old users we need to check if their old NFT emotes will be present. Old user is considered anyone who created his profile and added some NFT emotes and/or other wearables before this change was in the Renderer.

## Our Code Review Standards


Kinerius (approved):
LGTM

davidejensen (approved):

PR #2976, opened 2022-09-06
feat:  ECS7 update pointer events component with a new approach

Suduck:
## Context

In order to have a better understanding of why are we changing the way that pointer event result works you should go to this issue: decentraland/sdk#352

We have defined a new way to retrieve the result of the pointer event and we need to implement it. 

This is the first step going forward decentraland/sdk#358.

The reason behind implementing this approach instead of the complete `Append` decentraland/sdk#358. This is because of the time that we have to achieve feature parity with ECS6. This is an intermediary approach that won't break the API when we implement the `Append` to the CRDT message

## What does this PR change?

This PR is the second part of 

- Removes the old PointerEvents components and ( `OnPointerUp ` and `OnPointerDown`) and implements a generic `PointerEvent` component.

This component will contain an array of all the pointer-events specified by the user.

## Know issues

PointerEventUp is not working as expected since the hover text is not appearing correctly. This is happening already so it is not related to this PR

## How to test the changes?

1.- Download the scene https://github.com/decentraland/ecs7-template and use the following branch: `feat/pointer-event-test`
2.- Follow the instructions in the md of the repo and install this SDK version `npm install "https://sdk-team-cdn.decentraland.org/@dcl/js-sdk-toolchain/branch/feat/update-pointer-event-result/dcl-sdk-7.0.0-2978388596.commit-be4470d.tgz"`
3.- Start the scene with dcl start
4.- Use this URL: http://127.0.0.1:8000/?position=0%2C0&ENABLE_ECS7&renderer-branch=feat/update-pointer-event-result-component&kernel-branch=main&SCENE_DEBUG_PANEL=&realm=v1%7E127.0.0.1%3A8000
5.- Change the code with differents pointers events and differents buttons to see if everything works as expected
6.- Test those two function  `isPointerEventActive` and  `wasEntityClicked `

## Our Code Review Standards


pbosio:
replaced by: https://github.com/decentraland/unity-renderer/pull/3060

PR #2979, opened 2022-09-06
feat: enable or disable channel notifications

lorux0:
## What does this PR change?

Changed the way channel messages are loaded. The old implementation had issues when trying to reload channels that you already seen.

Added the possibility to enable or disable channel notifications.

## How to test the changes?

1. Join any channel (from the channel browser o through command).
2. Get into the channel, click the mute channel button in header of the window.
3. Start pressing `G` to receive fake messages. They should not be displayed in the notifications panel.
4. You will see that the taskbar grows the notification count, that is because we have a mocked implementation that is not taking in consideration the mute status of the channel.

## Our Code Review Standards


sandrade-dcl (approved):
LGTM!

sandrade-dcl (approved):

PR #2980, opened 2022-09-06
fix: ECS7 hover feedback

pbosio:


pbosio:
replaced by: https://github.com/decentraland/unity-renderer/pull/3060

PR #3004, opened 2022-09-09
Old emotes loading time hot fix

Antivortex:
## What does this PR change?

There are two different methods responsible for old and new emotes loading. After we removed version field I made the calls to these methods sequential which is a wrong approach and the right one without version field would be to use the one that comes first. With a sequential approach old emotes will not start loading until the call to new emote by the same id times out.

With this change both of them will be called at the same time and the first one finished will be used.


davidejensen (approved):
this change was made for before the catalyst update, now it is not required anymore but it works as expected

PR #3021, opened 2022-09-12
fix: avatar shape glitch caused by zero scale

### LINKED ISSUE #2997: AvatarShape glitch
Video of the issue: 

https://user-images.githubusercontent.com/7646450/189163597-cd61f411-f93b-4b51-a609-09de20fb5086.mp4

Possible issues:

- AvatarShape is getting combined with its parent meshes for some reason so everything goes crazy.
- Sometimes it does not glitch, maybe there's a race condition causing an issue
- Some errors about some objects having infinity scale

Kinerius:
## What does this PR change?

Closes #2997 

## How to test the changes?

- Download and extract this scene: [mirrors.zip](https://github.com/decentraland/unity-renderer/files/9550998/mirrors.zip)
- Enter the folder on a terminal
- Initialize the scene using `dcl start` ([Installation Guide](https://docs.decentraland.org/development-guide/SDK-101/) of the DCL Tools)
- When the browser opens add this `&renderer-branch=fix/avatar-shape-glitch` to the end of the url
- Url should look like this http://127.0.0.1:8000/?position=0%2C0&SCENE_DEBUG_PANEL&renderer-branch=fix/avatar-shape-glitch
- Check that there is no glitch

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
```diff
@@ -195,6 +195,17 @@ public override IEnumerator ApplyChanges(BaseModel newModel)
...
                 }
+
+                // NOTE (Kinerius): Since users can spawn avatar shapes with zero scale,
+                // we avoid the glitches generated by it by moving this object elsewhere with a normal scale (#2997)
+                var thisTransform = transform;
+                Transform prevParent = thisTransform.parent;
+                Vector3 prevScale = thisTransform.localScale;
+                Vector3 prevPos = thisTransform.localPosition;
+                thisTransform.SetParent(null);
+                thisTransform.localPosition = Vector3.one * 9999;
```
lorux0, line 206: cant we disable the gameObject instead of doing this?

lorux0:
i've noticed this glitch in the scene:
Screen Shot 2022-09-14 at 15 55 05

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
Kinerius: Disabling would do nothing because it would still be a child of an object hierarchy of zero scale

Kinerius:
> i've noticed this glitch in the scene: Screen Shot 2022-09-14 at 15 55 05

This is normal, there's an avatar with zero scale being spawned there, we have to give tools to content creators to avoid using the spawn FX

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
lorux0: so what if you unparent and disable the object?

dalkia (approved):

Issue #2982, opened 2022-09-07
Message on top of the 'nearby' channel overlaps with the Profile menu


kwspgoliasz:
**Description:**
Opening other user's Profile menu from a message located below the chat notification on top of the 'nearby' channel, results in part of the menu getting hidden behind the message. Reproducing the issue requires at least two users to log into the platform.

**Steps to reproduce:**

1. Load into the game as a guest or via a wallet.
2. Have other users write on the nearby channel until the messages reach the upper part of the chat.
3. Right-click on the received message to show the Profile menu.
4. Observe the issue.

**Results:**
The Profile menu is partially concealed by the notification on top of the 'nearby' channel.

**Expected results:**
The Profile menu is displayed over the notification on top of the chat.

**Notes:**

- The issue occurs on both the guest and wallet accounts.

- The issue occurs on both the production and development configurations.

- Please see the attached video for reference: https://images.zenhubusercontent.com/337227404/8a79b6f3-4b98-435e-8afb-d7c2df2a8369/_2982.mp4.

- Platform: Web Client, Windows Desktop Client.

popuz:
This issue can be reproduced in-editor much simpler: 
1. in-editor Right-click on the ChatController Component **(1)** attached to the Main GameObject
2. choose Fake Private Message several times **(2)**
3. step 3 from original "Steps to reproduce" in this report (see above)
![image.png](https://images.zenhubusercontent.com/631628793fb7f60ae36ae984/9449c87f-df2c-4e01-b12d-6118a5c1c138)

popuz:
Issue caused by the **DescriptionContainer** (1) being lower in Hierarchy than **UserContextMenuPanel** (2). It is also overlays **ConfirmationDialog**.
There are 2 solutions:
1. We can just move **DesctiptionContainer** inside **Chat** (3) hierarchy above the **UserContextMenuPanel** and the **ConfirmationDIalog**. 
2. We can add Canvases to the **DesctiptionContainer**,  **UserContextMenuPanel**, **ConfirmationDIalog** and play with the overriding Sorting Layers. 
![image.png](https://images.zenhubusercontent.com/631628793fb7f60ae36ae984/1568fdbd-fc49-4086-9065-189fcf1febae)

popuz:
resolved in the https://github.com/decentraland/unity-renderer/tree/feat/channels

PR #3029, opened 2022-09-13
fix: move DescriptionContainer inside Chat prefab hierarchy above other panels

popuz:
## What does this PR change?

- Order in the hierarchy for Chat elements is changed. Now, DescriptionContainer is higher and does not overlaps UserContextPanel (for more details see [issue](https://github.com/decentraland/unity-renderer/issues/2982) comments). 
- DescriptionContainer moved from GeneralChatChannelHUD.prefab inside ChatHUD.prefab

## How to test the changes?
Please refer to the [issue](https://github.com/decentraland/unity-renderer/issues/2982) comments) comments where a step-by-step guide to reproduce the issue is presented with screenshots.


dalkia (changes requested):
Hi @popuz ! 

By moving the `ChatDescriptionContainer` into `ChatHUD` the disclaimer message has been added into `PrivateChatHUD` prefab. Therefore, it is showing the disclaimer message from #nearby channel in every private conversation
image

On the other hand, I would ask @lorux0 o @sandrade-dcl if it makes sense to add this fixright now, since this two prefabs are heavily modified in the `feat/channels` branch. Maybe its better to do this changes directly in that branch and wait for their merge.

popuz:
hey @dalkia, thanks for the update. Another solution could be to keep the previous hierarchy but play with the Sorting of Canvases. The downside (or maybe not) of this solution would be that we would need to present `Canvas `component for `DescriptionContainer` and `UserContextPanel` (maybe also `ConfirmationDialog`)

sandrade-dcl:
> On the other hand, I would ask @lorux0 o @sandrade-dcl if it makes sense to add this fixright now, since this two prefabs are heavily modified in the `feat/channels` branch. Maybe its better to do this changes directly in that branch and wait for their merge.

Yes please... to avoid confilcts, any change on prefabs related to chat, lets do it using our current `feat/channels` branch. @popuz I would suggest you to make this PR be based on the channels branch (unless this fix is something urgent for prod).

PR #3038, opened 2022-09-14
fix: emotes play analytics fixed for new emotes

### LINKED ISSUE #2963: Add urn to equip_emote analytics event
Emote structure was changed and therefore one of the analytics events should be changed to to include urn of the emote. We should also check other analytics events related to emotes because those can break after recent changes.

Antivortex:
## What does this PR change?

This PR fixes analytics of playing an emote for new emotes. If the emote is not found in the collection of wearables in that case it will be found in the new emotes catalog.

Closes #2963 

## How to test the changes?

Check if analytics event actually coming to our analytics when you play a new custom NFT emote.

## Our Code Review Standards


kwspgoliasz (approved):
The feature appears to be implemented correctly, as the analytics are correctly displayed in the console when the emotes are used.

AjimenezDCL (approved):
Nice catch, thx!

davidejensen (approved):

PR #3042, opened 2022-09-15
Added empty list message for collectibles

PalauiDCL:
In the collectibles tab, if no collectibles are present, this note will be shown:

image


AjimenezDCL (approved):

kwssbocian (approved):
✔️ Empty list message for the Collectibles tab shows "You don't own any collectible yet." when the wallet login character does not posses any collectibles.

PR #3066, opened 2022-09-19
fix: ECS7 NFTShape

pbosio:
## What does this PR change?
**changes affecting production:**
[NFTShapeHQImageHandler.cs](https://github.com/decentraland/unity-renderer/pull/3066/files#diff-d41921f7df64b5f5d604ca0b5ed29344da9ffedbc7505de25599c12d253d76e1) was modified to contain only needed references (transform and collider) instead of a `INFTShapeLoaderController` reference

**changes affecting ecs7:**
updated `NFTShape` removing everything related to colliders and visibility fields
fixed exception while updating `NFTShape` model in runtime
added renderer component for `NFTShape`
disabled `NFTShape` prefabs colliders
added/updated tests accordingly
...

## How to test the changes?

Go to: https://play.decentraland.zone/?renderer-branch=fix%2Fecs7-nftshape-colliders&position=51,89 or any other location that has NFTShapes
Check that NFT image is in low resolution while not looking at it but it switch to a higher resolution while looking at it

low resolution image while not facing image directly:
Screen Shot 2022-09-21 at 11 10 56

high resolution when pointing the camera towards the image:
Screen Shot 2022-09-21 at 11 11 10

## Our Code Review Standards


unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/NFTShape/Handler/ECSNFTShapeComponentHandler.cs
```diff
@@ -1,126 +1,152 @@
...
+                return;
+
+            renderersComponent.RemoveRenderer(scene, entity, nftFrame.frameRenderer);
+            nftFrame.Dispose();
+            Object.Destroy(nftFrame.gameObject);
         }
-        
-        internal void ApplyModel(PBNFTShape model)
+
+        internal async void LoadNFT(string scr)
```
AjimenezDCL, line 113: We need cancellation here.

I would have a CancellationToken external in the param and another one internal (for dispossing purposes). There are a bunch of examples in  the project using linked cancellation token in case you need a working example.

AjimenezDCL (changes requested):
Good job, just left some reuqests on terms of cancellation

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/NFTShape/Handler/ECSNFTShapeComponentHandler.cs
AjimenezDCL: Nevermind, I saw that the `LoadNFT` method already has cancellation inside and also de `Disposal` should cover all the cases. Disregard my comment pls!

AjimenezDCL (approved):

PR #3067, opened 2022-09-19
feat: channels limit error

lorux0:
## What does this PR change?

Implementation of specific ways to show an error when the user tries to join a channel but already has reached the limit of available channels to join.

## How to test the changes?

1. Join or create 5 channels from `/join` command or create channel window
2. Go to create channel window
3. Try to create a channel
4. An error displaying that limit reached should be displayed
5. Type in chat `/join anychannel` command
6. A modal displaying that limit reached should be displayed

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WorldChatWindowHUD/ChannelLimitReachedWindowComponentView.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WorldChatWindowHUD/ChannelLimitReachedWindowComponentView.cs
@@ -1,0 +1,35 @@
+using System;
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace DCL.Chat.HUD
+{
+    public class ChannelLimitReachedWindowComponentView : BaseComponentView, IChannelLimitReachedWindowView
+    {
+        [SerializeField] internal Button[] acceptButton;
+        
+        public event Action OnClose;
+        public Transform Transform => transform;
+        public void Show() => gameObject.SetActive(true);
+
+        public void Hide() => gameObject.SetActive(false);
+
+        public static ChannelLimitReachedWindowComponentView Create()
+        {
+            return Instantiate(
+                Resources.Load<ChannelLimitReachedWindowComponentView>("SocialBarV1/ChannelLimitReachedModal"));
+        }
+
+        public override void Awake()
+        {
+            base.Awake();
+
+            foreach (var button in acceptButton)
+                button.onClick.AddListener(() => OnClose?.Invoke());    
+        }
+
+        public override void RefreshControl()
+        {
+        }
+    }
+}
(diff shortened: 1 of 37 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WorldChatWindowHUD/ChannelLimitReachedWindowComponentView.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace DCL.Chat.HUD
+{
+    public class ChannelLimitReachedWindowComponentView : BaseComponentView, IChannelLimitReachedWindowView
+    {
+        [SerializeField] internal Button[] acceptButton;
```
sandrade-dcl, line 9: Lets use a button from our UI Components library (`ButtonComponentView`). If you don't need some of the features this button have, you can simply deactivate the corresponding components in the used prefab or even we could create another button prefab more simple, but it's important to use the prefabs from our UI library to keep the consistency between our huds and to make them ready to be changed from the source prefab (for example for future common style's changes).

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WorldChatWindowHUD/ChannelLimitReachedWindowController.cs
```diff
@@ -0,0 +1,35 @@
+namespace DCL.Chat.HUD
+{
+    public class ChannelLimitReachedWindowController : IHUD
```
sandrade-dcl, line 3: Following the discussing about the use of the `HUDController`/`HUDFactory` classes to instantiate huds, we should stop using them and take advantage of our new plugin system. We have currently some huds working through a system and other ones through other system. In this case I think we could have integrated this hud through the plugin system in the same way we do for the JoinChannel HUD (`JoinChannelComponentController`) and not to make it dependant of the other features like the taskbar.

Anyway it is not something that we have to fix in this PR, but for start thinking about this and reach/implement a common way together for the near future 👍

sandrade-dcl (approved):
I left some suggestion that we can fix/implement later.

I have reviewed the code and tested the build and the feature is working fine! Good work!

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WorldChatWindowHUD/ChannelLimitReachedWindowComponentView.cs
lorux0: done!

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WorldChatWindowHUD/ChannelLimitReachedWindowController.cs
lorux0: agree, done!

sandrade-dcl (approved):

Issue #3056, opened 2022-09-19
Prepare code style guide draft


aixaCode:
As the team is growing we are missing alignment on how to style the code.
The goal is to provide a clear set of rules or guidelines that each developer on a team should contribute to when writing code.
Agreeing on a consistent style across the team can help you develop a clean, readable, and scalable codebase. A code style guide helps you remove the guesswork for every member of your team by establishing consensus around important coding conventions.

### Acceptance criteria
* [ ] Code style readme is updated with descriptive mini examples
* [ ] Style example class included in the project
* [ ] Collected feedback from the team

PR #3073, opened 2022-09-20
Docs: code conventions

popuz:
## What does this PR change?
close issue: https://github.com/decentraland/unity-renderer/issues/3056

With this PR we start the initiative on improving our Code Styles and Guidelines document. This PR will be the place to discuss and develop together a common code style. It should fix problems with ordering class members, choosing names and etc., providing faster navigation in the classes by knowing the agreed code structure.

As a final result, I propose that we will have one .md file with code snippets for focused areas (with comments) and then 1 example class. 

Since it is an OpenSource project, we want our code style to be user-friendly for contributors. That is why we try
- to stick to the most commonly used code style for C#/Unity
- to not over-complicate it
- our code style doc to be compact, but informative ([Google C# code guidelines doc](https://google.github.io/styleguide/csharp-style.html) is taken as template)

## References
[Goggle C# code guidelince](https://google.github.io/styleguide/csharp-style.html)
[Microsoft .NET guidelince](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/)
[Microsoft C# Coding Conventions](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions)
[dotnet/runtime](https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/coding-style.md)
[Unity Code Style Guide](https://blog.unity.com/technology/clean-up-your-code-how-to-create-your-own-c-code-style)
[Unity Code Style Guide - example](https://github.com/thomasjacobsen-unity/Unity-Code-Style-Guide/blob/master/StyleExample.cs)


popuz:
For me first open question here is
- Should we fully split Fields and Properties or consider it as one group to allow placing property with its backend field together? If yes, then consider how it will look with `[sfields]`.
```
private bool isInitialized;                         
public bool IsInitialized { get => isInitialized; set => isInitialized = value; }
```
-

unity-renderer/Assets/Scripts/Example.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            // subscribe?
+        }
+
+        private void OnDisable()
+        {
+            // unsubscribe?
+        }
+
+        private void OnCollisionEnter(Collision collision) { }
```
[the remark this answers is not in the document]
popuz: Fully agree with step by step transition. That is why I tried to not change the naming convention at all but just document it. Because re-naming could be quite a huge shift (as you mention about PRs). I was thinking that fixing the ordering could be the first step (which will improve navigation a lot).

However, answering your question, I would rather prefer to have `_` for all non-public Fields and leave the local Variables and methods Parameters with the `camelCase`. It is more of a standard in the industry. I think @Antivortex also likes `_`-for-fields solution. I see `_` as an additional tool, but in a small scope of the method's body you do not need such tools - everything is on the screen.

popuz:
> I love this convention and is also aligned with the standard in the industry.
> 
> I'd love to have this implemented, but we have to transition step by step towards it, taking in mind that the PRs are gonna be so polluted due to the syntax changes.

Fully agree with step by step transition. That is why I tried to not change the naming convention at all but just document it. Because re-naming could be quite a huge shift (as you mention about PRs). I was thinking that fixing the ordering could be the first step (which will improve navigation a lot).

unity-renderer/Assets/Scripts/Example.cs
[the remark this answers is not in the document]
AjimenezDCL: Sure! I dont really mind as long as we differentiate them.

docs/style-guidelines.md
```diff
@@ -17,4 +17,96 @@ To install it go to the root of your explorer repository and run the following c
...
+                       
+public bool Property1 { get; set; }                                        // Properties groups  
+public bool Property2 { private get; set; }
+public bool Property3 { get; }
+public bool Property4 { get; private set; }
+
+protected bool Property5 { get; set; }
+private bool Property6 { get; set; }
+
+public event Action Started;                                               // Events group
```
lorux0, line 110: i would add a prefix on events: `OnStarted`, `OnCompleted`

unity-renderer/Assets/Scripts/Example.cs
```diff
@@ -0,0 +1,62 @@
+﻿using System;
+using UnityEngine;
+
+namespace DefaultNamespace
```
lorux0, line 4: we should also have a standard for namespaces. I think they should follow the folder structure in the best case scenario (as Rider suggests), but in DCL we do have a mess in that regard which implies a very big refactor on the folder structure. For example:
- Folder: `DCL/Social/HUD`
- Namespace: `DCL.Social.HUD`

To mitigate this issue, we are setting a root namespace for the feature, ignoring the folder structure. For example:
- Folder: `/Scripts/MainScripts/DCL/Controllers/ChatController/`
- Namespace: `DCL.Social.Chat`

If you do have a nested folder then:
- Folder: `/Scripts/MainScripts/DCL/Controllers/ChatController/InnerGroup`
- Namespace: `DCL.Social.Chat.InnerGroup`

unity-renderer/Assets/Scripts/Example.cs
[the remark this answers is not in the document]
lorux0: i would also avoid `_`.. i prefer to referrence fields as `this.field` but rather rename the parameter so it does not collide

docs/style-guidelines.md
popuz: I am more of a fan of a classical Microsoft .NET notation for event naming, where the event itself means finished actions (`Started`, `Changed`)  and the subscriber has an `On` prefix. This is suggested [in Microsoft guidelines](https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members) 
image
and also in the Unity code-style guidelines ([blog](https://blog.unity.com/technology/clean-up-your-code-how-to-create-your-own-c-code-style), book(file:///C:/Users/Vitaly/Downloads/create-a-c-style-guide-write-cleaner-code-that-scales%20(1).pdf)), namely saying

> Choose a name that communicates the state change accurately. Use the present or past participle to indicate events “before” or “after.” For example, specify “OpeningDoor” for an event before opening a door or “DoorOpened” for an event afterward.

> Prefix the event raising method (in the subject) with “On”: The subject that invokes the event typically does so from a method prefixed with “On,” e.g. “OnOpeningDoor” or “OnDoorOpened.

popuz: @Antivortex, mentioned above .NET style (without On prefix), removes the need of having postfixes for subscribing method (like OnSomethingStarted**Handler**), because prefix now plays this role and identifies subscribed methods.

unity-renderer/Assets/Scripts/Example.cs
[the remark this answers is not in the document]
popuz: when defining code style for an open-source project, it could be beneficial to think about what is most commonly used notation, so it will be smoother onboarding for the external contributors. That is why I would vote for [Google C#](https://google.github.io/styleguide/csharp-style.html) and [.NET code style](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions), where _ is used for `private`/`protected` fields. 

However, if create our own style and agree on `_` for parameters, then @Antivortex question is valid - should we use this notation also for local variables (variables defined inside the method)?

unity-renderer/Assets/Scripts/Example.cs
popuz: @Antivortex , it is very easy In Rider  - right-click on the folder in the Explorer window (or Alt+Enter) and then disabling `Namespace provider` checkbox
image

docs/style-guidelines.md
lorux0: i think the prefixes helps to avoid overlapping names with methods or properties. Its also recommended by default on rider

popuz: It could be that the .NET notation is also related to a readability question, so "code reads like well-written prose" (Grady Booch).` Door.Opened` is more grammatically correct than `Door.OnOpened`, while `OnDoorOpened` also correct and could be used for a subscriber

popuz: and Event will not overlap with the method name, because Method name should be a verb in the present tense, while event is in the Past tense.

unity-renderer/Assets/Scripts/Example.cs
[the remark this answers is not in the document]
AjimenezDCL: Unless you wanna go the Hungarian Notation way (which I truly dislike) there's stuff we have to give up.

You have public <> private <> params <> local variables and not so many permutation of styles. Nowadays IDEs do the heavy lifting and with a good coloring schema that's usually just enough to know what you are dealing with.
I think it's just a matter of prioritizing what we want to stand out.

To be honest @popuz has got a really good point a bout using the most commonly used style. .NET code style seems like the correct way IMO.

docs/style-guidelines.md
AjimenezDCL: To be honest I've always seen them as `OnSomething`, I'd go with Nico's proposal.

unity-renderer/Assets/Scripts/Example.cs
AjimenezDCL: I think this doesn't work really well with Unity's asmdef because they are forced into different folders. It will be even more difficult once we merge desktop/webgl repo and become truly multiplatform. How would you deal with interfaces?

##  Case A

Let's say you have:
- `interface IMyService` 
- `class MyService_WebGL : IMyService`
- `class MyService_Desktop : IMyService`

I guess we can solve it with paths like:
- `Assets/Scripts/Common/MyService/IMyService.cs`
- `Assets/Scripts/WebGL/MyService/MyService_WebGL.cs`
- `Assets/Scripts/Desktop/MyService/MyService_Desktop.cs`

and we would strip `Assets/Scripts/****/`


## Case B
You have:
- `class ServiceA : IServiceA`
- `class SystemB`  (which uses `IServiceA`)

You want to split `ServiceA` into a _definitions_ and _implementation_ folders with different assemblies to avoid cycling references. Paths would be like
```
- ServiceA/
---- ServiceA.asmdef
---- ServiceA.cs
---- Definitions/
-------- ServiceADefinitions.asmdef
-------- IServiceA.cs
```

I would find really ugly to have a namespace `ServiceA.Definitions` to access `IServiceA`. How could we deal with this.

docs/style-guidelines.md
```diff
@@ -17,4 +17,96 @@ To install it go to the root of your explorer repository and run the following c
...
+{
+    public enum Side { Left, Right }                    // Enumeration and Enumerators -> PascalCase 
+    public delegate void Change<T> (T current);         // Delegate declaration -> PascalCase
+    public interface IInitializable { }                 // Interface -> PascalCase, starts with 'I'
+    
+    public class MyClass : IInitializable               // Class/Struct -> PascalCase.
+    {
+        public const string ASSET_PATH = "AvatarUI";    // Constant -> CAPITALS_SNAKE_CASE 
+        public int PublicField;                         // public Field -> PascalCase
+        private bool isVisible;                         // Field -> camelCase
```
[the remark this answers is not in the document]
AjimenezDCL: Personally I rely more on schema coloring than name style to identify scopes when reviewing (I do most of my PRs on Rider) or maintaining code I'm not familiar with.

To be honest I dont mind this idea, as long as the team needs it and we dont convolute the style too much, sure thing.

docs/style-guidelines.md
```diff
@@ -17,4 +17,96 @@ To install it go to the root of your explorer repository and run the following c
...
+}
+```
+Suggestions:
+* `Interface` - try to name it with adjective phrases (`IDamageable`). If it is difficult, then use descriptive noun (`IBaseVariable`) or noun phrase (`IAssetProvider`).
+* `Class`/`Struct` -  name with nouns (`Health`) or noun phrases (`InputService`).
+
+### Ordering conventions
+* `using` go first and placed alphabetically
+* Class members grouped and appears in the following order
+  * Enums, delegates declarations
```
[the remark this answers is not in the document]
AjimenezDCL: I agree with Nico.

99% of the cases an Action or Func is just enough, but let's avoid having `Action<string, string, string, string>` and missing context on every param.

docs/style-guidelines.md
popuz: Then let's go with `OnSomething` 😉

unity-renderer/Assets/Scripts/Example.cs
popuz: It is a really good point and complex topic about namespaces vs asmdefs vs folder structure. I suggest discussing it separately (as the third iteration of refinement) - first namings, second order, and third namespaces. So let's take these arguments to later discussion.

docs/style-guidelines.md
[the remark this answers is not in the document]
popuz: .NET Microsoft standard suggests using `Action<>` and `Func<>`. The example above could be solved by presenting the `Struct' to replace 4 arguments with one (as suggested in the CleanCode approach). However, I also found custom delegates to be useful in some automation processes and tests (where, for example, you can create a certain reflection Util around your delegate types and improve reports and readability of tests)

docs/style-guidelines.md
```diff
@@ -17,4 +17,96 @@ To install it go to the root of your explorer repository and run the following c
...
+```csharp
+namespace MyProject                                     // Namespace -> PascalCase
+{
+    public enum Side { Left, Right }                    // Enumeration and Enumerators -> PascalCase 
+    public delegate void Change<T> (T current);         // Delegate declaration -> PascalCase
+    public interface IInitializable { }                 // Interface -> PascalCase, starts with 'I'
+    
+    public class MyClass : IInitializable               // Class/Struct -> PascalCase.
+    {
+        public const string ASSET_PATH = "AvatarUI";    // Constant -> CAPITALS_SNAKE_CASE
```
popuz, line 39: guys, what do you think also about the `constants`? I would rather go with the .NET standard and just use `PascalCase` for it instead of `CAPITALS_SNAKE_CASE`, or even go with Google C# standards where 
> _"Naming convention is unaffected by modifiers such as const, static, read-only, etc."_
My point here would be similar to @AjimenezDCL, that IDE is so powerful nowadays and makes constants always Bold, so it is easily distinguishable.

unity-renderer/Assets/Scripts/Example.cs
popuz: One solution for the problem **Case B** mentioned by @AjimenezDCL, would be to make namespaces wider and not present them for small folders (so to not go very detailed). In Case B, I would propose to not present `ServiceA.Definitions` namespace and only have the namespace `ServiceA` for both definitions and implementation. 
In the end, the main goal of namespaces is to avoid names collisions

docs/style-guidelines.md
```diff
@@ -1,20 +1,147 @@
...
+```csharp
+// Note: indentaion for helper methods is used only for clarity. It is not a part of our formating style 
+public void Test1()                   // called by other class
+{
+    Test1Helper1();
+    Test1Helper2();
+}
+
+  private void Test1Helper1() => A();  // called by Test1() 
+    internal void A() { }              // called by Test1Helper1()
```
aixaCode, line 131: What about using local functions for those types of invocations? 
```suggestion
  private void Test1Helper1()
  { 
     LocalA();
     void LocalA() { }
  };  // called by Test1() 
```

unity-renderer/Assets/Scripts/CodeStyleExample.cs
```diff
@@ -0,0 +1,130 @@
+﻿// NOTE: following separation lines are used only for clarity and do not take part of our commenting code style!
+//----------------------------------------------------------------------------------------------------  Usings group
+
+using System;
+using System.Collections.Generic;
+using JetBrains.Annotations;
+using UnityEngine;
+using UnityEngine.Events;
+
+namespace UIComponents.Scripts.Components.SectionSelector
```
lorux0, line 10: i think we should skip including `Scripts` as part of the namespace since it does not describe anything useful

unity-renderer/Assets/Scripts/CodeStyleExample.cs
```diff
@@ -0,0 +1,130 @@
...
+            IsInitialized = true;
+
+        public void ApplyDamage(float damage) // Public - called from outside of the class (by other class)
+        {
+            float resultDamage = ComputeDamage(damage);
+            TakeHit(resultDamage);
+        }
+
+        private float ComputeDamage(float damage) => // Called by ApplyDamage
+            damage - armor;
```
lorux0, line 75: inline methods, on cases that the implementation is very short, we could do it in the same line:
```
private float ComputeDamage(float damage) => damage -= armor;
```

docs/style-guidelines.md
lorux0: agreed

docs/style-guidelines.md
```diff
@@ -1,20 +1,147 @@
...
+private void Awake() { }               // called by Unity
+```
+  * Not-helper methods should follow the order (where its helper methods follows previous rule and allowed to be placed in between of this order):
+    * constuctor
+    * destructor 
+    * `public`
+    * `public` with `[PublicApi]` and `[UsedImplicitly]` attributes
+    * Unity-callbacks 
+      * `Awake`, `Start`, `OnEnable`, `OnDisable`, `OnDestroy`,
+      * other callbacks (with respect to `Enter`-`Stay`-`Exit` order)
```
pbosio, line 144: wdyt about moving `Unity-callbacks` up to the same place as constructor/destructor?
1. because, similar to constructor/destructor, they are events related to the script lifecycle
2. it's easier to spot that we are looking at a `MonoBehaviour` script, since most of the time `: MonoBehaviour` can be easier to miss

docs/style-guidelines.md
```diff
@@ -1,20 +1,147 @@
...
+  private void Test1Helper1() => A();  // called by Test1() 
+    internal void A() { }              // called by Test1Helper1()
+  private void Test1Helper2() => B();  // called by Test1()
+    public void B() { }                // called by Test1Helper2()
+
+private void Awake() { }               // called by Unity
+```
+  * Not-helper methods should follow the order (where its helper methods follows previous rule and allowed to be placed in between of this order):
+    * constuctor
+    * destructor
```
pbosio, line 139: should we also consider `Dispose` method to be placed at this level?

lorux0: agree

unity-renderer/Assets/Scripts/CodeStyleExample.cs
popuz: Yeah, it was just more as a scratch. We will discuss it again in Slack more in detail and vote. But I personally like to have experession-body for properties to be on one line, but for methods always put in the new line. In such way it is easier for me to visually separate from properties and methods

docs/style-guidelines.md
popuz: We can try to propose it. I was using this approach once, but it wasn't that good. Also, we there are some edge-cases for heap-allocation to keep in mind with local functions  https://devblogs.microsoft.com/premier-developer/dissecting-the-local-functions-in-c-7/

popuz: @lorux0, yep, I think I mentioned in the comment that this indentation is not a part of our code style and only used for clarity in this example

docs/style-guidelines.md
popuz: It was proposed to the team, but the majority chose `public` to go first :(

docs/style-guidelines.md
popuz: So `Initialize` then as well?

unity-renderer/Assets/Scripts/CodeStyleExample.cs
Kinerius: Indeed, namespaces get an automatic name by Rider but they are actually their folder structures. 
We should use a Domain approach to names, in this case, a `DCL.UIComponents` is more than enough.

docs/style-guidelines.md
```diff
@@ -1,20 +1,313 @@
...
+/// <summary>
+/// This is the InitialScene entry point.
+/// Most of the application subsystems should be initialized from this class Awake() event.
+/// </summary>
+public class Main: MonoBehaviour
+{
+    /// ...
+}
+```
+* **only** use comments `//` inside of the class when it is necessary and there is a need to provide additional information which cannot be covered by good naming:
```
menduz, line 278: What is the rationale behind this?

AjimenezDCL: Just following best practices and avoiding ofuscating smelly code with smelly comments.
https://www.codeguru.com/csharp/commenting-best-practices-in-c/
https://www.dotnetstuffs.com/code-commenting-best-practices-csharp/
https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/
https://www.codingem.com/how-to-comment-code/

Specifically cases like this:
image

image

aixaCode (approved):

docs/style-guidelines.md
```diff
@@ -1,20 +1,313 @@
...
+```
+
+### Attribute usages
+* **Class**
+  * Use `[RequireComponent(typeof(MyComponent))]` attribute when you cache components via `GetComponents` on `Awake` and `Start` callbacks
+  * Use `[DisallowMultipleComponent]` attribute when there should be only one `Component` of such type on the `GameObject`
+* **Methods**
+  * Use `[PublicAPI]` attribute if `public method` is exposed to be called from **outside** of the solution 
+  * Use `[UsedImplicitly]` attribute for implicitly called methods, such as calls from Unity animation events, via Unity `GameObject.SendMessage()`, `GameObject.BroadcastMessage()` and similar. 
+  * Use `[Button]` attribute from `EasyButtons` (requires reference in `.asmdef` file) instead of `[ContectMenu]` attribute. Consider using `Mode` parameter of this attribute for `Editor`/`PlayMode`-only methods.
```
sandrade-dcl, line 247:
```suggestion
  * Use `[Button]` attribute from `EasyButtons` (requires reference in `.asmdef` file) instead of `[ContextMenu]` attribute. Consider using `Mode` parameter of this attribute for `Editor`/`PlayMode`-only methods.
```

sandrade-dcl (approved):

docs/style-guidelines.md
```diff
@@ -1,20 +1,313 @@
...
+* insert one space between the comment delimiter (`//`) and the comment text
+* don't use asterix syntax for comments 🔴 `/* comment */`
+* remove commented out code
+
+### Tests
+* **Class name** should contain Tests postfix after the feature name to the class that tests the current feature
+* **Test method name**
+  * try to reflect core of the **arrange**/**act**/**assert** part of the test in the test name (especially **assert**)
+  * consider using word `Should` to describe **assert** part and word `When` to describe **act** and **arrange** part of the test
+* The method body should be split by `// arrange` - `// act` - `// assert` comments to respective blocks.
```
sandrade-dcl, line 293: This is just a silly thing that is the result of my manias but I always like to start all comments with upper case 😄

```
// Arrange

// Act

// Assert
```

popuz: good one 💪 
There is also a possibility to make them CAPITAL to distinguish them from normal comments. However, since generally there are not so many comments in tests, I would go with first_capital as you proposed

sandrade-dcl: I completely agree!

AjimenezDCL (approved):

PR #3075, opened 2022-09-20
Feat/unity 2021 shaders compatibility

### LINKED ISSUE #2706: [Unity 2021] Transparency not working on WebGL
Check branch feat/unity-2021
The problem was found here #2602

Transparencies rendering issue [Happens only in build]
This can be seen in Genesis Plaza's waterfall pipe that takes to the main hall, as well as in some user's wearables:

It seems the alpha cutoff is not working correctly, there is no error in the shader while building.

MichaelBitsakis:
## What does this PR change?

...

## How to test the changes?

You can apply this to a material in a demo scene on a cube and test it in a test build.

1. Go to: https://play.decentraland.zone/?renderer-branch=feat/unity-2021-shadersCompatibility
2. ...

## Our Code Review Standards


Kinerius (changes requested):
Looking Good!

I would clean up this PR a bit, we don't need the test scene and the new materials since the usage of `DCL/UniversalRenderer/Lit` is mostly used by code and materials are created that way.

Kinerius (approved):

aixaCode:
Closing, as this is not valid anymore.

Issue #2972, opened 2022-09-06
An empty field area flashes on the screen after initiating the teleportation


kwspgoliasz:
**Description:**
The loading screen for the teleport is preceded by an empty field scene briefly flashing on the screen.

**Steps to reproduce:**

1. Load into the game as a guest or via a wallet.
2. Teleport to any field by using the Map ('M' button) or a chat command ('**/goto random**' or '**/goto X,Y**')
3. Observe the issue.

**Results:**
A scene showing an empty field can be briefly observed right after initiating the teleportation.

**Expected results:**
Lack of any visual issues when using the teleportation.

**Notes:**

- The issue occurs on both the guest and wallet accounts.

- The issue occurs on both the production and development configurations.

- The empty field scene is displayed for varying periods of time depending on the attempt.

- Please see the attached video for reference: https://images.zenhubusercontent.com/337227404/43fb2c1e-d622-499c-af6f-ed3eff619bad/_2972.mp4.

- Platform: Web Client, Windows Desktop Client.

**Acceptance criteria**

- [x] Screen does not show a empty camera flash after teleport

Issue #1615, opened 2021-12-03
Loading screen fade-in animation problems


pravusjif:
The loading screen feels super broken due to its fading animations and the timings that animation has with the world teleportation:

1. When a new user is created we see the broken world before the fade-in animation of the loading screen...

https://user-images.githubusercontent.com/1031741/182938938-4bdaa2d3-2c60-4b16-936d-8d529abf1ff9.mp4

2. When a user executes a jump-in using the scene-triggered jump-in prompt the same problem can be seen...

https://user-images.githubusercontent.com/1031741/182939245-555dbed1-7cb3-430e-9086-6a62db5d9f0b.mp4

3. When a user executed a jump-in from the Explore panel, the panel closes showing the world, then the world is unloaded and then the loading screen fade-in animation starts...

https://user-images.githubusercontent.com/1031741/182939310-fde380db-67cb-404f-953e-8d1bcaa3df72.mp4

4.  The same happens doing a jump-in from the world map...

https://user-images.githubusercontent.com/1031741/182939490-e06a1e72-125f-4241-b9e6-a002fe9cc61c.mp4

-----------------------

### I think all of this would be fixed with 2 changes
- Every time a teleportation is triggered from a full-screen UI (like the Explore panel or the Worldmap) the loading screen should appear instantly (no fade-in animation)
- Every time a teleportation is triggered in-world (`/goto` chat command, jump-in prompts, etc.) the loading screen should have the fade-in animation BUT the teleportation logic should start AFTER the fade-in animation finishes...

The fade-out animation of the loading screen is never a problem, only the fade-in.

pravusjif:
This is affecting the desktop client

pravusjif:
Updated issue description and proposed solution.

The main problem to fix this issue with the loading screen fade-in is that Kernel controls the toggling of the loading screen.

These are the 2 possible approaches I can think of:
### A. Update Client-Kernel communication to take into account when we are triggering a teleport and affect the loading screen fade-in
1. Implement new `requestedFromFullScreenUI` boolean to be sent from the Client to the Kernel with every `GoTo`/`JumpIn` messages sent
2. Track that new value internally in Kernel to be able to know in which situation Kernel should tell the client to enable the loading screen instantly or using fade-in. This requires refactoring the `teleportObservable` logic and `teleportTriggered` dispatching.
3. Update Kernel->Client `SetLoadingScreen` message to have the new boolean so that Kernel can tell the Client when to set the loading screen instantly
[...]

PR #3102, opened 2022-09-23
fix: Camera flash after accepting terms

dalkia:
## What does this PR change?

Fixes the first case of #1615.  All the other cases had already been solved in #2972 

## How to test the changes?

1. Start with a fresh copy of DCL
2. Accept terms and conditions
3. Check that the terms and condition canvas stays on until the loading screen fully fades in
4. Start the tutorial normally

Its also a good opportunity to confirm that the loading screen teleports work correctly for all the cases of #1615. For all the following cases, there should be no flash after teleporting.

1. Teleport from chat
2. Teleport from the Explore panel
3. Teleport from map
4. Teleport to any event from the Jump In button present on Genesis Plaza

## Our Code Review Standards


kwssbocian (changes requested):
Issue seems to be fixed for some but not all of the outlined cases.
✔️Terms and condition canvas stays on until the loading screen fully fades in before loading.
Loading screens also work correctly for:
✔️Teleport from chat
✔️Teleport to any event from the Jump In button present on Genesis Plaza

Cases that result in a brief bright flash of a white background before the loading screen appears:
❌Teleport from the Explore panel
❌Teleport from map

davidejensen (approved):
Code looks good

PR #3143, opened 2022-09-29
refactor: ECS7 internal components lifecycle

pbosio:
*added possibility of removing an internal component while setting a default value.
for example: currently removing a "visibility" component while it was set to "invisble" will never restores the visibility state to "visible", with this PR it is possible to set the system to set renderers visible before removing the component.
+added handlers for internal component that must auto-destroy themself.
for example: a collider component should be removed when it's not holding any collider
*unify dirty components handling. previous to this PR every internal component use to handle it's own "dirty". this PR adds a system to automatically set every modified internal component as dirty and removes it's dirty state after systems are called
-some tests for deprecated components where removed, the rest of the code for those components will be removed soon


unity-renderer/Assets/DCLPlugins/ECS7/InternalECSComponents/InternalECSComponents.cs
```diff
@@ -66,4 +83,28 @@ public void Dispose()
...
+                }
+                else
+                {
+                    writeData.scene.crdtExecutor.ExecuteWithoutStoringState(writeData.entityId, writeData.componentId, null);
+                }
+            }
+            scheduledWrite.Clear();
+        };
+    }
+}
```
AjimenezDCL, line 110: This is super strange.

Why cannot `scheduleWrite` be a class variable `WriteSystem` a parameterless void method in the interface?

It will also create garbage due to the closure using a local scoped variable.

AjimenezDCL (changes requested):

unity-renderer/Assets/DCLPlugins/ECS7/InternalECSComponents/InternalECSComponents.cs
pbosio: dependency injection.
not having the method tied to a class state makes room for easier, cleaner and better tests
and at the same time it can be tied to the class state in it's implementation. 
anyway, I applied the suggested change here https://github.com/decentraland/unity-renderer/pull/3143/commits/685c90c25fec6e127d53bae91d7c3446c72bfd15

AjimenezDCL (approved):

PR #3146, opened 2022-09-30
Stop animations when character and tutorial robot are not rendered

### LINKED ISSUE #1794: "Tutorial robot" and "character preview avatar renderer" shouldn't be animated/active when not rendered
- TeacherCanvas GameObject shouldn't be active while not being rendered (after the tutorial finishes)
- _CharacterPreviewController->AvatarRenderer GameObject shouldn't be active while not being rendered (when not using the backpack)

https://images.zenhubusercontent.com/337227404/4f101ddd-846d-4246-8f10-47719b61b001/not_rendered_animated_objects.mp4

PalauiDCL:
Fixes #1794 

When the prefabs containing character animation or tutorial robot are hidden or the specific element containing the animations is hidden, the animations stops. 

QA
once we stop rendering an avatar or the tutorial robot, the preview should stop being animated.

PR in progress, more modifications to catch specific moments when we should stop animations might arise following tests and investigations.


kwspgoliasz (approved):
The feature appears to be implemented correctly, as no issues with the tutorial were observed.

PR #3147, opened 2022-09-30
Refactored culling object tracker

### LINKED ISSUE #2712: CullingObjectsTracker lazy overhead
`CullingObjectsTracker` uses a lazy approach to get all renderers when it's dirty (a few frames), when there are a lot of objects in the hierarchy this can get pretty slow. 

- [ ] Avoid the usage of `Object.FindObjectsOfType`
- [ ] Avoid the usage of LinQ to filter renderer types.

PalauiDCL:
Fixes #2712 

I decided to still use `FindObjectsOfType<T>` for checking the needed objects, I reduced the usage of these using one to check all renderers. System.Linq usage has been removed.

QA

This was not the bottleneck currently, so no improvement in performance should be visible (When the real bottleneck gets corrected this changes performance improvement should take effect)
The app should work exactly the same, if any problems of culling appear this PR should be revised.


Kinerius:
I've run the Performance meter tool for 10 seconds after loading all genesis plaza and online avatars.

This PR
 ```
* PERFORMANCE SCORE (0-100) -> 67
 * lowest frame time -> 29.6ms
 * average frame time -> 39.2ms
 * highest frame time -> 100.0ms
 * 1 percentile frame time -> 30.1ms
 * 50 percentile frame time -> 36.5ms
 * 99 percentile frame time -> 100.0ms
 * error percentage -> ±3.1%
 * total hiccups (>0.05ms frames) -> 21 (8.2% of frames were hiccups)
 * total hiccups time (seconds) -> 1.375996
 * total frames -> 256
 * total frames time (seconds) -> 10.0271
 * lowest allocations (kb) -> 0
 * highest allocations (kb) -> 0
 * average allocations (kb) -> 0
 * total allocations (kb) -> 0
```
 
 Prod
```
  * PERFORMANCE SCORE (0-100) -> 72
 * lowest frame time -> 31.6ms
 * average frame time -> 46.6ms
 * highest frame time -> 73.8ms
 * 1 percentile frame time -> 35.5ms
 * 50 percentile frame time -> 45.8ms
 * 99 percentile frame time -> 67.7ms
 * error percentage -> ±2.0%
 * total hiccups (>0.05ms frames) -> 56 (26.05% of frames were hiccups)
 * total hiccups time (seconds) -> 3.13839
 * total frames -> 215
 * total frames time (seconds) -> 10.01649
 * lowest allocations (kb) -> 0
 * highest allocations (kb) -> 0
 * average allocations (kb) -> 0
 * total allocations (kb) -> 0
```

Conclusion:
- There was a big hiccup for some reason while doing the first test on the PR so the score is lower
- What matters here are the lower frame times:
```
lowest frame time (PR vs PROD) => 6.32 % less
average frame time (PR vs PROD) => 15.87 % less
```

Good job!

File: unity-renderer/Assets/Rendering/Culling/CullingObjectsTracker.cs [hunk 1]
```diff
@@ -34,17 +36,23 @@ public IEnumerator PopulateRenderersList()
...
+                    && !(renderer is ParticleSystemRenderer))
+                {
+                    if (renderer is SkinnedMeshRenderer)
+                        skinnedRenderersList.Add((SkinnedMeshRenderer)renderer);
+                    else
+                        renderersList.Add(renderer);
+                }
+            }
+            renderers = renderersList.ToArray();
+            skinnedRenderers = skinnedRenderersList.ToArray();
```
Kinerius, line 55: I think we can improve this even further by reducing the allocations, can we change the type of those fields to `IEnumerator<Renderer>` and `IEnumerator<SkinnedMeshRenderer>`

Kinerius (changes requested):

unity-renderer/Assets/Rendering/Culling/CullingObjectsTracker.cs
PalauiDCL: @Kinerius  Of course ^^ I am out of office but tomorrow this changes will be integrated

Kinerius (approved):
LGTM, nice job! 🔥

PR #3164, opened 2022-10-03
fix: add wait for loading screen on teleport

dalkia:
## What does this PR change?

Fixes the first case of https://github.com/decentraland/unity-renderer/issues/1615. Adds a new interaction for kernel, in which kernel waits for the loading screen to be visible before taking any action on the teleport.

Kernel branch: https://github.com/decentraland/kernel/pull/562

## How to test the changes?

TEST USING https://play.decentraland.zone/?renderer-branch=fix/Add-Fade-Function-For-Kernel&kernel-branch=fix/added-wait-function-for-loading

1. Start with a fresh copy of DCL
2. Accept terms and conditions
3. Check that the terms and condition canvas stays on until the loading screen fully fades in
4. Start the tutorial normally
5. Its also a good opportunity to confirm that the loading screen teleports work correctly for all the cases of https://github.com/decentraland/unity-renderer/issues/1615. For all the following cases, there should be no flash after teleporting.

Teleport from chat
Teleport from the Explore panel
Teleport from map
Teleport to any event from the Jump In button present on Genesis Plaza

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/LoadingHUD/Bridge/LoadingBridge.cs
```diff
@@ -26,4 +36,33 @@ public void SetLoadingScreen(string jsonMessage)
             DataStore.i.HUDs.loadingHUD.message.Set(payload.message);
         DataStore.i.HUDs.loadingHUD.showTips.Set(payload.showTips);
     }
+    
+    public void FadeInLoadingHUD(string jsonMessage)
+    {
+        StartCoroutine(WaitForLoadingHUDVisible(jsonMessage));
+    }
+    
+    IEnumerator WaitForLoadingHUDVisible(string jsonMessage)
```
AjimenezDCL, line 45: I'm not sure this is the place for this logic. Bridges should have no logic whatsoever, they should just put info in the datastore or call stuff.

In any case, this is temporary until we own full control of the loading flow in the renderer so we can let it pass through. In any case let's add comments and TODOs.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/NavMap/NavmapView.cs
```diff
@@ -1,29 +1,24 @@
-using System;
+using UnityEngine;
```
AjimenezDCL, line 1: Please remove the formating/cleanup rearrangment of the code in this class.

It's hard to read the changes.

AjimenezDCL (changes requested):
Left some comments, but it looks good

AjimenezDCL (approved):

kwspgoliasz (approved):
The feature appears to be implemented correctly. On the kernel branch, the empty camera flash was not observed when teleporting, or during the transition after the T&C.
When testing without the kernel branch, the issue with the empty camera returns for all of the teleportation methods. Regardless, the transition after T&C is functioning correctly even without the kernel branch.

Issue #2381, opened 2022-02-15
Merge both debug panels into one


nearnshaw:
Today we have two different panels in preview mode that show separate lists of things that are grouped arbitrarily. 

It's confusing to know which is which, and where to find what data. They could both fit into a single panel.

See design mockups for the new panel here:
https://www.figma.com/file/HvW1lICE214xtY3pex5R5P/DebugTools?node-id=42%3A230

This panel has expandable sections so that it can take up less space.
Notice that the FPS section shows the FPS number even when the section is minimized.

Nice to have: If constantly calculating the info on this panel takes up significant processing power, could we maybe not calculate the metrics of the sections that are currently hidden?

We should also remove the notification on the top-right corner that says "Y for pannel", since we could keep toggling this main panel from the preview debug menu

**Acceptance criteria**

- [x] Merge both panels into one according to FIGMA design
- [x] Having the ability to minimize section
- [ ] Add the possibility to toggle colliders and box boundaries

PR #3180, opened 2022-10-06
feat: combined fps panel

dalkia:
## What does this PR change?

Adds a new debug panel according to #2381. 

For correct functionality, we are going to wait till the notification system is merged to merge this new debug panel. Right now, the debug panel is covering completely the chat area, which is not ideal. But once the notification area is implemented, that area will dissappear; and we are going to be able to have both spaces coexisting at the same time.

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=feat/new-debug-panel&kernel-branch=feat/memory-metrics-for-debug-panel
2. Input `/showfps` in the chat window. 
3. Check that you see the panel
4. Try to drag and drop the window
5. Check that all subpanels are closable when clicking the arrow

COMMENT: Its expected in WEBGL that the Graphics Memory value remains empty

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Debugging/FPSDisplay/FPSDisplay.cs
```diff
@@ -24,22 +29,111 @@ public class FPSDisplay : MonoBehaviour
...
+            updateValueDictionary.Add(DebugValueEnum.General_Realm, () => currentRealmValue.ToUpper());
+            updateValueDictionary.Add(DebugValueEnum.General_NearbyPlayers, () => lastPlayerCount.ToString());
+            
+            updateValueDictionary.Add(DebugValueEnum.FPS, GetFPSCount);
+            updateValueDictionary.Add(DebugValueEnum.FPS_HiccupsInTheLast1000, () => $"{fpsColor}{performanceData.Get().hiccupCount.ToString()}</color>");
+            updateValueDictionary.Add(DebugValueEnum.FPS_HiccupsLoss, GetHiccupsLoss);
+            updateValueDictionary.Add(DebugValueEnum.FPS_BadFramesPercentiles, () => $"{fpsColor}{((performanceData.Get().hiccupCount) / 10.0f).ToString(NO_DECIMALS)}%</color>");
+            
+            updateValueDictionary.Add(DebugValueEnum.Scene_Name, () =>  GetStringID());
+            updateValueDictionary.Add(DebugValueEnum.Scene_ProcessedMessages, () =>  $"{totalMessagesCurrent} of {totalMessagesGlobal}");
```
Kinerius, line 99: I would change this to:

`%{totalMessagesCurrent / totalMessagesGlobal * 100}`

 since those big numbers wont give us much information, if that number falls from 100 then we have a problem.

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Debugging/FPSDisplay/FPSDisplay.cs
```diff
@@ -24,22 +29,111 @@ public class FPSDisplay : MonoBehaviour
...
+            updateValueDictionary.Add(DebugValueEnum.Scene_PendingOnQueue, () =>  $"{totalMessagesGlobal - totalMessagesCurrent}");
+            updateValueDictionary.Add(DebugValueEnum.Scene_Poly, () => $"{metrics.triangles}/{limits.triangles}");
+            updateValueDictionary.Add(DebugValueEnum.Scene_Textures, () => $"{metrics.textures}/{limits.textures}");
+            updateValueDictionary.Add(DebugValueEnum.Scene_Materials, () => $"{metrics.materials}/{limits.materials}");
+            updateValueDictionary.Add(DebugValueEnum.Scene_Entities, () => $"{metrics.entities}/{limits.entities}");
+            updateValueDictionary.Add(DebugValueEnum.Scene_Meshes, () => $"{metrics.meshes}/{limits.meshes}");
+            updateValueDictionary.Add(DebugValueEnum.Scene_Bodies, () => $"{metrics.bodies}/{limits.bodies}");
```
Kinerius, line 106: I belive this is hard to understand without reading the code:
I would add a `current: ### / max: ###` format.
We can also give this some color: orange for > 90%, red for > 100%, else white.

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Debugging/FPSDisplay/FPSDisplay.cs
```diff
@@ -24,22 +29,111 @@ public class FPSDisplay : MonoBehaviour
...
+        private void SetupDictionary()
+        {
+            if (updateValueDictionary != null)
+                return;
+            
+            updateValueDictionary = new Dictionary<DebugValueEnum, Func<string>>();
+            updateValueDictionary.Add(DebugValueEnum.Skybox_Config, () => DataStore.i.skyboxConfig.configToLoad.Get());
+            updateValueDictionary.Add(DebugValueEnum.Skybox_Duration, () => $"{DataStore.i.skyboxConfig.lifecycleDuration.Get()}");
+            updateValueDictionary.Add(DebugValueEnum.Skybox_GameTime, () => $"{DataStore.i.skyboxConfig.currentVirtualTime.Get().ToString(TWO_DECIMALS)}");
+            updateValueDictionary.Add(DebugValueEnum.Skybox_UTCTime, () => $"{DataStore.i.worldTimer.GetCurrentTime()}");
```
Kinerius, line 87: I would remove those

Kinerius (changes requested):
I like this change so much <3 

This is a good opportunity to get rid of the skybox debug values since neither we nor the users need them.

This is also a good opportunity to add the current memory usage, we might have to get that number from the kernel since we don't have those exact values and we can extract that info from the js dom. Also giving this value a nice color: red > 1.8 gb, orange > 1.5 gb, else green.

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Debugging/FPSDisplay/FPSDisplay.cs
dalkia: Done. Im also thinking, does it make sense to show those numbers? Or should we just show the percentage?

Kinerius: I think they are needed for the scene creators, a % might be less informative

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Debugging/FPSDisplay/FPSDisplay.cs
dalkia: Removing!

RominaMarchetti:
@dalkia here goes my feedback :)

### Control panel window

Screenshot 2022-10-07 at 09 20 41

1- The size of the panel's title should match the size of the chat and friends window. Its size is 14. Then decrease 1 point the font size of the rest of the texts.
2- While the opacity of the control panel is 100% its color must be #161518 instead of black.
3- The height of the nav bar doesn't match with the chats and friends windows.
4- The close button container is missing. Its color is #2D2C2F and should be aligned with the arrows and text, not the cross icon.
5- Increase the gap a bit more between the text and the icon.

### Name copied toast

Screenshot 2022-10-07 at 09 31 21

1- Remove the icon from the toast. 
2- Put the dot at the end of the sentence.
3- Make the container width dependent of the text size. The text must be centered. [Here](https://www.figma.com/file/HvW1lICE214xtY3pex5R5P/DebugTools?node-id=696%3A16578) goes the example.
4- It must be aligned on top with the minimap.

RominaMarchetti:
@dalkia aesthetically speaking it's impecable.
On the other hand I saw that the window is bigger now, and honestly it might grow in the future. Having said that, I have a question: is it possible to add the feature of moving this window all around the screen? This way we tackle scalability and flexibility.

RominaMarchetti (approved):

Kinerius:
> would it be helpful to have some memory metrics when using in desktop?

Yes

Kinerius (changes requested):
The changes are looking good!

I've found that if you switch scenes when the fps panel is open, it stops being updated and you have to close and open it again.

Kinerius (approved):
Nice!

kwssbocian (approved):
✔️`/showfps` command causes the Control Panel to pop up.
✔️The panel is draggable and can be moved on the screen.
✔️The subpanels collapse and expand upon clicking the arrow icon.

PR #3208, opened 2022-10-11
Refactor/culling controller refactor

### LINKED ISSUE #3196: CullingController needs to be optimised
## CullingController has several problems to be improved upon

- Calculating all avatar renderers even when an item of that avatar must be shown (Making the avatar to be rendered in full)
- Calculating too many elements when culling has been disabled
- Not caching the most costly calculations
- Others that ight arise

---

## Optimizing calculus for culling

Removing the main calculus for SkinMeshRenderer provoked a performance increase from 25%-40%, so, an optimisation of the current system could suppose an important performance increase.
+ When there are avatars present the spikes in framerate got reduced almost completely under the test conditions.

---

PalauiDCL:
fixes #3196 

Culling now optimizes a set of elements in order to have faster calculation times and increased performance as a consequence.

## How to test the changes? QA

The culling system has received an overhaul, we have to make sure the culling is still working as expected.
- Shadows should work the same as in dev.
- Objects should be drawn under the same rules as in dev
- Animations should be displayed like in dev.

### DIFF
--- a/unity-renderer/Assets/Rendering/Culling/CullingController.cs
+++ b/unity-renderer/Assets/Rendering/Culling/CullingController.cs
@@ -5,6 +5,7 @@ using UnityEngine;
 using UnityEngine.Rendering;
 using UniversalRenderPipelineAsset = UnityEngine.Rendering.Universal.UniversalRenderPipelineAsset;
 using static DCL.Rendering.CullingControllerUtils;
+using System.Linq;
 
 namespace DCL.Rendering
 {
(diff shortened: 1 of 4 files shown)


kwssbocian (approved):
Fix appears to be implemented correctly from the in-game/functionality perspective.
✔️Shadows, drawn objects and animations are all behaving the same as in dev. No additional issues have been observed.

unity-renderer/Assets/Rendering/Culling/CullingController.cs
```diff
@@ -143,52 +143,42 @@ internal IEnumerator ProcessProfile(CullingControllerProfile profile)
...
-                    // or if the player is inside the bounding box of the object
-                    || bounds.Contains(playerPosition)
-                    // or if the player distance is below the threshold
-                    || distance < profile.visibleDistanceThreshold
-                    // at last, we perform the expensive queries of emmisiveness and opaque conditions
-                    // these are the last conditions because IsEmissive and IsOpaque perform expensive lookups
-                    || viewportSize > profile.emissiveSizeThreshold && IsEmissive(r)
-                    || viewportSize > profile.opaqueSizeThreshold && IsOpaque(r)
+                    bounds.Contains(playerPosition) ||
+                    distance < profile.visibleDistanceThreshold ||
```
menduz, line 179: Prefer this as first option to prevent executing the .Contains if possible, short circuit for performance. You may also all back the comments for the future maintainers

PalauiDCL: I will be changing the order and getting back the comments for the last two lines, the lines about distance and bounds seem 100% understandable from the code and in my opinion, bring no value.

davidejensen (approved):
LGTM!

unity-renderer/Assets/Rendering/Culling/CullingController.cs
menduz: It is clear what the line in the void does. Simply a condition, but overall we are doing a poor job commenting the code for other humans.. it is bad that we are removing precious bits of information for other maintainers

PR #3221, opened 2022-10-12
refactor: Player name optimization

### LINKED ISSUE #3170: PlayerName Update Improvement
When we have 100 avatars, the PlayerName is updating its alpha every frame even if the alpha didn't change.

- [ ] We can gain a slight performance boost if we ignore all updates if the alpha is rounded up for each 0.1 and if the previous alpha is the same as the one we have now.

- [ ] We can also remove the LinQ since its allocating memory.

PalauiDCL:
fixes #3170 

Currently the alpha for the player name works in steps, this PR avoids recalculating new alpha values every frame if the alpha has not changed or has not been modified to a different step.
The only difference is the alpha will have a discrete amount of possible values instead of being a continuous one.
With this current PR the amount of steps is stored in the constant `ALPHA_STEPS`.
In order to get the discrete available values, these are all the values from 0 to 1 that are multiples of `1f / ALPHA_STEPS`

## QA

- The player name alpha should work very similar to dev / production, alpha values are as such: 0, 0.0625, 0.125... multiples of 0.0625 from 0 to 1

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/PlayerName/PlayerName.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/PlayerName/PlayerName.cs
@@ -31,12 +31,38 @@ public class PlayerName : MonoBehaviour, IPlayerName
     internal BaseVariable<float> namesOpacity => DataStore.i.HUDs.avatarNamesOpacity;
     internal BaseVariable<bool> namesVisible => DataStore.i.HUDs.avatarNamesVisible;
 
-    internal float alpha;
-    internal float targetAlpha;
-    internal float previousAlphaStep;
     internal bool forceShow = false;
     internal Color backgroundOriginalColor;
     internal List<string> hideConstraints = new List<string>();
+
+    private float alpha;
+    private float targetAlpha;
+
+
+    internal float Alpha 
+    {
+        get => alpha;
+        set
+        {
+            alpha = Mathf.Clamp01(value);
+            if (alpha < 0.01f)
+            {
+                UpdateVisuals(0);
+                SetRenderersVisible(false);
+                return;
+            }
+
+            UpdateVisuals(alpha);
+            SetRenderersVisible(true);
+        }
+    }
+
+    internal float TargetAlpha
+    {
+        get => targetAlpha;
+        set => targetAlpha = Mathf.Clamp01(value);
+    }
+
 
     private void Awake()
     {
@@ -74,14 +100,13 @@ public class PlayerName : MonoBehaviour, IPlayerName
             UpdateVisuals(0);
             return;
         }
-        
+
+        float previousAlphaStep = GetNearestAlphaStep(alpha);
         float finalTargetAlpha = forceShow ? TARGET_ALPHA_SHOW : targetAlpha;
         alpha = Mathf.MoveTowards(alpha, finalTargetAlpha, ALPHA_TRANSITION_STEP_PER_SECOND * deltaTime);
         float currentAlphaStep = GetNearestAlphaStep(alpha);
         if (currentAlphaStep == previousAlphaStep)
             return;
-
-        previousAlphaStep = currentAlphaStep;
 
         if (currentAlphaStep == 0)
         {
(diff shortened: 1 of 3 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/PlayerName/PlayerName.cs
```diff
@@ -72,18 +100,23 @@ internal void Update(float deltaTime)
...
             return;
         }
-        
+
+        float previousAlphaStep = GetNearestAlphaStep(alpha);
         float finalTargetAlpha = forceShow ? TARGET_ALPHA_SHOW : targetAlpha;
+        alpha = Mathf.MoveTowards(alpha, finalTargetAlpha, ALPHA_TRANSITION_STEP_PER_SECOND * deltaTime);
+        float currentAlphaStep = GetNearestAlphaStep(alpha);
+        if (currentAlphaStep == previousAlphaStep)
+            return;
```
Kinerius, line 109: This return is going to cause some issues regarding scaling and rotation, you might want to raise transform calls above alpha or you can also extract all of this logic into a new Method for more clarity. `UpdateVisuals` is a good candidate to have this logic also.

Kinerius (changes requested):

kwspgoliasz (changes requested):
The feature introduced several issues with the username tags displayed in-world. The issues are as follows:

- The usernames are displayed in various sizes: 


- Lack of username tag rotation - the username tag is not adjusted to other peer's camera angle: https://user-images.githubusercontent.com/110464313/195353993-3a2bab19-d6ed-4c9f-8a06-e94721a07b17.mp4.

- The username tags do not fade out - when walking away from another avatar, the username disappears instead of gradually fading out: https://user-images.githubusercontent.com/110464313/195354019-2ba80a45-17b8-4468-b0ea-bf4e583f2a72.mp4.

- The username tags are not automatically displayed - ex. after the nametag disappears due to distance between the users, it does not reappear automatically after closing the distance again: https://user-images.githubusercontent.com/110464313/195354053-244d9955-9651-4df9-a8ca-c239b8eed2ef.mp4.

kwspgoliasz (changes requested):
The problems mentioned in the previous comment were successfully fixed, however, a new issue with the name tags' fade out function was observed.
When backing away from another user, the name tag stops fading out after reaching a certain point. This results in the outline of the username being displayed even at large distances - https://user-images.githubusercontent.com/110464313/196378309-ef5dd26f-f4eb-432d-855f-5e1e2668af05.mp4.

Kinerius:
image
I just went to Wondermine and I cant see any player name, I even toggled off an on the player names at the settings menu but they wont show.

kwspgoliasz (approved):
The feature appears to be implemented correctly. All of the issues mentioned in the previous comments were fixed.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/PlayerName/PlayerName.cs
```diff
@@ -93,11 +121,37 @@ internal void Update(float deltaTime)
...
+        SetRenderersVisible(true);
+        float resolvedAlphaStep = GetNearestAlphaStep(resolvedAlpha);
+        float canvasAlphaStep = GetNearestAlphaStep(canvasGroup.alpha);
+
+        if (resolvedAlphaStep != canvasAlphaStep)
+            UpdateVisuals(resolvedAlpha);
+
+
+        // Local Methods
+        bool ChechHide(float checkAlpha)
```
Kinerius, line 141: Not only this has a typo but the name also tells me that you are checking if something is hidden if alpha is < MINIMUM_ALPHA_TO_SHOW. This method also has the side effect of setting the renderer off if the condition is met.
We can give it a more proper name like `GetAndUpdateVisibility`

Kinerius (approved):
It works fine for me! 
I added one last thingy.

sandrade-dcl (approved):
Looks good to me!

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/PlayerName/PlayerName.cs
PalauiDCL: Applying your commenet and merging after tests ^^

PR #3230, opened 2022-10-12
fix: hotfix for NRE preventing backpack from loading

Antivortex:
## What does this PR change?

This PR fixes null reference exception happening in GetAllSlots method of EmoteSlotSelectorComponentView class for a user who has custom wearables. The exception happens during emote customization UI initialization, field emotesSlots (a serialized field of EmoteSlotSelectorComponentView) is null on that frame. This field is accessed on the same frame when game object of EmotesCustomizationSection is instantiated and the field is part of nested prefab EmoteSlotSelector but accessed with a chain of public methods before nested prefab EmoteSlotSelector is fully initialized with its instantiated parent.

Solution is to unpack EmoteSlotSelector prefab and make it part of EmotesCustomizationSection. This is the only usage of this nested prefab.

## How to test the changes?

Test everything regarding emotes and wearables:

1. If custom emotes and wearables are being selected, saved and loaded without issues.
2. If custom emotes and wearables appear on an avatar from his perspective and peer perspective.
3. If both embedded and custom emotes play without any issues.

## Our Code Review Standards

### DIFF
--- a/ci-build.sh
+++ b/ci-build.sh
@@ -16,6 +16,7 @@ xvfb-run --auto-servernum --server-args='-screen 0 640x480x24' $UNITY_PATH/Edito
   -customBuildTarget "$BUILD_TARGET" \
   -customBuildName "$BUILD_NAME" \
   -customBuildPath "$BUILD_PATH" \
+  -customBuildOptions "Development" \
   -executeMethod BuildCommand.PerformBuild
 
 UNITY_EXIT_CODE=$?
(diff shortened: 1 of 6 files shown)


sandrade-dcl (approved):
Looks good to me!

AjimenezDCL (approved):

Antivortex:
This one turned out to be an editor only bug so it makes sense to move it to the dev branch rather than shpping as a hotfix

PR #3235, opened 2022-10-12
feat: ECS7 avatar texture

pravusjif:
[protocol PR](https://github.com/decentraland/protocol/pull/63)
[sdk toolchain PR](https://github.com/decentraland/js-sdk-toolchain/pull/285)
[Kernel PR](https://github.com/decentraland/kernel/pull/618)

Material Component is now configured either with a `texture` (e.g. `texture: { src: "images/scene-thumbnail.png" }`) or with an `avatarTexture` (e.g. `avatarTexture: { userId: "USER_ID_HERE" }`)

### ECS7 Test Scene
Replace `USER_ID_HERE` with a real user id address.

```
function createCube(x: number, y: number, z: number): Entity {
  const entity = engine.addEntity()

  Transform.create(entity, {
    position: { x, y, z },
    scale: { x: 1, y: 1, z: 1 },
    rotation: { x: 0, y: 0, z: 0, w: 1 }
  })

  const mat = Material.create(entity, {
    texture: {   
      texture: undefined,
      avatarTexture: {
        userId: "0x51777c0b8DBA8B4dfE8a1c3d0a1eDAA5B139B4e0" 
      }
    }
  })

  MeshRenderer.create(entity, { box: { uvs: [] } })
  MeshCollider.create(entity, { box: {} })

  return entity
}
```

Can be tested installing [the sdk-toolchain PR](https://github.com/decentraland/js-sdk-toolchain/pull/285) npm package in the root of the scene and running the scene locally.

https://user-images.githubusercontent.com/1031741/195731384-75e6e052-bfdd-4c9a-84fe-9a7562e917e0.mp4


unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Material/MaterialHandler.cs
```diff
@@ -32,9 +32,25 @@ public void OnComponentModelUpdated(IParcelScene scene, IDCLEntity entity, PBMat
...
-            AssetPromise_Material_Model? promiseModel = null;
-            AssetPromise_Material_Model.Texture? albedoTexture = CreateMaterialPromiseTextureModel(model.Texture, scene);
+            if (model.AlbedoTextureCase == PBMaterial.AlbedoTextureOneofCase.AvatarTexture)
+            {
+                string avatarTexUrl = Configuration.APIUrls.AVATAR_TEXTURE_API_URL + model.AvatarTexture.UserId;
+                AssetPromise_Material_Model.Texture materialModel = new AssetPromise_Material_Model.Texture(avatarTexUrl,
+                    (UnityEngine.TextureWrapMode)model.AvatarTexture.GetWrapMode(),
+                    (UnityEngine.FilterMode)model.AvatarTexture.GetFilterMode());
+
+                CreateAndConfigureMaterialPromise(scene, entity, model, materialModel);
```
pbosio, line 42:
```suggestion
                AssetPromise_Material_Model.Texture albedoTextureModel = new AssetPromise_Material_Model.Texture(avatarTexUrl,
                    (UnityEngine.TextureWrapMode)model.AvatarTexture.GetWrapMode(),
                    (UnityEngine.FilterMode)model.AvatarTexture.GetFilterMode());

                CreateAndConfigureMaterialPromise(scene, entity, model, albedoTextureModel);
```

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Material/MaterialHandler.cs
```diff
@@ -32,9 +32,25 @@ public void OnComponentModelUpdated(IParcelScene scene, IDCLEntity entity, PBMat
...
+
+                return;
+            }
 
+            CreateAndConfigureMaterialPromise(scene, entity, model, CreateMaterialPromiseTextureModel(model.Texture, scene));
```
pbosio, line 47:
```suggestion
            AssetPromise_Material_Model.Texture? albedoTextureModel = CreateMaterialPromiseTextureModel(model.Texture, scene);
            CreateAndConfigureMaterialPromise(scene, entity, model, albedoTextureModel);
```

pbosio (approved):

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Material/MaterialHandler.cs
pravusjif: Done

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Material/MaterialHandler.cs
pravusjif: Done

protocol-gen/package.json
```diff
@@ -17,7 +17,7 @@
     "typescript": "^4.2.3"
   },
   "dependencies": {
-    "@dcl/protocol": "^1.0.0-3250759085.commit-6cf3623",
+    "@dcl/protocol": "https://sdk-team-cdn.decentraland.org/@dcl/protocol/branch//dcl-protocol-1.0.0-3267930139.commit-2b6939d.tgz",
```
AjimenezDCL, line 20: Is this expected?

unity-renderer/Assets/Scripts/MainScripts/DCL/Configuration/Configuration.cs
```diff
@@ -236,4 +236,9 @@ public static class PhysicsLayers
         public static int playerInfoCardMenu = LayerMask.NameToLayer("PlayerInfoCardMenu");
         public static int avatarTriggerMask = LayerMask.GetMask("AvatarTriggerDetection");
     }
+
+    public static class APIUrls
+    {
+        public const string AVATAR_TEXTURE_API_URL = "https://synapse.decentraland.org/profile-pictures/";
```
AjimenezDCL, line 242: As a rule of thumb we rarely compose urls ourselves in the renderer. Shouldnt kernel provide this with the correct environment?

AjimenezDCL (approved):
Approving with a concern.

protocol-gen/package.json
pravusjif: Yes, I will actually update that dependency again after we merge the [Protocol PR](https://github.com/decentraland/protocol/pull/63), meanwhile this is poiting to the package build in that PR.

lorux0 (approved):

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Material/MaterialHandler.cs
```diff
@@ -31,10 +31,28 @@ public void OnComponentModelUpdated(IParcelScene scene, IDCLEntity entity, PBMat
...
+                albedoTextureModel = new AssetPromise_Material_Model.Texture(avatarTexUrl,
+                    (UnityEngine.TextureWrapMode)model.AvatarTexture.GetWrapMode(),
+                    (UnityEngine.FilterMode)model.AvatarTexture.GetFilterMode());
 
-            AssetPromise_Material_Model? promiseModel = null;
-            AssetPromise_Material_Model.Texture? albedoTexture = CreateMaterialPromiseTextureModel(model.Texture, scene);
+                CreateAndConfigureMaterialPromise(scene, entity, model, albedoTextureModel);
+
+                return;
+            }
```
AjimenezDCL, line 46: This doesn't scale at all and violates the single responsibility principle.  This material does (and is) too many things, 

Doesnt it make sense to have different materials as we used to have? PBRMaterial, AvatarMaterial, UnlitMaterial or whatever?

AjimenezDCL (changes requested):

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Material/MaterialHandler.cs
menduz: Here are references to this `oneof` occurrence
- https://github.com/decentraland/protocol/blob/main/proto/decentraland/sdk/components/mesh_renderer.proto#L24-L29

To map this to a concept, ideally we have a single `oneof` for each interface like this


```proto
message ITexture {
  oneof tex {
    ImageTexture image = 1;
    AvatarTexture avatar = 2;
    GltfTexture avatar = 3;
    // ResourceManagerTexture from_resource = 4;
  }
}

message Material {
  ITexture albedo_texture = 1;
  ITexture alpha_texture = 2;
}
```

And then in the implementation, the component decides how to get the specific (real) resource based on the union type like this

```ts

function getResourceFromTexture(protoTexture: PB.ITexture) {
  switch (......)
}

function updateMaterial(materialModel) {
  realMaterial.texture = getResourceFromTexture(materialModel.albedoTexture)
  realMaterial.alphaTexture = getResourceFromTexture(materialModel.alphaTexture)
}
```

That is extremely important to enable protocol extensibility and new kinds of textures. We can see clearly that once we need VideoTexture, it can be added to the `oneof`

```diff
 message ITexture {
   oneof tex {
     ImageTexture image = 1;
     AvatarTexture avatar = 2;
     GltfTexture avatar = 3;
+    VideoTexture video = 4;
   }
 }
```

AjimenezDCL: I was not talking about the protocol but its implementation.

My scalability concern comes from the `Handler` class. Once we have 10 different ITexture we will have a giant switch that needs maintenance. We can wrap that in a pure function (that conveniently fits really well your ITexture definition). 

A few benefits from this decision would be:
- Avoid maintenance in the handler.
- Encapsulate the responsibility.
- Wrap the smelly enum in a controled environment to be replaced in the near future (see below).

Once we upgrade to Unity 2021 (C# 9.0) we will be able to use pattern matching and also ADTs 
https://docs.unity3d.com/Manual/CSharpCompiler.html
https://ecsharp.net/lemp/pattern-matching.html.

This should allow us to mimic the protocol with an ITexture and use pattern matching to get either:
- Diferent material: I still think for Unity they should be different materials (therefore the wrapper). It would simplify stuff like shaders properties and also make the flow lighter from the component downwards (i.e. unlit materials not needing metallic, normals, roughness maps). 
- The correct texture implementation: I dont disregard this implementation, I just find it more complex but equally suitable. Maybe pattern matching would also be useful here to reduce that complexity.


Performance Wise: if your benchmarks show this design is more efficient for both the protocol and the comunication; I fully believe it. But once we hit C# this design will bring negligible degradation in stuff like memory allocation; C# is gonna reserve memory to allocate the whole model (or at least the pointers) , even if only one of the `oneof` is being set. 
As I said, in this case it is negligible and it doesnt matter, but if we bring performance to the table take renderer and its language idiosyncrasies into account as well.

AjimenezDCL: Also I just remembered this article (9 years ago) https://tinylabproductions.wordpress.com/2013/09/22/simulating-algebraic-data-types-aka-tagged-unions-and-pattern-matching-them-in-c/.

There was a cool presentation about this and Unity, gotta try to find it, but to summarize. Performance was not so great (C# 9.0 is solving this for sure).

AjimenezDCL: Way cleaner! and now it can be tested separatedly =D!

AjimenezDCL (approved):

PR #3245, opened 2022-10-13
fix: ECS7 AudioStream not playing on start

pbosio:
since browser can't reproduce media without any user input
we have the issue that most of the time AudioStream can't play when user spawn on the scene
to mitigate this situation we have added the condition for the pointer to be locked before trying to play stream's audio


sandrade-dcl (approved):
Looks good to me!

PR #3248, opened 2022-10-14
fix: nested prefab null reference exception fix

Antivortex:
## What does this PR change?

This PR fixes null reference exception happening in GetAllSlots method of EmoteSlotSelectorComponentView class for a user who has custom wearables. The exception happens during emote customization UI initialization, field emotesSlots (a serialized field of EmoteSlotSelectorComponentView) is null on that frame. This field is accessed on the same frame when game object of EmotesCustomizationSection is instantiated and the field is part of nested prefab EmoteSlotSelector but accessed with a chain of public methods before nested prefab EmoteSlotSelector is fully initialized with its instantiated parent.

Solution is to unpack EmoteSlotSelector prefab and make it part of EmotesCustomizationSection. This is the only usage of this nested prefab.

## How to test the changes?

Test everything regarding emotes and wearables:

If custom emotes and wearables are being selected, saved and loaded without issues.
If custom emotes and wearables appear on an avatar from his perspective and peer perspective.
If both embedded and custom emotes play without any issues.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CatalogController/CatalogController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CatalogController/CatalogController.cs
@@ -98,6 +98,8 @@ public class CatalogController : MonoBehaviour
             return;
 
         AddWearablesToCatalog(request.wearables);
+        
+        
 
         if (!string.IsNullOrEmpty(request.context))
         {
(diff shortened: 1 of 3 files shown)


sandrade-dcl (approved):

kwssbocian (approved):
Issue appears to be fixed.
✔️Custom emotes and wearables are being selected, saved and loaded without issues.
✔️Custom emotes and wearables appear on the player avatar from their own perspective as well as other players'.
✔️Both default embedded and custom emotes play without any issues.

Issue #3247, opened 2022-10-14
Video being throttled as user was not on the parcel


aixaCode:
I am seeing an odd behavior where the video plays like its being throttled (given lower fps) sometimes as if i was not on the parcel (which I know throttling is applied then).  If i walk off and back on it fixes it self usually. Its deployed here.  The video is cached on my browser so I do not think its network buffering.

The settings are controlled in DCLVideoTexture.cs.
Could be that isPlayerInScene is not working properly.

Test scene:
https://play.decentraland.org/?NETWORK=mainnet&position=53,53&CATALYST=peer-testing.decentraland.org&realm=v2~peer-testing.decentraland.org&DEBUG_SCENE_LOG=&island=Iru4

More info in thread: 
https://decentralandteam.slack.com/?redir=%2Farchives%2FC03BFMYEZK2%2Fp1665673563248619

PR #3250, opened 2022-10-14
Fix/bug video throttled even if close

0xD-Fabio:
## What does this PR change?

Fix video throttling as described in issue #3247

There were a couple of bugs that prevented the correct notification of scene change.
One was the overflow of subscriptions to events regarding parcel change and scene change.
The second and most important fix was about an oversight on the SceneController which was not checking the scene position properly.

## How to test the changes?

To test the changes, you need to check if the videos are being throttled when you enter the scene containing them, using this link:
[https://play.decentraland.zone/?renderer-branch=fix%2Fbug-video-throttled-even-if-close&NETWORK=mainnet&position=53%2C53&CATALYST=peer-testing.decentraland.org&realm=testing&DEBUG_SCENE_LOG=](url)

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
```diff
@@ -156,9 +158,18 @@ public override IEnumerator ApplyChanges(BaseModel newModel)
...
+
             string videoId = (!string.IsNullOrEmpty(scene.sceneData.id)) ? scene.sceneData.id + id : scene.GetHashCode().ToString() + id;
             texturePlayer = new WebVideoPlayer(videoId, dclVideoClip.GetUrl(), dclVideoClip.isStream, videoPluginWrapperBuilder.Invoke());
             texturePlayerUpdateRoutine = CoroutineStarter.Start(OnUpdate());
+            // unsub first
+            CommonScriptableObjects.playerCoords.OnChange -= OnPlayerCoordsChanged;
+            CommonScriptableObjects.sceneID.OnChange -= OnSceneIDChanged;
+            scene.OnEntityRemoved -= SetPlayStateDirty;
+            Settings.i.audioSettings.OnChanged -= OnAudioSettingsChanged;
+            // then resub
```
popuz, line 172: This subscribe/unsubscribes in one place doesn't fill correct for me. As I understand, this `Initialize` method will be called only once, means that it will not be any subscription before. Also, since unsubscription happens in Dispose, maybe it highlights the problem with the disposal of the object. Did you try to remove this part and check if it works the same?

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
```diff
@@ -56,6 +57,7 @@ public DCLVideoTexture()
         {
             model = new Model();
 
+            DataStore.i.virtualAudioMixer.sceneSFXVolume.OnChange -= OnVirtualAudioMixerChangedValue;
             DataStore.i.virtualAudioMixer.sceneSFXVolume.OnChange += OnVirtualAudioMixerChangedValue;
```
popuz, line 61: Same concern here. It is a consctructor, which means no subscription of  `OnVirtualAudioMixerChangedValue` of this instance of the class can be for  `DataStore.i.virtualAudioMixer.sceneSFXVolume.OnChange`. Probably again it highlights some problems with an instance not properly being disposed. Can you try to remove it and check if everything works ok without it?

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
```diff
@@ -312,11 +323,14 @@ private bool IsPlayerInSameSceneAsComponent(string currentSceneId)
...
-        {
+        {   
             SetPlayStateDirty();
         }
 
-        private void OnSceneIDChanged(string current, string previous) { isPlayerInScene = IsPlayerInSameSceneAsComponent(current); }
+        private void OnSceneIDChanged(string current, string previous) 
+        { 
+            isPlayerInScene = IsPlayerInSameSceneAsComponent(current);
+        }
```
popuz, line 333: What about using lambda expression here (to align with our code-style)?
```suggestion
        private void OnSceneIDChanged(string current, string previous) =>
            isPlayerInScene = IsPlayerInSameSceneAsComponent(current);
```

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
```diff
@@ -330,6 +344,8 @@ public override void AttachTo(ISharedComponent component)
             SetPlayStateDirty();
             attachedMaterials.Add(component.id, component);
 
+            component.OnAttach -= SetPlayStateDirty;
+            component.OnDetach -= SetPlayStateDirty;
```
popuz, line 348: this also looks a little bit like a hack. Because, as I understand, this component, if it is already subscribed, should be inside `attachedMaterials` and be rejected to attach due to the line 339. Maybe solution could be of movind line 345 to line 341. Also, unsubscritption is not happening for the remained components so maybe it makes sense to Dettach from all components on Disposale...
In any case, can you please try to remove this subsc/unsubsc and check if everything is still working correctly

popuz (changes requested):
Hi @0xD-Fabio. Good catch with the coordinates x/y 🦾Also initialization flag seems to be a proper approach. Nice job 👍 However, I am strongly disagree with subsc/unsubsc approach for the events. In 90% of cases it is considered as a hack and probably highlights some deeper broken logic with Initialization and Disposal. So I kindly ask you to re-check it please 🙏

popuz:
Can I also ask you to clean-up this class a bit with the our code-styles. Please, add explicit access modifier  `private` to `OnAudioSettingsChanged`  and `SetPlayStateDirty`, and change the to the lambda-style as mentiond above for `OnSceneIDChanged` method

popuz:
And there are 2 more Rider's suggestions
1. order on line 26: `new public class Model ` -> `public new class Model`
2. use null propagation on line 361: `if (texturePlayer != null)  texturePlayer.Pause();` -> `texturePlayer?.Pause();`

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
0xD-Fabio: done ;)

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/DCLVideoTexture.cs
0xD-Fabio: done

popuz (approved):
Thank you for applying requested changes 🙏 LGTM 🟢 
(Please, merge only after removing unneeded unsubscription in constroctor of DCLVideTexture)

kwspgoliasz (commented):
The feature appears to be implemented correctly. No throttling or stuttering was observed for the videos.

PR #3258, opened 2022-10-17
Destroy particles on finish refactor

### LINKED ISSUE #3168: DestroyParticlesOnFinish is super slow when we have a lot of avatars
When we have 100 avatars running, we have a lot of particles spawning and being destroyed.

Almost 9% of the frame time is being used by 7846 MonoBehaviuours checking if that object should be destroyed, this is costing us the same CPU time as the GPUSkinning.

We should investigate how we can deal with this without a lazy polling system. We can:
- Pool them.
- Manage their updates on a centralized update manager.
- Avoid using `!= null` and `IsAlive` as they are expensive calls.

PalauiDCL:
fixes #3168 

## Changes
Destroy particles on finish now will:
- Remove itself as a component if no particles are attached, (the variable is private and can not be updated later)
- Sectioned the calls for fixed update based on a random index for each component, as there are 10 possible indexes, performance cost should cost about 10% of the previous one.

## QA
- No behavior changes on the particles should be observable, particles should keep acting exactly the same as in production.

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/Utils/DestroyParticlesOnFinish.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/Utils/DestroyParticlesOnFinish.cs
@@ -1,14 +1,30 @@
-﻿using UnityEngine;
+using UnityEngine;
 
 public class DestroyParticlesOnFinish : MonoBehaviour
 {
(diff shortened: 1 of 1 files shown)


kwspgoliasz (approved):
The feature appears to be implemented correctly, as no issues or irregular beahviour were observed for the particles.

AjimenezDCL (approved):

PR #3261, opened 2022-10-17
fix: scene reload

pbosio:
## What does this PR change?
unloading and shortly after loading again the same scene was never destroying the `ParcelScene` `GameObject`
leaving several loaded GameObjects unused for the same scene.
*modified `ParcelScenesCleaner` to make sure scene's `GameObject` are destroyed properly.
this change shouldn't make much of the difference in prod in any way but better not to have garbage leftovers for previously loaded scenes.

## How to test the changes?
walk around the world, if everything keeps loading and unloading as it should, then we are good :)

## Our Code Review Standards


AjimenezDCL (approved):

kwssbocian (approved):
✔️No issues were observed when walking around the world and having the tiles load and unload.
No additional issues were found.

PR #3275, opened 2022-10-19
Fix: fixed duplicated private chat notifications

davidejensen:
Fixes the issue that caused notifications to be duplicated for private messages when opening the first time.

To test: 
* go to https://play.decentraland.zone/?NETWORK=goerli&ENABLE_NOTIFICATION_CHAT=&renderer-branch=fix/duplicated-notification&kernel-branch=fix%2Fchannels-profiles&realm=artemis&position=99%2C99
* Check private messages
* Open a new conversation from interacting with the notification
* Verify that notifications appear only when necessary and are not duplicated

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/NotificationMessagesHUD/ChatNotificationController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/NotificationMessagesHUD/ChatNotificationController.cs
@@ -25,6 +25,7 @@ namespace DCL.Chat.Notifications
         private BaseVariable<string> openedChat => dataStore.HUDs.openedChat;
         private CancellationTokenSource fadeOutCT = new CancellationTokenSource();
         private UserProfile ownUserProfile;
+        private List<string> notificationEntries;
 
         public ChatNotificationController(DataStore dataStore,
             IMainChatNotificationsComponentView mainChatNotificationView,
@@ -36,6 +37,7 @@ namespace DCL.Chat.Notifications
             this.userProfileBridge = userProfileBridge;
             this.mainChatNotificationView = mainChatNotificationView;
             this.topNotificationView = topNotificationView;
+            notificationEntries = new List<string>();
             mainChatNotificationView.OnResetFade += ResetFadeOut;
             topNotificationView.OnResetFade += ResetFadeOut;
             mainChatNotificationView.OnPanelFocus += TogglePanelBackground;
@@ -68,7 +70,12 @@ namespace DCL.Chat.Notifications
                     ? "nearby"
                     : message.recipient);
                 if (channel?.Muted ?? false) return;
-                
+
+                if(notificationEntries.Contains(message.messageId))
+                    return;
+                else
+                    notificationEntries.Add(message.messageId);
+
                 var peerId = ExtractPeerId(message);
                 var peerProfile = userProfileBridge.Get(peerId);
                 var peerName = peerProfile?.userName ?? peerId;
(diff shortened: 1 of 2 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/NotificationMessagesHUD/ChatNotificationController.cs
```diff
@@ -25,6 +25,7 @@ public class ChatNotificationController : IHUD
         private BaseVariable<string> openedChat => dataStore.HUDs.openedChat;
         private CancellationTokenSource fadeOutCT = new CancellationTokenSource();
         private UserProfile ownUserProfile;
+        private List<string> notificationEntries;
```
lorux0, line 28: you could use a `HashSet<string>` which allows unique elements only.. it is more performant for this kind of operation

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/NotificationMessagesHUD/ChatNotificationController.cs
```diff
@@ -68,7 +70,12 @@ private void HandleMessageAdded(ChatMessage message)
                     ? "nearby"
                     : message.recipient);
                 if (channel?.Muted ?? false) return;
-                
+
+                if(notificationEntries.Contains(message.messageId))
+                    return;
```
lorux0, line 75: i think you should not return, instead the entry should be updated with the newest data to get a consistent state in memory. For example the timestamp of the message could change which may affect the sorting. Do you think it is feasible or it would require a major refactor?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/NotificationMessagesHUD/ChatNotificationController.cs
```diff
@@ -80,23 +87,21 @@ private void HandleMessageAdded(ChatMessage message)
...
                         var publicModel = new PublicChannelMessageNotificationModel(message.messageId,
                             message.body, channel?.Name ?? message.recipient, channel?.ChannelId, message.timestamp, peerName);
 
-                        mainChatNotificationView.AddNewChatNotification(publicModel);
-
-                        if (topNotificationPanelTransform.Get().gameObject.activeInHierarchy)
+                        if(channel?.ChannelId != openedChat.Get())
                         {
-                            if(channel?.ChannelId != openedChat.Get())
+                            mainChatNotificationView.AddNewChatNotification(publicModel);
```
lorux0, line 103: what if you have the #nearby opened when you receive the message. Will you be able to see the notification after you close the window? I think it should be there, but you can confirm with product though.

lorux0 (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/NotificationMessagesHUD/ChatNotificationController.cs
davidejensen: good point, will change now

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/NotificationMessagesHUD/ChatNotificationController.cs
davidejensen: if you saw the message arriving you shouldn't also receive the notification, as it is for the top notification

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/NotificationMessagesHUD/ChatNotificationController.cs
lorux0: ok then, can you add a comment then? like: `TODO: entries may have an inconsistent state. We should update the entry with new data`

sandrade-dcl (approved):

PR #3296, opened 2022-10-21
feat: implemented mute status on nearby channel

lorux0:
## What does this PR change?

Adds the possibility to mute/unmute the nearby channel.

## How to test the changes?

1. Open the #nearby channel
2. Click on the bell icon to mute it
3. Get someone to write messages in #nearby
4. Close any chat window
5. You should not receive any chat notification from #nearby
6. Open #nearby then unmute the channel
7. You should start receiving notifications again

## Our Code Review Standards


sandrade-dcl (approved):
LGTM! The only thing missing would be to confirm with Product if we're going to need to persist the mute configuration for the nearby channel between different sessions (in that case we should store it in our local storage).

PR #3298, opened 2022-10-24
Pooling Stickers Controller

### LINKED ISSUE #3169: StickersController refactor
When we have many avatars running around and using emotes, we spawn many particles.

We need to refactor the` StickersController` class so instead of just instantiating self-destructible objects we use our pool system.

- [x] Pre allocate a good amount of particles ( 10 of each at least, too much will cost us a lot of memory )
- [x] Avoid the usage of `Instantiate`
- [x] Carefully control the lifecycle of the particles
- [ ] (BONUS) Add a new Setting on the Graphics Settings to have a slider with the max amount of particles that can be spawned at the same time, with this we can control the max allowed amount of particles to be instantiated at the same time, if a new one is required, we re-use the last one even if its being used elsewhere.

PalauiDCL:
fixes #3169 

## Task
Use the DCL pooling system to instantiate stickers instead of instantiating and destroying a virtually infinite amount of elements

## QA
- The stickers refactor changed how stickers are instantiated, but there should be no visible changes for them
- If more than 10 stickers of the same exact time exist at the same time, only the last 10 will be visible, the other ones will get destroyed. (Only 10 stickers of the same exact time can be present at a given moment)

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/StickersController/StickersController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/StickersController/StickersController.cs
@@ -1,13 +1,21 @@
+using System.Collections.Generic;
 using UnityEngine;
 
 namespace DCL
 {
     public class StickersController : MonoBehaviour
     {
+        private const int POOL_PREWARM_COUNT = 10;
+        private Dictionary<string, Pool> pools = new Dictionary<string, Pool>();
         private StickersFactory stickersFactory;
         private bool isInHideArea;
 
-        private void Awake() { stickersFactory = Resources.Load<StickersFactory>("StickersFactory"); }
+        private void Awake()
+        {
+            stickersFactory = Resources.Load<StickersFactory>("StickersFactory");
+
+            ConfigurePools();
+        }
 
         public void PlaySticker(string id)
         {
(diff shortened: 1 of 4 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/Utils/DestroyParticlesOnFinish.cs
```diff
@@ -25,6 +28,16 @@ public void FixedUpdate()
             return;
 
         if (particles != null && !particles.IsAlive())
+        {
+            if (Finished != null)
+            {
+                Finished.Invoke();
+                Finished = null;
+                gameObject.SetActive(false);
+                return;
```
AjimenezDCL, line 37: I see an issue with pattern, it's ofuscating logic, `DestroyParticlesOnFinish` now is somehow aware of the existance of the pool.

There are a couple of options to solve this. I think the best one is to have a new script `ReleaseParticlesOnFinish`. This would require the `PoolableObject` to be injected (probably via an Initialize method). A side benefit is avoiding the `Finished` event, with a reference to the poolable object, the script can release it itself.

AjimenezDCL (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/Utils/DestroyParticlesOnFinish.cs
```diff
@@ -25,6 +28,16 @@ public void FixedUpdate()
...
+            }
+
             Destroy(gameObject);
+        }
```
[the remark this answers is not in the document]
PalauiDCL: Agree

AjimenezDCL (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/StickersController/StickersController.cs
```diff
@@ -30,21 +38,49 @@ public void PlaySticker(string id, Vector3 position, Vector3 direction, bool fol
...
+            ReleaseParticlesOnFinish releaser = stickerGo.GetComponent<ReleaseParticlesOnFinish>();
+            if (releaser != null)
+                releaser.Initialize(sticker);
 
             if (followTransform)
             {
-                FollowObject emoteFollow = emoteGameObject.AddComponent<FollowObject>();
-                emoteFollow.target = transform;
-                emoteFollow.offset = position;
+                FollowObject stickerFollow = stickerGo.AddComponent<FollowObject>();
```
aixaCode, line 52: Can this cause `FollowObject.cs` to be added multiple times if the object is pooled?
If yet, then we can create and extension method, i.e `public static T GetOrAddComponent<T>(this GameObject gameObject)`

PalauiDCL: Fantastic catch!
This was not happening before because the object was destroyed, but now that we are pooling it this would bbecome an issue, correcting it

AjimenezDCL: Nice catch Aga! we have that extension method already.

PR #3326, opened 2022-10-26
fix: Culling controller refactor

PalauiDCL:
fixes PR https://github.com/decentraland/unity-renderer/pull/3208

The new Culling tracker loads the renderers information in different frames avoiding spikes, this caused an issue when an already detected renderer was removed during the process, this new PR fixes that.

Old merged PR:
Culling now optimizes a set of elements in order to have faster calculation times and increased performance as a consequence.

## How to test the changes? QA

The culling system has received an overhaul, we have to make sure the culling is still working as expected.
- Shadows should work the same as in dev.
- Objects should be drawn under the same rules as in dev
- Animations should be displayed like in dev.


PalauiDCL:
image
This is the only change compared to the last approved and merged PR

Kinerius (approved):
Nicely done!

kwssbocian (approved):
✔️Shadows, animations and object drawing appears the same way as it was before.
No additional issues have been observed.

PR #3340, opened 2022-10-27
refactor: ECS7 remove CRDT messages queueing 

pbosio:
*process CRDT messages as soon(ish) as they arrive to the CRDTService. it will await for the scene not having any pending message and do the processing of CRDT messages on the main thread.
so scene will stop sending messages until messages are processed.

*added method `bool HasScenePendingMessages(string sceneId)` to `IMessagingControllersManager` to get if a scene has any pending message to be processed.

-removed `KernelBinaryMessageDeserializer` and other related classes and moved it deserialization logic to `CRDTDeserializer`

*removed `BinaryMessageBridge` from `Bridges.prefab` and from the project, since it was replaced by the RPC server


unity-renderer/Assets/DCLPlugins/ECS7/CrdtExecutorsManager/CrdtExecutorsManager.cs
```diff
@@ -0,0 +1,91 @@
...
+                if (scene.crdtExecutor == null)
+                {
+                    cachedCrdtExecutor = new CRDTExecutor(scene, componentsManager);
+                    scene.crdtExecutor = cachedCrdtExecutor;
+                    crdtExecutors[sceneId] = cachedCrdtExecutor;
+                }
+                else
+                {
+                    cachedCrdtExecutor = scene.crdtExecutor;
+                }
```
AjimenezDCL, line 79: We can probably encapsulate this in a method `GetCachedExecutor`

unity-renderer/Assets/Scripts/MainScripts/DCL/CRDTProtocol/Executor/CRDTExecutor.cs
```diff
@@ -26,6 +26,9 @@ public CRDTExecutor(IParcelScene scene, ECSComponentsManager componentsManager)
 
         public void Dispose()
         {
+            if (disposed)
+                return;
```
AjimenezDCL, line 30: We should assert above that we are not doing any operation if `disposed` (calling `Assert` so the  comprobation doesn't occur on production).

AjimenezDCL (approved):
Approving with small requests! good job!

unity-renderer/Assets/DCLPlugins/ECS7/CrdtExecutorsManager/CrdtExecutorsManager.cs
pbosio: https://github.com/decentraland/unity-renderer/pull/3340/commits/5e2293302ee88491208faa2e0596a9205f9eb476

PR #3353, opened 2022-10-28
chore: explorev2 performance

AjimenezDCL:
Replaced most of animators with simple animations by tweens.

# Test
Test extensively most of UIs of the experience andd compare them with production. No visible change should appear.
Pay extra attention to the Explore menu, the fade animations, and the cards focus/idle states.


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Common/ShowHideAnimator/ShowHideAnimator.cs
```diff
@@ -1,78 +1,80 @@
...
+    public event System.Action<ShowHideAnimator> OnWillFinishHide;
+    public event System.Action<ShowHideAnimator> OnWillFinishStart;
+
     public bool hideOnEnable = true;
     public float animSpeedFactor = 1.0f;
     public bool disableAfterFadeOut;
-    public string visibleParam = "visible";
 
-    private Animator animatorValue;
+    [SerializeField] private CanvasGroup canvasGroup;
```
sandrade-dcl, line 15: Why we need to serialize the `canvasGroup` field in the inspector if you are already caching it in the `Awake()` method? It could be private directly, no?

sandrade-dcl (changes requested):
Love the improvement of replacing our ShowHide animators by tweens ❤️

But now, with the new approach, I feel everything is opening/closing instantly (the exploreV2 menu, the transition between sections, the events and place cards...) instead of using a quick fade in/out effect. Be careful with this because this kind of effects were aligned with Product during the implemetation of the ExploreV2 menu.

I guess we should have to modify a bit the code of the current tweens in order to make them similar to our old version of the default ShowHide animator.

sandrade-dcl (changes requested):
Apart from that, it seems a lot of tests broke after this refactor.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Common/ShowHideAnimator/ShowHideAnimator.cs
AjimenezDCL: I dont really know hahaha, probably leftovers from some tests. Gotta revert it.

sandrade-dcl (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Common/ShowHideAnimator/ShowHideAnimator.cs
AjimenezDCL: I remembered now, some items were instantiated and worked with without time for the Awake to be called. As a workaround I assigned the reference in the prefab. Im putting a check in the awake to avoid the `GetComponent` if is it assigned.

kwspgoliasz (approved):
Feature appears to be implemented correctly. The UI menu is interactable and specific sections are properly highlighted.

PR #3370, opened 2022-10-31
chore: change web socket message from Send to SendAsync

Kinerius:
## What does this PR change?

Changed how we send messages through web sockets for the unity editor and desktop, they won't block the main thread now.

## How to test

You can't, this should increase the performance a little bit for unity and desktop.

## Our Code Review Standards


Kinerius:
This is too risky and its going to change in the future

PR #3393, opened 2022-11-03
Chore: improved avatar revealer script

### LINKED ISSUE #3383: BaseAvatarReveal overhead when idle
## Problem

After the avatar is shown, the avatar revealer is still updating some properties to their shaders, this has a clear overhead that we need to get rid of, In this case, we have 110 avatar revealers updating the same values over and over again with no purpose.

## Solution

Fix BaseAvatarReveal to: 
- Avoid using MonoBehaviour's Update() method by registering this update to an Update Manager
- Unregister the update once the transition to the visible avatar is complete.
- The resulting overhead of this class should be 0 when idling and the avatar is loaded and visible.

davidejensen:
## What does this PR change?

Fix #3383
Removed the Update method from the avatar revealer

...

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=chore/avatar-reveal-opt
2. Verify that all the avatars load as previously with the hologram shader
3. Go in different crowded areas and check that everything is looking normal for avatars

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Rendering/LoadingAvatar/CrossSection/BaseAvatarReveal.cs
+++ b/unity-renderer/Assets/Rendering/LoadingAvatar/CrossSection/BaseAvatarReveal.cs
@@ -25,8 +25,6 @@ public class BaseAvatarReveal : MonoBehaviour, IBaseAvatarRevealer
 
     private ILOD lod;
 
-    public float revealSpeed;
-
     public float fadeInSpeed;
     Material _ghostMaterial;
 
@@ -50,7 +48,7 @@ public class BaseAvatarReveal : MonoBehaviour, IBaseAvatarRevealer
         {
             _materials.Add(r.material);
         }
-        StartCoroutine(FadeInGhostMaterial());
+        FadeInGhostMaterial();
     }
 
     public SkinnedMeshRenderer GetMainRenderer()
@@ -96,10 +94,14 @@ public class BaseAvatarReveal : MonoBehaviour, IBaseAvatarRevealer
             isRevealing = true;
             animation.Play();
             await UniTask.WaitUntil(() => !isRevealing, cancellationToken: cancellationToken).AttachExternalCancellation(cancellationToken);
-            StopCoroutine(updateAvatarRevealerRoutine);
         }
         catch(OperationCanceledException)
         {
+            SetFullRendered();
+        }
+        finally
+        {
+            StopCoroutine(updateAvatarRevealerRoutine);
             SetFullRendered();
         }
     }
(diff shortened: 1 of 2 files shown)


unity-renderer/Assets/Rendering/LoadingAvatar/CrossSection/BaseAvatarReveal.cs
```diff
@@ -121,13 +96,37 @@ public async UniTask StartAvatarRevealAnimation(bool withTransition, Cancellatio
...
 
+    private IEnumerator UpdateRevealerPosition()
+    {
+        _ghostMaterial.SetVector("_RevealPosition", revealer.transform.localPosition);
+        foreach (Material m in _materials)
+        {
+            m.SetVector("_RevealPosition", -revealer.transform.localPosition);
+        }
+        yield return new WaitForSeconds(0.01f);
+        updateAvatarRevealerRoutine = StartCoroutine(UpdateRevealerPosition());
```
Kinerius, line 115: Wouldn't be much clearer to just do a `while(true)` with a `yield return null;` at the end of it?

Kinerius (approved):
I like this change, I left a comment

unity-renderer/Assets/Rendering/LoadingAvatar/CrossSection/BaseAvatarReveal.cs
```diff
@@ -121,13 +96,37 @@ public async UniTask StartAvatarRevealAnimation(bool withTransition, Cancellatio
...
+
+    private IEnumerator FadeInGhostMaterial()
+    {
+        Color gColor = _ghostMaterial.GetColor("_Color");
+        Color tempColor = new Color(gColor.r, gColor.g, gColor.b, gColor.a + Time.deltaTime * fadeInSpeed);
+        _ghostMaterial.SetColor("_Color", tempColor);
+        yield return new WaitForSeconds(0.01f);
+        if(_ghostMaterial.GetColor("_Color").a < 0.9f)
+        {
+            StartCoroutine(FadeInGhostMaterial());
```
lorux0, line 126: i would avoid the "recursive" coroutine, instead do a loop?
```
while (_ghostMaterial.GetColor("_Color").a < 0.9f)
{
    // update the material property
    yield return null;
}
```

unity-renderer/Assets/Rendering/LoadingAvatar/CrossSection/BaseAvatarReveal.cs
lorux0: agree

lorux0 (changes requested):

kwspgoliasz (changes requested):
The fix introduced an issue with avatars not fully loading. In most cases, only the feet are spawned for the broken avatars, however, sometimes the head or torso gets partially cut off instead.
The easiest way to reproduce the issue is to teleport to a crowded location (ex. Ice Poker) and wait for the avatars to spawn - https://user-images.githubusercontent.com/110464313/199946459-3bead160-bca2-43c2-af0a-4edafbb70cf7.mp4.

lorux0 (approved):

kwspgoliasz (approved):

PR #3403, opened 2022-11-03
fix: chat taskbar navigation

lorux0:
## What does this PR change?

We have a bug only the first time you click on the chat button in the taskbar, it opens the nearby channel instead of the conversation list. This is worst now that we have the toast for promoting channels. This change will force always to open the conversation list when clicking the chat buton in the taskbar.

## How to test the changes?

1. Click the chat button in the taskbar
2. Should open the conversation list
3. Navigate between different chats: privates, channels and nearby
4. Press ENTER, it should open the last active chat window
5. Click the chat button in the taskbar, it should open the conversation list

## Our Code Review Standards


kwssbocian (approved):
Issue appears to be fixed.
✔️Taskbar chat icon now always opens up the conversation list.
✔️Pressing Enter now always opens up the Chat window with the last active conversation.

davidejensen (approved):
looks good

PR #3415, opened 2022-11-04
fix: removed unnecessary chat allocations

lorux0:
## What does this PR change?

Removed chat memory allocations from `ChatController` which were unnecessary to keep, since we retrieve the historic of channels & DMs from kernel. #nearby messages are never cleared.

## How to test the changes?

1. Open different direct messages
2. The last messages should be loaded correctly
3. Open direct messages with more than 30 messages in history
4. Scroll up, older messages should be loaded correctly
5. Open any custom channel from the conversation list, it should load the last 30 messages correctly
6. Open #nearby channel
7. Write & receive messages
8. Close the window and open any other
9. Open the #nearby window again, messages should remain

## Our Code Review Standards


kwspgoliasz (approved):

PR #3420, opened 2022-11-07
Refactor: optimize billboards with controller

### LINKED ISSUE #3363: Billboards optimization
## Moving the camera

When moving the camera all billboards get updated with the same information at the same time, that's 255 calls every frame while moving the camera at -66,-55 during Halloween event, there are 255 billboards there to test.

## Pasive degradation

Billboards also contain polling logic that are pretty slow for what they do.

## What to do?

Since all those components do the same thing, we can create a system to do the same thing in one single update to avoid most of the overhead of calling hundreds of methods.

PalauiDCL:
fixes #3363 

This PR created a Billboard controller in order to centralize calls and optimize the process, certain previous elements have been found redundant and have been eliminated correctly.

##QA

The billboards should work exactly the same as before.


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Billboard/BillboardsController.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            List<Billboard> currentBillboards = new List<Billboard>(billboards);
+            int billboardCount = 0;
+            yield return waitForFrameEnd;
+
+            camUpdated = lastCamPosition != CameraPosition;
+            lastCamPosition = CameraPosition;
+            foreach (Billboard billboard in currentBillboards)
+            {
+                billboardCount++;
```
Kinerius, line 78: What is this Count for?

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Billboard/BillboardsController.cs
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+
+    private IEnumerator UpdateCoroutine()
+    {
+        WaitForEndOfFrame waitForFrameEnd = new WaitForEndOfFrame();
+
+        while (true)
+        {
+            List<Billboard> currentBillboards = new List<Billboard>(billboards);
```
Kinerius, line 70: Do we really need to copy the list every iteration?
If the list is changed while iterating on it we can avoid issues by using an `inverse-for`

Kinerius (changes requested):
Great changes, I left some comments.
Also, tests are failing :D

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Billboard/BillboardsController.cs
```diff
@@ -0,0 +1,92 @@
...
+        if (billboard.EntityTransform == null)
+            return;
+
+        Vector3 lookAtVector = billboard.GetLookAtVector(CameraPosition);
+        if (lookAtVector != Vector3.zero)
+            billboard.EntityTransform.forward = lookAtVector;
+    }
+
+
+    private IEnumerator UpdateCoroutine()
```
mikhail-dcl, line 64: Do we still need to use coroutines or we can convert it to `UniTask`?

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Billboard/BillboardsController.cs
mikhail-dcl: Considering the list is removed from and added to from Awake and OnDestroy, the list is not changed while iterating

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Billboard/Billboard.cs
```diff
@@ -78,25 +60,26 @@ Vector3 GetLookAtVector()
...
-                return;
-            Vector3 lookAtVector = GetLookAtVector();
-            if (lookAtVector != Vector3.zero)
-                entityTransform.forward = lookAtVector;
-        }
+            model = new Model();
+            Tr = transform;
 
-        private void CameraPositionChanged(Vector3 current, Vector3 previous) { ChangeOrientation(); }
+            Environment.i.serviceLocator.Get<IBillboardsController>().BillboardAdded(gameObject);
```
mikhail-dcl, line 77: It creates a circular dependency between `BillboardsController` and `Billboard` (as you also use Billboard from the controller). In my opinion it's a bad design and dependency management violation.
I am not sure how you approach it here.
Is it possible to call `BillboardsController` from the place you create a billboard?

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Billboard/BillboardsController.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        return controller;
+    }
+
+    public void Initialize()
+    {
+        updateCoroutine = CoroutineStarter.Start(UpdateCoroutine());
+    }
+
+    public void BillboardAdded(GameObject billboardContainer)
```
mikhail-dcl, line 28: In the current implementation you call it from `Billboard` itself, thus you can pass the component directly that makes `TryGetComponent` redundant

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Billboard/Billboard.cs
mikhail-dcl: So the proper approach would be via `DataStore`. From `Billboard` you add itself to the DataStore and the controller works with that collection

mikhail-dcl: So the proper approach would be via `DataStore`. From `Billboard` you add itself to the DataStore and the controller works with that collection

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Billboard/BillboardsController.cs
PalauiDCL: The first idea was to calculate on more than one frame (hence not allowing the list to be modified in the frame wait) if there where too many billboards, as the result was visible (meaning when a high number of billboards were present some of them updated weirdly), I removed the feature and now creating a new list is not needed anymore

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Billboard/BillboardsController.cs
PalauiDCL: Exactly the same as in the next answered comment, old removed feature that I had to remove ^^

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Billboard/BillboardsController.cs
PalauiDCL: BillboardsController and Billboard belong to the same assembly, having the interface IBillboardsController knowing about Billboard would create a circular dependency, a solution for this would be to separate BillboardsController and Billboard into different dependencies.

Kinerius: The billboard as a Component shouldn't be passed as a reference, instead, I would create an interface for the Billboard itself, you will have even greater decoupling. 

By doing that you can even create a nice test coverage for BillboardsController without spawning game objects

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Billboard/BillboardsController.cs
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    public void Initialize()
+    {
+        updateCoroutine = CoroutineStarter.Start(UpdateCoroutine());
+    }
+
+    public void BillboardAdded(IBillboard inBillboard)
+    {
+        Billboard billboard = inBillboard as Billboard;
```
Kinerius, line 30: Why cast it back to Billboard? This breaks the Liskov substitution principle, the idea of having an interface is to not deal with the implementation itself. This means that we could have different billboards in the future that use the same interface and we won't have to deal with this implementation of BillboardsController.

For that to work, IBillboard should have the definitions for the functions that we need, in this case:
- You could replace [these lines](https://github.com/decentraland/unity-renderer/blob/ffe59599d71484dfde13a1a85ec365388252f625/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Billboard/BillboardsController.cs#L56-L58) with just a `IBillboard.LookAt(Vector3)`
- [Here](https://github.com/decentraland/unity-renderer/blob/ffe59599d71484dfde13a1a85ec365388252f625/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Billboard/BillboardsController.cs#L76-L81) `Tr`, `EntityTransform` and `LastPosition` can be declared as properties in the interface

Kinerius (changes requested):

kwspgoliasz (approved):
The feature appears to be implemented correctly. No major FPS drops were noticed when testing the billboards.

Kinerius (approved):

PR #3437, opened 2022-11-08
fix: setting default scene load value to 2

dalkia:
## What does this PR change?

Set default scene load value to 2

## How to test the changes?

1) Try a cache-free version. The default value for scene load radius should be 2
2) Open your normal DCL account. Go to the settings section. Press Reset All. Scene load radius should default to 2

## Our Code Review Standards


menduz:
Rationale?

aixaCode:
The main reason is to mitigate poor performance for new users. With the default setting set to 4, the broad experience during the MVMF was around 10fps, most users might not check the settings. Having the settings lower for the start but promoting framerate would better impact the first-time experience.

m3taphysics (approved):

aixaCode (approved):

PR #3439, opened 2022-11-08
fix: Replace non supported characters in chats

sandrade-dcl:
## What does this PR change?
We were receiving huge hiccups when we opened a chat that contained messages with non-supported characteres for the font we use.

image

It seems the process to replace and calculate the size of all these texts is too expensive and it was provoking the hiccups. It happened specially when we had a lot of "corrupted" messages in the same chat.

As solution, before putting a text into the chat, we are identifying all the non-supported characters contained and replacing manually by the `?` character.

## How to test the changes?
1. Go to: https://play.decentraland.zone/?renderer-branch=fix/replace-non-supported-characters-in-chats
2. Open a chat with non-supported characters.
3. Notice all these characters are replaced by the '?' character and you don't have any hiccup.
4. Check the CJK charactes keep working as before.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/DefaultChatEntry.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/DefaultChatEntry.cs
@@ -40,16 +40,14 @@ namespace DCL.Chat.HUD
         public override event Action OnCancelHover;
         public override event Action OnCancelGotoHover;
 
-        public override void Populate(ChatEntryModel chatEntryModel)
-        {
-            chatEntryModel.bodyText = ReplaceNonSupportedCharacteres(body.font, chatEntryModel.bodyText, '?');
+        public override void Populate(ChatEntryModel chatEntryModel) =>
             PopulateTask(chatEntryModel, populationTaskCancellationTokenSource.Token).Forget();
-        }
 
         private async UniTask PopulateTask(ChatEntryModel chatEntryModel, CancellationToken cancellationToken)
         {
             model = chatEntryModel;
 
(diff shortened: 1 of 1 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/DefaultChatEntry.cs
```diff
@@ -40,8 +40,11 @@ public class DefaultChatEntry : ChatEntry, IPointerClickHandler, IPointerEnterHa
         public override event Action OnCancelHover;
         public override event Action OnCancelGotoHover;
 
-        public override void Populate(ChatEntryModel chatEntryModel) =>
+        public override void Populate(ChatEntryModel chatEntryModel)
+        {
+            chatEntryModel.bodyText = ReplaceNonSupportedCharacteres(body.font, chatEntryModel.bodyText, '?');
```
lorux0, line 45: this operation should be done inside `PopulateTask` method, since all the curating process is there, what do you think?

sandrade-dcl: You're right. Done!

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/DefaultChatEntry.cs
```diff
@@ -1,9 +1,9 @@
-using System;
-using System.Threading;
 using Cysharp.Threading.Tasks;
 using DCL.Helpers;
 using DCL.Interface;
 using DCL.SettingsCommon;
+using System;
+using System.Threading;
```
aixaCode, line 6: Is `using System.Threading;` needed?

aixaCode (approved):

lorux0 (approved):

PR #3459, opened 2022-11-09
fix: video url with query params

pbosio:
*Fix playing video which url contains a query param.

situation:
js-hls does not support mp4 videos, mp4 videos are played by native js video element.
checks for using or not js-hls was done with something like `url.EndWith(".mp4")`, which is not the case if the url has a url param.
this PR does some modifications to that file format checking.

example:
`http://some-url.something/video.mp4` works
`http://some-url.something/video.mp4?param=something `was not working

test:
if videos are still playing correctly, then we are good to go


kwspgoliasz (approved):

lorux0 (approved):

PR #3470, opened 2022-11-10
chore: safely handle messagebus exceptions

m3taphysics:
If any type of queue message throws an exception, no more messages will be processed. This will catch the exception and process the next message.


menduz (approved):

menduz:
What does LogException do? is there a way we can have a traceable metric with that? Otherwise we may risk sending errors to /dev/null and never knowing about them

m3taphysics:
@menduz it logs an exception to the console, do we not capture these in rollbar?

menduz:
Kernel was hiding unity logs because of all "behaviour for gameobject not found" kind if errors. Will remove filter soon

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Messaging/MessagingBus.cs
```diff
@@ -193,75 +193,83 @@ public bool ProcessQueue(float timeBudget, out IEnumerator yieldReturn)
...
-                if (shouldLogMessage)
+                    OnMessageProcessed();
+    #if UNITY_EDITOR
+                    if (shouldLogMessage)
+                    {
+                        LogMessage(m, this);
+                    }
+    #endif
+                }
+                catch (Exception e)
```
menduz, line 268: There is an extreme penalty cost on logging and producing stacktraces on this method, it is called thousands of times per second, and if there is an error then it most likely affect the global performance. We should handle this case with a special function that shuts down the scene and tracks only one error in a way that is discoverable using monitoring tools

menduz: For the SDK6 (this code) all should be processed in order. And no message should fail, otherwise we will have an inconsistent state.

We may ask @AjimenezDCL about how to send crash errors to kernel to be tracked

AjimenezDCL: I think Ashley's proposal is the way to go (by now). Otherwise any uncaught exception in any processing of any message will break de routine and stop the messages processing. I also dont think the performance overhead is noticeable, it's not like half of our messages are having exceptions (and would that be the case, catching the exception would be the lesser of the problems).

How is it gonna work with SDK7 and the new protocol @menduz @kuruk-mm ? Because maybe we can leave this as it is with 
 Ashley's patch and move the discussion into how are we gonna handde it in the new system.

To report crash errors to kernel I would suggest a Service that would be use explicitly when catching exceptions to report them to kernel. One of the great advantages is that we could have a different implementation in desktop (for example loging the errors into a local file that can be requested to the users when triaging issues!)

AjimenezDCL (approved):

PR #3487, opened 2022-11-11
fix: Wrong positioning of empty direct messages container

sandrade-dcl:
## What does this PR change?
The position of the image was wrong. The only way to fix it was closing channels.

## How to test the changes?
1. Go to: https://play.decentraland.zone/?renderer-branch=fix/wrong-positioning-of-empty-direct-messages-container
2. Close channels & DMs subsections in the conversation list
3. Open channels
4. Open DMs
5. Notice the position of the emty image is correct

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WorldChatWindowHUD/WorldChatWindowComponentView.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/WorldChatWindowHUD/WorldChatWindowComponentView.cs
@@ -398,6 +398,7 @@ namespace DCL.Chat.HUD
         
             scroll.enabled = !visible;
             isLoadingPrivateChannels = visible;
+            UpdateLayout();
         }
 
         private void UpdateHeaders()


RominaMarchetti:
@sandrade-dcl there is an inconsistent behaviour going on:
When I opened the panel for the first time everything looked good. Once I collapsed and expanded direct messages again to double check it looked wrong. I went to a different chrome tab and when I went back to dcl the image and message were then well placed.
After math: most of the times the message is not well aligned but after a sex it gets fixed automatically, and in a few opportunities it doesn't.

kwssbocian (changes requested):
Issue does not appear to be completely fixed.
✔️Opening up the chat window now displays the empty direct messages container below without having it overlap with other UI elements. Although on the this overlap still occurs for a fraction of a second the first time the window is opened after which the container is moved further down.
❌Overlap still occurs after doing the following:
1. Collapse the Direct Messages section.
2. Collapse the Channels section.
3. Expand the Direct messages section.
chat1
Overlap also occurs if you collapse both sections then expand them back out.
chat2

sandrade-dcl:
> ❌Overlap still occurs

It should be already fixed 👍

kwssbocian (approved):
✔️New fixes have been applied correctly. The overlap that has occurred before is no more.

PR #3488, opened 2022-11-11
fix: tracking of auto-joining channels

lorux0:
## What does this PR change?

Auto-joined channels were not being tracked as `player_joins_channel`.
Now the do with an extra param: `method:auto` or `method:manual`.

## How to test the changes?

1. Go to: https://play.decentraland.zone/?ENABLE_matrix_channels_enabled
2. Login with a wallet that never joined any automatic-channel, for example #mvvm
3. The auto-joined channel should be listed as usual
4. Go to [metabase](https://metabase.decentraland.systems/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7ImRhdGFiYXNlIjozNiwidHlwZSI6InF1ZXJ5IiwicXVlcnkiOnsic291cmNlLXRhYmxlIjo1MTU1fX0sImRpc3BsYXkiOiJ0YWJsZSIsInZpc3VhbGl6YXRpb25fc2V0dGluZ3MiOnt9fQ==) and check that the metric is triggered correctly with the param `method:auto`. It may take a couple of mins before appearing
5. Join any other existing channel from the channel browser
6. Go to [metabase](https://metabase.decentraland.systems/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7ImRhdGFiYXNlIjozNiwidHlwZSI6InF1ZXJ5IiwicXVlcnkiOnsic291cmNlLXRhYmxlIjo1MTU1fX0sImRpc3BsYXkiOiJ0YWJsZSIsInZpc3VhbGl6YXRpb25fc2V0dGluZ3MiOnt9fQ==) and check that the metric exists with the param `method:manual`. It may take a couple of mins before appearing

## Our Code Review Standards


davidejensen (approved):
Yup looks good, only question is, do we want to try and find a way even for the leave stats? Like separating normal leave from auto join channels leave?

lorux0:
> Yup looks good, only question is, do we want to try and find a way even for the leave stats? Like separating normal leave from auto join channels leave?

the requirement is not needed yet, metrics are the same for leaving auto and manual

PR #3489, opened 2022-11-11
fix: ECS7 magenta material

pbosio:
When material model changed, the previous material promise was being forgotten before the new promise was loaded and applied to the renderers, leaving the renderers with a null material == magenta color for a couple of frames.
We are executing a task to forget (on late update) previous material promises after the new material is loaded.


unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Material/MaterialHandler.cs
```diff
@@ -75,10 +90,20 @@ public void OnComponentModelUpdated(IParcelScene scene, IDCLEntity entity, PBMat
                 {
                     material = materialAsset.material
                 });
+                UniTask.RunOnThreadPool(async () =>
```
menduz, line 93: I'd recommend that this kind of pieces of code include a comment, it is really hard to reason what it does and why

Kinerius: Does this work on WebGL?
Anyways, I'd recommend you use the util class `TaskUtils.Run`

Kinerius (changes requested):

Kinerius (approved):

PR #3529, opened 2022-11-17
feat: (ecs7) change billboard approach

leanmendoza:
## What does this PR change?
Change the approach of Billboard in the ECS7. With this, a system is iterating every billboard and updates each frame the game object transform. The previous approach only waited for the camera changes to update the transform and it worked wrong if the entity updated its transform.


unity-renderer/Assets/DCLPlugins/ECS7/Systems/BillboardSystem/ECSBillboardSystem.cs
```diff
@@ -0,0 +1,58 @@
...
+namespace ECSSystems.BillboardSystem
+{
+    public class ECSBillboardSystem : IDisposable
+    {
+        
+        private class State
+        {
+            public ECSComponent<PBBillboard> billboards;
+            public DataStore_Camera camera;
+        }
```
AjimenezDCL, line 16: Why are we wrapping `billboards` and `camera` on a class? it seems like an unnecessary allocation. We could have some readonly fields instead.

AjimenezDCL (approved):
Approving with a small request.

unity-renderer/Assets/DCLPlugins/ECS7/Systems/BillboardSystem/ECSBillboardSystem.cs
```diff
@@ -0,0 +1,58 @@
...
+                billboards = billboards,
+                camera = camera
+            };
+
+        }
+
+        public void Dispose()
+        {
+        }
+
```
pbosio, line 33: If no disposing is needed, we could remove this `IDisposable` implementation

popuz:
hi @leanmendoza, since it will be included in the release, are there some steps to test it (for QA)?

leanmendoza:
> hi @leanmendoza, since it will be included in the release, are there some steps to test it (for QA)?

Sure, although there isn't any `ECS7` scene deployed, you can test here: http://playground.decentraland.org/?&renderer-branch=
1. Open the playground
2. After loading, go to Examples>Billboard
3. Click on Run
you should see how the billboard example works fine.

PR #3535, opened 2022-11-17
fix: ECS7 scene UI renderer over loading screen

pbosio:
*fix ecs7 scene's Ui being renderer over loading screen.
since it's not possible to sort ui order between unity's traditional ui system and uitoolkit we are setting `display:none` every time the loading screen is visible.
also removed `State` class from `ECSScenesUiSystem` and moved it fields directly to `ECSScenesUiSystem` class


AjimenezDCL (approved):

popuz:
hi @pbosio, since it will be included in the release, are there some steps to test it (for QA)?

PR #3539, opened 2022-11-18
spike: sbc shader

AjimenezDCL:
https://play.decentraland.zone/?renderer-branch=feat%2Fsbc-shader&position=62%2C-69&NETWORK=goerli&realm=peer-ue-2.decentraland.zone&ENABLE_SBC_DEBUG=&island=I1t8


unity-renderer/Assets/Rendering/Shaders/URP/SceneBoundaries.hlsl
```diff
@@ -0,0 +1,29 @@
+#ifndef SCENE_BOUNDARIES_FILTER
+#define SCENE_BOUNDARIES_FILTER
+    bool SceneBoundaries_IsInParcel(float3 position, half4 parcel)
+    {
+        const half size = parcel.w;
```
menduz, line 5: please always keep this as a parameter! ;)

Also, does it consider global scenes? to not do the check at all

AjimenezDCL: It's just a proof of concert, I should have stated that haha. The size is encoded in the parcels array

PR #3556, opened 2022-11-21
refactor: UserProfileemotes caused a hiccup

### LINKED ISSUE #3374: Playing emotes causes a hiccup
# The cause

When you play an emote the internal profile gets updated and 10+ systems get their updates triggered.

# The solution

- [ ] Avoid updating the profile.

If this is completely necessary then we have to solve #3375

PalauiDCL:
fixes #3374 

This could also be considered a fix, but, as this was not a bug, but an intended Update that might not be needed anymore, I considered it a refactor.

The call for update of the whole profile when showing an emote was causing a hiccup in performance, as emotes showing does not change the profile, this call has been removed.

## QA

Showing emotes should act exactly the same as in dev, no further modifications should be observable.


kwspgoliasz (approved):
The feature appears to be implemented correctly. No issues with emote functionality were observed during testing.

PalauiDCL:
Meging as I amused the proposition by @Kinerius and QA give the approval too

PR #3557, opened 2022-11-21
fix: TextShape billboard

### LINKED ISSUE #3342: TextShape 'billboard' property is not working anymore
This can be tested at the beginning of the Salmonomicon, the Textshape above Mika's head has the `TextShape.billboard` in `true` and it's not working anymore.

Screen Shot 2022-10-27 at 19 37 06

PalauiDCL:
fixes #3342 

This issue was fixed by adding a Debug.Log... and then removing it.
Found something similar in the past, sometimes Unity seems to create a "cache" on how certain methods work, like a "precompilation status", once any change is done to that code directly, it updates and starts behaving as expected again.

For this issue, I don't know if the solution will just work on my end or on everyone else's to, let's see what QA finds.

In order to create a meaningful change so it can be committed, I removed the brackets of the single lined if.

## QA

Very weird behavior about Unity compilation, it works on my end now, but it could perfectly not work at all on others.

- Go to coordinates (-53, 1) approx.
- In the woods of a set of scenes called **Salmonomicon** you will be able to see a text over the head of a floating head.
- This text should be rotating along the **y** axis looking at the camera.
- If the text does not rotate towards the camera when you move around the floating head, in your end is not working.

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
@@ -65,6 +65,23 @@ namespace DCL.Components
 
         public override string componentName => "text";
 
+
+        private bool CameraFound
+        {
+            get
+            {
+                if (!cameraFound)
+                {
+                    mainCamera = Camera.main;
+                    if (mainCamera != null)
+                        cameraFound = true;
+                }
+
+                return cameraFound;
+            }
+        }
+
+
         private void Awake()
         {
             model = new Model();
@@ -79,14 +96,7 @@ namespace DCL.Components
         private void OnUpdate()
         {
             // Cameras are not detected while loading, so we can not load the camera on Awake or Start
-            if (!cameraFound)
-            {
-                mainCamera = Camera.main;
-                if (mainCamera != null)
-                    cameraFound = true;
-            }
-
-            if (cachedModel.billboard && mainCamera != null)
+            if (cachedModel.billboard && CameraFound)
                 transform.forward = mainCamera.transform.forward;
         }
 
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
@@ -61,26 +61,35 @@ namespace DCL.Components
         private Model cachedModel;
         private Material cachedFontMaterial;
         private Camera mainCamera;
+        private bool cameraFound;
 
         public override string componentName => "text";
 
         private void Awake()
         {
             model = new Model();
-            mainCamera = Camera.main;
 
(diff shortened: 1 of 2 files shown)


davidejensen (approved):
Best fix ever!

kwspgoliasz (changes requested):
Fix failed. The text above Mika's head is fixed in one place regardless of the user's position.

PalauiDCL:
I seem to have detected the root of the problem, teleporting to the position does change the camera and is not detected on Start() or Awake() calls, that's why going there from afar or making a slower call before the camera detection worked prior.
Now the detection is on Update() and I added a boolean variable to avoid checking for null as is a slower check.
@kwspgoliasz I am asking for a recheck if it works on your end, thanks!!

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
```diff
@@ -61,50 +61,57 @@ public class Model : BaseModel
...
+            // Cameras are not detected while loading, so we can not load the camera on Awake or Start
+            if (!cameraFound)
             {
-                transform.forward = mainCamera.transform.forward;
+                mainCamera = Camera.main;
+                if (mainCamera != null)
+                    cameraFound = true;
             }
+
+            if (cachedModel.billboard && mainCamera != null)
```
aixaCode, line 89: You have added here the `cameraFound` but in the line bellow we anyway check for `mainCamera`. Could we combine the optimization?

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
```diff
@@ -61,50 +61,57 @@ public class Model : BaseModel
...
         private void OnUpdate()
         {
-            if (cachedModel.billboard && mainCamera != null)
+            // Cameras are not detected while loading, so we can not load the camera on Awake or Start
+            if (!cameraFound)
             {
-                transform.forward = mainCamera.transform.forward;
+                mainCamera = Camera.main;
+                if (mainCamera != null)
+                    cameraFound = true;
```
aixaCode, line 86: Could we maybe move this logic to a property for readability?

aixaCode (commented):
Few comments

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
PalauiDCL: Agree, I think we can add that, if there is a teleport removing the camera, the billboard is also removed, so it should be no problem at all

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/TextShape/TextShape.cs
PalauiDCL: Will do ^^

PR #3571, opened 2022-11-22
fix: ecs7 input event result hit position

pravusjif:
### WHY
When using input events hit result, the **hit position** and **raycast origin** wasn't returning the correct values if the scene base position wasn't at 0,0.

For example using the [beer dispenser example scene](https://github.com/decentraland-scenes/sdk7-goerli-plaza/tree/main/beer-dispenser), when taking a glass with [E] and placing it on to another table with the same key, the glass is not positioned in the correct place unless using this PR's fix.

### WHAT
* Abstracted input result component event adding into a new method
* Applied raycasthit point and ray origin positions conversion into scene-space before adding the input result component event


dalkia (approved):
Confirmed working! 

https://user-images.githubusercontent.com/1999557/204642983-dbf2299b-c833-4878-b38e-acd4ed1cae1d.mov

Issue #1838, opened 2022-01-26
Avatar in backpack does not have posprocessing effects


olavra:
The camera rendering the avatar of the backpack does not have the same postprocessing profile than in world, skipping bloom, ambient occlussion, color correction etc.

![image.png](https://images.zenhubusercontent.com/603cdb58edb53386a3d0d8d5/e40e7c45-57d0-40aa-9bd4-bf7ddc322b60)
![image.png](https://images.zenhubusercontent.com/603cdb58edb53386a3d0d8d5/e759a900-2dfe-4ddc-8c69-b8e68105edba)

**Acceptance Criteria**
- [ ] Avatar in backpack is rendered using the same effects than in-world

DavideJensen:
I already made some changes in the PR: #1855 the problem is that the post processing stack has some issues with the transparency.
I had to change the render texture color format to R16G16B16A16_SFLOAT because the previous one didn't support the bloom effect, I then enabled the post processing effect and set the proper volume layer in the preview camera.
The issue now is that the solid color alpha is not considered for the transparency anymore, this has been a known issue and there are various post online of people manually changing the shader code to restore the alpha. It also appears that the colors are slightly darker:

![Screenshot 2022-01-28 at 10 47 45](https://user-images.githubusercontent.com/4254522/151526236-d5386817-a048-4509-aac6-29fa3f8de0c7.png)

![Screenshot 2022-01-28 at 10 47 59](https://user-images.githubusercontent.com/4254522/151526127-6e31f9cd-78c8-4c53-942b-2d38231de43b.png)

I'd ask the support of @BrianAmadori or @lorux0 to go on with the issue

PR #3578, opened 2022-11-23
fix: Modify UberPost shader to return the actual alpha instead of 1

mikhail-dcl:
## What does this PR change?

fixes [#1838](https://github.com/decentraland/unity-renderer/issues/1838)

**The problem definition:**
When post processing is enabled `RenderTexture` has no transparency. 
It happens because `UberPost.shader` in URP does not support transparency (as it is used a full screen effect) and returns alpha=1 always.

**The Fix:**
Make a copy of the shader and return the sampled alpha. 
```
half alpha = SAMPLE_TEXTURE2D_X(_SourceTex, sampler_LinearClamp, uvDistorted).w;
return half4(color, alpha);
```

**Take into consideration**
Postprocessing makes appealing darker. IMO it does not look sexy so @MichaelBitsakis probably we will need to add extra post-processing/light for Character Preview so it will look better.
If the current visuals you don't like I can commit the technical fix only leaving post-processing disabled for character preview.

image
image
...

## How to test the changes?

2. ...

## Our Code Review Standards


AjimenezDCL (approved):

davidejensen:
Screenshot 2022-11-24 at 11 49 18
yeah as you said there is a big difference in colours in the preview (example screen you can see the big difference between the face snapshot and the renderer), I think we should ask @olavra how to move next

davidejensen (approved):
colour difference a part, looks good, thanks for the quick fix! :)

olavra:
Face snapshot and backpack render can be affected by the environment lighting of the skybox, so the differences can be high. I say it looks good enough, actually though the backpack avatar look to bright. Please double check emissive textures

mikhail-dcl:
When it comes to emissive it looks much more fidelitous 
Before


After:

kwspgoliasz (approved):
The feature appears to be implemented correctly. No issues were found with the avatar preview.

PR #3612, opened 2022-11-28
feat: move rpc server creation to the service and add ensure rpc

kuruk-mm:
Promise to ensure the RPC is ready

## Our Code Review Standards


AjimenezDCL (approved):
Looks good!

As discussed let's extract an interface for `ClientEmotesKernelService` and use it in `RPC` so it can be mocked (not blocking this PR, but let's do it soon).

PR #3618, opened 2022-11-29
feat: ecs7 plugin always enabled

kuruk-mm:
## Why?

Because we released the ECS7. There will be scenes with ECS7 in the World. So we must have ECS7

## Our Code Review Standards


lorux0:
ecs7 scenes looks OK, but ecs6 scenes are loading empty. This is genesis plaza:
Screen Shot 2022-11-29 at 15 12 50

lorux0 (approved):
looks ok now, i could not reproduce the issue anymore. Looks like my internet service was very unstable

PR #3645, opened 2022-11-30
fix: SDK7 set scene init message on first Crdt

pbosio:
kernel will stop sending `InitMessagesFinished` message for sdk7 scenes.
so instead of sending that message we flag the scene as "initialized" once the first crdt message arrived.

related to https://github.com/decentraland/sdk/issues/474


unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Interfaces/IParcelScene.cs
```diff
@@ -37,5 +37,7 @@ public interface IParcelScene
         void GetWaitingComponentsDebugInfo();
         void SetEntityParent(long entityId, long parentId);
         void RemoveEntity(long id, bool removeImmediatelyFromEntitiesList = true);
+        void SetInitMessagesDone();
```
AjimenezDCL, line 40: Let's call it `MarkInitMessagesDone` or something else, for me `Set` implies that you can enable/disable a flag or similar.

AjimenezDCL: This is not super clean, but as discussed I cannot think of any other way without  refactoring the whole system and ParcelScene and that would be dangerous AF.

AjimenezDCL (approved):

PR #3671, opened 2022-12-02
hotfix: SDK7 Ui breaking scenes loading

pbosio:
copy from this PR already merged to `dev` https://github.com/decentraland/unity-renderer/pull/3637
unloading an sdk7 scene with ui breaks the loading of other scenes due to an exception


pbosio:
dropped for https://github.com/decentraland/unity-renderer/pull/3672

PR #3743, opened 2022-12-12
fix: character moving on characteronly colliders

pravusjif:
### WHY
Character is not recognizing "CharacterOnly" layered colliders as ground. GLTF colliders on SDK7 Scenes use that layer by default, so they are not being recognized as ground or moving platforms.

Issue: https://github.com/decentraland/sdk/issues/502

### WHAT
Added CharacterOnly layer as another ground layer for the character controller.

### TEST
- Locally it can be tested with [moving platforms SDK7 scene](https://github.com/decentraland-scenes/sdk7-goerli-plaza/tree/main/moving-platforms).
- Remotely it can be tested with that same scene deployed (-129,-77): https://play.decentraland.org/?position=-129%2C-77&renderer-branch=fix%2Fchara-moving-on-characteronly-colliders

With this branch's renderer build the moving platforms move the character, with any other build they don't.


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/Resources/Prefabs/CharacterController.prefab
```diff
@@ -373,7 +381,7 @@ MonoBehaviour:
   runningSpeedMultiplier: 0.36
   groundLayers:
     serializedVersion: 2
-    m_Bits: 1025
+    m_Bits: 132097
```
menduz, line 384: Can you add a test to prevent this from accidentally change?

menduz (commented):
I requested a change to prevent accidents! Also canyou track it in the sdk repo?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/Resources/Prefabs/CharacterController.prefab
pravusjif: Sure, I mean it will change if any other layer is added/removed from the ground layers for the character.

Right now in production with SDK6, GLTF colliders are not defaulted to use `CharacterOnly` and use `Default` layer (that's already in the ground layers of the character controller) because of the old problem of builder scenes deployed with an incorrect state of their GLTFComponent collision property.

With the SDK7 correcting that, GLTF colliders get configured correctly and that's why they have the `CharacterOnly` layer by default (and not considered ground in the character controller as it wasn't needed before).

Anyway I'll code a test to confirm the character detects ground on the now 3 layers considered as ground.

pravusjif: done! added test at https://github.com/decentraland/unity-renderer/pull/3743/commits/1a95834569a6fca8d5fb45a9d0b2c44f72c6a17c

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/CharacterController/Tests/CharacterControllerTests.cs
```diff
@@ -384,16 +384,27 @@ void OnMessageFromEngine(string type, string payload)
...
             yield return InitCharacterPosition(50, 2, 0);
             DataStore.i.player.lastTeleportPosition.Set(Vector3.zero);
             UnityEngine.Assertions.Assert.AreApproximatelyEqual(0, Vector3.Distance(DCLCharacterController.i.characterPosition.worldPosition, Vector3.zero), 2.0f);
         }
+
+        [Test]
+        public void CharacterContainsCorrectGroundLayers()
+        {
+            var groundLayerMask = DCLCharacterController.i.groundLayers.value;
+            Assert.IsTrue((groundLayerMask
```
menduz, line 403: It was going to return true all the times otherwise
```suggestion
            Assert.IsTrue((0
```

pravusjif: I tested that implementation yesterday and it was working fine but let me check if there's a test case I may have not taken into account and change it to your way

pravusjif: Yeah the old way only tested if the target layers were in the layermask but not if there were extra layers, [changed it](https://github.com/decentraland/unity-renderer/pull/3743/commits/712df68b71bd1a7c46a21a9c95fd41ea59289bc3) to your sugestion, thanks!

kwspgoliasz (approved):
The feature appears to be implemented correctly. The avatar is moving when positioned on the moving platforms.

lorux0 (approved):

PR #3747, opened 2022-12-13
fix: TriggerSelfUserExpression executing in a background thread

AjimenezDCL:
TriggerSelfUserExpression call is performing some accesses to unity members in a bg thread. 

## Test 

Go to any scene with a self dancing event (like the ongoing event in -123,-126 ). You should dance properly.


kwspgoliasz (approved):

PR #3770, opened 2022-12-14
fix: sdk7 ui fixes

pravusjif:
- Corrected wrapmode default value
- Fixed sort order problem with sdk7 scenes ui and dcl ui
- Applied correct defaults to scene root entity ui element upon creation
- Applied 100% width and height to scene root entity ui element upon creation
- ~Applied absolute position and 100% width and height to ui label (UiText component) entity ui element upon creation, since that component doesn't have its own size properties and shouls use its container element size. Also implemented way of matching its side values (left, right, top, bottom) to its container element padding.~ (will go in another PR, removed fix)
- Fixed root entity invisible ui element blocking pointer
- Changed ScenesUiPanelSettings scale mode to Constant Pixel Size, since that one makes more sense [if we want to keep elements pixel size regardless of viewport size](https://docs.unity.cn/Manual/UIE-Runtime-Panel-Settings.html).

This PR solves the following issues and more:
* https://github.com/decentraland/sdk/issues/491
* https://github.com/decentraland/sdk/issues/492
* https://github.com/decentraland/sdk/issues/538
* ~https://github.com/decentraland/sdk/issues/536~ (will go in another PR, removed fix)

### TESTING
This PR only affects SDK7 Ui entities and components, the improvements can be tested checking each issue referenced above.


unity-renderer/Assets/DCLPlugins/ECS7/ScenesUI/ScenesUiPanelSettings.asset
```diff
@@ -15,14 +15,14 @@ MonoBehaviour:
   themeUss: {fileID: -4733365628477956816, guid: de12f39179342b6408bb91a479f95503,
     type: 3}
   m_TargetTexture: {fileID: 0}
-  m_ScaleMode: 1
+  m_ScaleMode: 0
```
menduz, line 18: Could you include an assertion to check that this value is always 0? to prevent accidentally changing the .asset yaml

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/UIText/Tests/UiTextHandlerShould.cs
```diff
@@ -65,6 +66,69 @@ public void CreateUiContainer()
...
+                Value = "temptation",
+                TextAlign = TextAlignMode.TamMiddleCenter
+            });
+
+            // 2. Check that the UiText style adapted to its padding
+            var style = handler.uiElement.style;
+            Assert.AreEqual(uiTransformModel.PaddingBottom, style.bottom.value.value);
+            Assert.AreEqual(uiTransformModel.PaddingRight, style.right.value.value);
+            Assert.AreEqual(uiTransformModel.PaddingLeft, style.left.value.value);
+            Assert.AreEqual(uiTransformModel.PaddingTop, style.top.value.value);
```
popuz, line 109: following _"one (logical) assertion per test"_ best practice, I would split this test in two. In the first one, I would have only these assertions without the 3. and 4. code in it, while second test will be similar to the current one but without these assertions.

unity-renderer/Assets/DCLPlugins/ECS7/InternalECSComponents/Interfaces/InternalECSComponentModels.cs
```diff
@@ -56,10 +57,11 @@ public class EventData
     public class InternalUiContainer : InternalComponent
     {
         public readonly VisualElement rootElement = new VisualElement();
+        public Action<VisualElement> OnUiElementStyleUpdated;
```
popuz, line 60: having a delegate which is called from outside doesn't feel good for me (because of encapsulation). Would it be possible to call it from inside this class on some of its properties changed (and then we can encapsulate this delegate via `event` keyword)?

popuz (approved):
From the code perspective it looks ok (despite the delegate, that represents event being called from outside). 
However, as for PR description, can you please provide Test Steps or at least small comment that it is "not meant for QA, because ...". It will help QA and also simplify Release test plan creation

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/UIText/UiTextHandler.cs
```diff
@@ -83,25 +105,25 @@ private static TextAnchor ToUnityTextAlign(TextAlignMode align)
                     return TextAnchor.UpperLeft;
                 case TextAlignMode.TamTopRight:
                     return TextAnchor.UpperRight;
-                
+
                 case TextAlignMode.TamBottomCenter:
                     return TextAnchor.LowerCenter;
                 case TextAlignMode.TamBottomLeft:
                     return TextAnchor.LowerLeft;
                 case TextAlignMode.TamBottomRight:
```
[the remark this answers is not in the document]
pravusjif: Thanks for the suggestion! due to the urgency to merge these fixes, we'll tackle those improvements in another PR.

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/UITransform/UITransformHandler.cs
```diff
@@ -255,8 +251,6 @@ private static StyleEnum<UnityEngine.UIElements.Position> GetPosition(YGPosition
         {
             switch (positionType)
             {
-                case YGPositionType.YgptRelative:
-                    return UnityEngine.UIElements.Position.Relative;
                 case YGPositionType.YgptAbsolute:
                     return UnityEngine.UIElements.Position.Absolute;
```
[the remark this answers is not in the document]
pravusjif: Thanks for the suggestion! due to the urgency to merge these fixes, we'll tackle those improvements in another PR.

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/UIText/Tests/UiTextHandlerShould.cs
pravusjif: This fix and test have been removed since it was decided to be tackled with a different approach in another PR

unity-renderer/Assets/DCLPlugins/ECS7/InternalECSComponents/Interfaces/InternalECSComponentModels.cs
pravusjif: This fix and test have been removed since it was decided to be tackled with a different approach in another PR

PR #3772, opened 2022-12-15
Fix: AssetBundles detection commands stop working when triggered again outside the initial range

### LINKED ISSUE #3191: AssetBundles detection commands stop working when triggered again outside the initial range
Description: 
Using the `/detectabs` command then going outside the initial highlight range, onto the unhighlighted tiles, causes all of the `/detectabs` commands to stop working. Issue stops occurring only after having restarted the client.

Steps to reproduce:    
1. Log in with either a guest or a wallet account.   
2. Use the command "/detectabs" to turn on detection of AssetBundles for the nearby loaded tiles.
3. Go beyond the range of the loaded tiles until you encounter non-highlighted tiles and walk into one such tile.
4. When standing on a non-highlighted tile use the "/detectabs" command again.
5. Notice new nearby tiles not being highlighted.
6. Use any combination of "/detectabs" and "/detectabs off".
7. Observe the nearby tiles.

Results:    
All AB detection commands stop working after using "/detectabs" command on an initially non-highlighted tile.

Expected results:    
All AB detection commands continue working after using "/detectabs" command on an initially non-highlighted tile.

Notes:    
- Issue occurs for both production and dev branches.
https://images.zenhubusercontent.com/337227404/18498c79-67e6-4809-ab2b-f2bd3ce2b953/2022_10_07_16_26_47.mp4

PalauiDCL:
fixes #3191 

## Cause
A null reference exception was being fired when triggering again from outside initial range.
The renderers for the initial calculation where already destroyed but we were trying to access their materials, as the method for recalculation came after destroying the previous ones, all the process was blocked by the null ref.

## QA
Steps to reproduce the error in the issue:

- Log in with either a guest or a wallet account.
- Use the command "/detectabs" to turn on detection of AssetBundles for the nearby loaded tiles.
- Go beyond the range of the loaded tiles until you encounter non-highlighted tiles and walk into one such tile.
- When standing on a non-highlighted tile use the "/detectabs" command again.
- Notice new nearby tiles not being highlighted.
- Use any combination of "/detectabs" and "/detectabs off".
- Observe the nearby tiles.

How is supposed to work after this fix:
All AB detection commands continue working after using "/detectabs" command on an initially non-highlighted tile.

### DIFF
--- a/unity-renderer/Assets/DCLPlugins/ABDetectorPlugin/ABDetectorTracker.cs
+++ b/unity-renderer/Assets/DCLPlugins/ABDetectorPlugin/ABDetectorTracker.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using DCL.Controllers;
 using UnityEngine;
@@ -105,7 +105,8 @@ namespace DCL
         {
             foreach (KeyValuePair<Renderer,Material[]> keyValuePair in rendererDict)
             {
-                keyValuePair.Key.materials = keyValuePair.Value;
+                if (keyValuePair.Key != null)
+                    keyValuePair.Key.materials = keyValuePair.Value;
             }
         
             rendererDict.Clear();


kwspgoliasz (changes requested):
The changes were implemented successfully. All of the issues mentioned in the ticket were fixed, however, a new issue with highlighting was discovered. Some parcels and objects in-world are not affected by the asset bundles detector. As a result, the affected parcels are not highlighted in green. The issue seems to occur mostly for empty parcels.

Please refer to the screenshots below for examples of the affected areas:
#3772_Empty

kwspgoliasz (approved):
The issues mentioned in the previous comment are to be addressed in a separate PR. As the main issues with the commands was fixed, the changes are approved.

PR #3804, opened 2022-12-20
chore: update code gen to use interfaces for the clients

kuruk-mm:
From: https://github.com/decentraland/sdk/issues/508

Now the clients are implemented from an interface

## Our Code Review Standards


lorux0:
what is the reason behind adding interfaces for the clients?

kuruk-mm:
> at is the reason behind adding interfaces for the c

It was requested by @AjimenezDCL to have the capacity to mock the clients

lorux0 (approved):

PR #3811, opened 2022-12-20
feat: Decoupling loading screen from kernel

dalkia:
## What does this PR change?

Implements [ADR 101](https://adr.decentraland.org/adr/ADR-101), the decoupling of the loading screen. Behavior should be the same as the current one. Some minor differences should be visible:

1. The only loading message is `Loading scenes, 3D models, and sounds... N% complete`
2. The loading percentage will only show the loading of the destination scene. The change visible here should be a more consistent loading bar.
3. A notification pops-up when there is a loading screen timeout.
4. Solves this [issue](https://app.zenhub.com/workspaces/explorer-6047bad476c3c0001942ee91/issues/decentraland/unity-renderer/1615). No jumpy image should be visible between teleports.

## How to test the changes?

1. Enter DCL through https://play.decentraland.zone/?renderer-branch=feat/decoupling-loading-screen-from-kernel

2. We need to test every teleport method. All of them should take you to the correct destination, with no jumpy image shown before teleporting.

Teleport methods to test are:
Teleport from chat (`/goto coordinates, random, position,magic, crowd`)
Teleport from all the tabs of the Explore panel
Teleport from map
Teleport to any event from the Jump In button present on Genesis Plaza

3. Teleporting to an already loaded scene should not trigger the loading screen. This can be confirmed by teleporting to the current location (Do `/goto 0,0` twice)

4. `/changerealm` command should not trigger the loading screen. Try changing between realms and see that the loading screen is not triggered.

5. Going to a world (and back) should always trigger the loading screen. From (0,0) try going to `/world ile.dcl.eth` and then back to genesis city. The loading screen behaviour should be correct.

6. Same as above, but jump to `/world menduz.dcl.eth` instead of 0,0. Loading screen behaviour should be as expected

7. Test the timeout by going to `/goto 70,70`. There is a scene that always takes more than 30 seconds to load. The loading screen should go away with the notification, but the scene is not going to be loaded

8. Create a new user and check that the signup flow is correctly behaving with the loading screen

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/LoadingScreen/Scripts/LoadingScreenView.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace DCL.LoadingScreen
+{
+    /// <summary>
+    /// Display the current state of the Loading Screen Controller
+    /// </summary>
+    public class LoadingScreenView : MonoBehaviour, ILoadingScreenView
+    {
+        private const string PATH = "_LoadingScreen";
+
+        public static LoadingScreenView Create() =>
```
mikhail-dcl, line 12: I would propose to inherit from `BaseComponentView` so it already has a method `Create`

mikhail-dcl (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/LoadingScreen/Tests/DCL.LoadingScreen.Tests.asmdef
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "DCL.LoadingScreen.Tests",
+    "rootNamespace": "",
+    "references": ["GUID:410d113a616fe7a46bdbf47941da22a4","GUID:97d8897529779cb49bebd400c7f402a6"],
+    "includePlatforms": [],
```
mikhail-dcl, line 5: Please include "Editor" only, otherwise these tests will be for "Play Mode" that is not needed

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/LoadingScreen/Scripts/LoadingScreenView.cs
dalkia: Done!

dalkia: Although, It raised some questions @mikhail-dcl 

1. By using the `Create` method from `BaseComponentView`, the path to the resource its no longer needed in `LoadinScreenView`. But then, Im now adding this path everytime I create the `LoadingScreenView`. Wouldnt this be error prone? Should I make a static variable in `LoadingScreenView` that points to the path?
2. Given that I'm instantiating as  `ILoadingScreenView` inside the controller, to be able to call the `Dispose()` pattern I have to re-define the method inside the interface. This is redundant, since this is already present as a virtual method in `BaseComponentView`.  WDYT of this redundancy? Its something acceptable? I see no really way around it, as I cannot declare `LoadingScreenView` in the controller as a `BaseComponentView` since I will have to cast it everytime I want to call any of its methods.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/LoadingScreen/Tests/DCL.LoadingScreen.Tests.asmdef
dalkia: Done! Lets see how we can add the EditMode tests into the CI. Our current flow will ignore these tests otherwise

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/LoadingScreen/Scripts/LoadingScreenView.cs
mikhail-dcl: 1. I would do the following way in `LoadingScreenView`:
```
public static new LoadingScreenView Create() =>
            base.Create("<path>");
```

mikhail-dcl: 2. Inheriting `ILoadingScreenView` from `IBaseComponentView` will solve the issue?

mikhail-dcl (approved):
I was reviewing the changes in smaller chunks and every concern was addressed. 
It's an excellent job, and a big step towards to decoupling from Kernel and moving some responsibilities from it.

kwssbocian (approved):
Feature implemented correctly.
✔️The loading bar is present for all teleportation methods and moves accordingly to the current loading progress. All available methods of teleportation tested.
✔️Message on the loading screen says `Loading scenes, 3D models, and sounds... N% complete`
✔️A notification pops-up when there is a loading screen timeout.
✔️No jumpy image or skybox without the rest of the world being briefly visible before teleportation.

PR #3816, opened 2022-12-21
Feat: send the user to home after they have been kicked from a world

### LINKED ISSUE #3780: Send the user to home or 0,0 after he has been kicked from a world
When a user gets to a world, the creator's coordinate for that world gets embedded in the url. If the user is kicked out fo the world (due to it being full) the coordinate is already embedded in the url and the user gets teleported to that coordinate. We need to find a way to send him back to home or 0,0

### Acceptance criteria

- [ ] When a user gets kicked out of a world, it teleports to home location

popuz:
## What does this PR change?
Closes  #3780 

- Subscribed `JumpToHomeController` to the `DataStore.i.common.isWorld` change and if it was from World to Non-World, I teleport user to Home

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=feat/go-home-from-world

2. Change realm to the world (`mgoldman.dcl.eth` for example)
3. Then change it back to the non-world (`loki` for example)
4. Observe that you teleported to home (Plaza)

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/DCLPlugins/RealmsPlugin/Utils/JumpToHomeController.cs
+++ b/unity-renderer/Assets/DCLPlugins/RealmsPlugin/Utils/JumpToHomeController.cs
@@ -19,17 +19,39 @@ namespace DCLPlugins.RealmPlugin
         [SerializeField] private RectTransform positionWithoutMiniMap;
 
         private BaseCollection<RealmModel> realms => DataStore.i.realm.realmsInfo;
+        private BaseVariable<bool> isWorld => DataStore.i.common.isWorld;
+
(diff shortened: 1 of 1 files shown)


dalkia (changes requested):
This logic breaks all the other teleports when leaving a world (`/goto` commands (position, random, magic), teleporting from map, teleporting from any tab on the explore menu).

I think we will be able to achieve this catching the kick from kernel and refactoring the url, rather than trying to do it in renderer

kwspgoliasz (approved):
The feature appears to be implemented correctly. The teleportation feature is functioning correctly.

Issue #3494, opened 2022-11-14
Spike: Scene Bounds Checker in shader


AjimenezDCL:
Create a prototype to evaluate performance. SRP batcher will be broken between scenes, but we will gain lot of CPU usage.

#  Acceptance Criteria:
- [x] Have a working build with the new system in place to start comparing performance.
- [ ] Look at potential performance improvements as the current solution, the performance is worse

AjimenezDCL:
So far the new approach degrades performance. Interesting facts I've found while working on this:

- Disabling the current Scene Bounds Checker bring little performance gains. Tested extensively in goerli's hellmap (64,-64) and Genesis plaza.
- Breaking the SRP batcher has little performance regresions in Editor but huge ones in WebGL builds.
- Trying to batch materials per scene doesnt quite work, my leads point towards the rendering order not being fully inforced and materials collision (same CRC) not being correctly handled in the prototype.

PR #3857, opened 2022-12-29
feat: SDK7 scene boundaries checker integration

pravusjif:
https://github.com/decentraland/unity-renderer/issues/3494

- Integrated `GltfContainer`, `MeshRenderer` and `MeshCollider` components to entity `MeshesInfo` to automatically work with current Scene Bounds Checker.


nearnshaw:
@pravusjif   what about NFTShape, is it also possible to track this component with the same mechanics?

pbosio (commented):
Could we discuss about this implementation? I don't think using `entity.meshesInfo` is a good idea

pravusjif:
> Could we discuss about this implementation? I don't think using `entity.meshesInfo` is a good idea

Not using meshesinfo would require a big refactor because MeshesInfo already calculates the merged bounds of meshes and gets the collection of renderers, colliders, etc. but sure let's look for another approach if you like.

pravusjif:
We'll be going with a different approach.

pravusjif:
1. Implement this approach using old hacky/messy SBC
2. Have SBC working in prod for SDK7 scenes
3. Implement new clean SBC for SDK7 scenes removing the hacky stuff.

PR #3868, opened 2022-12-30
feat: sdk7 video player

kuruk-mm:
Basic description of how it works:

You can have a 0-n VideoTexture that refers to a VideoPlayer.

The VideoPlayer plays the video and has the texture of the video.

The VideoTexture by default has no texture (just 1x1 texture resolved by the AssetPromise_EmptyTexture)

There is an internal system called `ECSVideoPlayerSystem` that loops over all the Materials with VideoTextures and assigns the texture of the Video to the Material of that VideoTexture.

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Texture/AssetPromise_EmptyTexture.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+using DCL.Helpers;
+using MainScripts.DCL.Analytics.PerformanceAnalytics;
+using UnityEngine;
+using UnityEngine.Networking;
+using UnityEngine.Video;
+
+namespace DCL
+{
+    public class AssetPromise_EmptyTexture : AssetPromise_Texture
```
AjimenezDCL, line 10: We can use `Texture2D.whiteTexture` and avoid this class.

unity-renderer/Assets/DCLPlugins/ECS7/Systems/VideoPlayerSystem/ECSVideoPlayerSystem.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        public ECSVideoPlayerSystem(
+            IInternalECSComponent<InternalVideoPlayer> videoPlayerComponent,
+            IInternalECSComponent<InternalVideoMaterial> videoMaterialComponent)
+        {
+            this.videoPlayerComponent = videoPlayerComponent;
+            this.videoMaterialComponent = videoMaterialComponent;
+        }
+
+        public void Update()
```
AjimenezDCL, line 20: This approach seems highly inefficient
1. Getting all the playerComponents.
2. Looping through them.
3. Getting all the materials,
4. Loop through them
5. Loop through the textures
6. Get the player for the scene and the videoId
7. Check for all of them IsMaterialAssigned (which relies on LinQ).
8. if material is not assigned, assign it.

has any other alternative have evaluated? mimicking ECS in an environment that is not suitable for ECS (or doesnt bring ECS advantages) seems like having the worst of both worlds.

AjimenezDCL (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/Texture/AssetPromise_EmptyTexture.cs
kuruk-mm: Texture2D.whiteTexture is read-only texture. I tried to use it. But it has problems with the VideoTexture re-assignment.

And this class is to have a Promise of an EmptyTexture to mantain the same system to assign the asset. And without modifying core parts of our system. If you suggest something simplier, it should help a lot 🙏🙌

AjimenezDCL: Oh ofc, I didnt think about the texture needed to be writeable (which is obvious now that I think about it), it's fine, let's keep it that way.

AjimenezDCL: I didnt think about the texture readabiliity...

Yeah, let's keep it as it is. This is the problem with forcing ECS in an environment not suited for it, it's so inflexible we will never find a clean solution for this.

Ideally we could use the dependency of inversion principle and have a resolver gathering the texture but of course that's impossible in the current architecture with the material being a component itself.
Anyway, let's go with this!

unity-renderer/Assets/DCLPlugins/ECS7/Systems/VideoPlayerSystem/Tests/ECSVideoPlayerShould.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            playerComponent.PutFor(scene0, entity00, new InternalVideoPlayer()
+            {
+                videoPlayer = videoPlayer,
+                assignedMaterials = new List<InternalVideoPlayer.MaterialAssigned>(),
+            });
+
+            videoMaterialComponent.PutFor(scene0, entity00, new InternalVideoMaterial()
+            {
+                material = new Material(Shader.Find("Standard")) { mainTexture = new Texture2D(1, 1), },
```
Kinerius, line 76: You should be using `Universal Render Pipeline/Lit` since standard is somewhat deprecated a long time ago.

unity-renderer/Assets/DCLPlugins/ECS7/Systems/VideoPlayerSystem/Tests/ECSVideoPlayerShould.cs
```diff
@@ -0,0 +1,109 @@
...
+            Assert.IsNotNull(videoPlayer.texture);
+
+            systemsUpdate();
+
+            var videoPlayerComponent = playerComponent.GetFor(scene0, entity00);
+            var videoMaterial00 = videoMaterialComponent.GetFor(scene0, entity00);
+            var videoMaterial01 = videoMaterialComponent.GetFor(scene0, entity01);
+            var videoMaterial10 = videoMaterialComponent.GetFor(scene1, entity10);
+
+            Assert.True(ReferenceEquals(videoMaterial00.model.material.GetTexture(ShaderUtils.BaseMap), videoPlayerComponent.model.videoPlayer.texture));
```
Kinerius, line 104: `_BaseMap` property is going to work only if you use `Universal Render Pipeline/Lit`. 
The texture property of `Standard` is `_MainTex`

Kinerius (changes requested):
I found out why the tests are failing

AjimenezDCL (approved):
Unblocking

Kinerius (approved):

PR #3870, opened 2022-12-30
fix: sdk7 hotreload

pbosio:
after hot-reloading a scene renderer wasn't sending messages back to the scene any more due to an exception


Kinerius (approved):
LGTM

PR #3890, opened 2023-01-03
fix: sdk7 set Material's albedo color as type Color4

pbosio:
sdk7 Material's albedo color was of type Color3 so applying alpha value was not possible.
changed type to Color4


protocol-gen/package.json
```diff
@@ -15,7 +15,7 @@
     "typescript": "^4.2.3"
   },
   "dependencies": {
-    "@dcl/protocol": "^1.0.0-3825036757.commit-42e850e",
+    "@dcl/protocol": "https://sdk-team-cdn.decentraland.org/@dcl/protocol/branch//dcl-protocol-1.0.0-3831655577.commit-eb379fa.tgz",
```
pbosio, line 18: TODO: change package when released

sandrade-dcl (approved):

PR #3900, opened 2023-01-04
fix: gltfast wont load assets with relative paths

### LINKED ISSUE #3889: Spike: [GLTFast] Scene 74,47 fails to load

Kinerius:
## What does this PR change?

Fixes #3889

### Problem
Some scenes have `glb` in root folders that has textures on relative paths, for example:
- `root/model.glb`
- `root/textures/male/albedo.png`

Then when loading the `glb` we were loading it like `http://peer.decentraland.org/content/contents/{HASH}` and the relative paths of the textures were formatted like this `http://peer.decentraland.org/content/contents/textures/male/albedo.png`.
When we looked up that file in the content table we were looking for  `albedo.png` instead of `textures/male/albedo.png`

## How to test the changes?

- Go to [74,47](https://play.decentraland.zone/?renderer-branch=fix/gltfast-paths&position=74%2C47) (Etermon)
- Everything should load correctly and no 404 errors should be seen in the console ( ignore those with the URLs with `content-as-asset-bundle` )

### Errors to ignore
image
Why? because this means that some assets are not converted to asset bundles so they fallback to gltf

## Our Code Review Standards


PalauiDCL (approved):
Looking good ^^

PR #3906, opened 2023-01-05
Added fix for GLTF Fast and Parametrized tags for AB Detection tool

### LINKED ISSUE #3783: Issue with parcel highlighting for the Asset Bundles detection
**Description:**
Some parcels and objects are not affected by the Asset Bundles detection, resulting in a lack of highlighting for the affected locations. The issue seems to occur mostly, but not exclusively, for empty parcels.

**Steps to reproduce:**

1. Load into Decentraland as a guest or via a wallet.
2. Find an empty parcel or go to one of the following coordinates: **-45,0** or **-82,-107**.
3. Use the /detectabs command in order to turn on the Asset Bundels detection.
4. Observe the issue.

**Reproduction rate:**
5/5

**Results:**
Some parcels and objects are not affected by the Asset Bundles detection.

**Expected results:**
All parcels and objects are highlighted when the detection is turned on.

**Notes:**

- The issue can be reproduced on both the guest and wallet accounts.

- The issue can be reproduced on both the production and development branches.

- Please see the attached screenshots for reference:

- Platforms: Web client, Windows client.

PalauiDCL:
fixes #3783 

## PR
This task was about investigating why some parcels where not being drawn correctly when using the AB detection tool.
The two issues were:
- Some scenes did not lit up when /detectabs was introduced
- Some scenes were not loaded

## QA
- using /detectabs on the scenes that were not lit previously should now lit them (In color blue) - Test coord (-82,-107)
- The non loaded scenes should see no changes as this tool was discarded as the root issue for that problem, (No changes applied then)

### DIFF
--- a/unity-renderer/Assets/DCLPlugins/ABDetectorPlugin/ABDetectorMaterialsHolder.cs
+++ b/unity-renderer/Assets/DCLPlugins/ABDetectorPlugin/ABDetectorMaterialsHolder.cs
@@ -1,4 +1,4 @@
-﻿using UnityEngine;
+using UnityEngine;
 
 namespace DCL
 {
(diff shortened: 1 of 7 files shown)


kwssbocian (approved):
Issue appears to be fixed.
✔️The previously unlit scenes are now highlighted in blue. 
✔️Non loaded scenes are showing the same as before.

Kinerius (approved):
LGTM

PR #3907, opened 2023-01-05
feat: decoupled loading screen percentage controller

dalkia:
## What does this PR change?

Percentage controller for the decoupled loading screen

## Our Code Review Standards


mikhail-dcl:
Compilation Error on else branch
```
#if UNITY_WEBGL
        fatalErrorReporter = new WebFatalErrorReporter();
#else
        fatalErrorReporter = new DefaultFatalErrorReporter(DataStore.i);
#endif
```

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/LoadingScreen/Scripts/LoadingScreenPercentageController.cs
```diff
@@ -0,0 +1,79 @@
...
+            if (currentScene)
+            {
+                currentScene.sceneLifecycleHandler.sceneResourcesLoadTracker.OnResourcesStatusUpdate -= ResourcesStatusUpdate;
+                currentScene.sceneLifecycleHandler.OnStateRefreshed -= Scene_OnStateRefreshed;
+            }
+        }
+
+        private void SceneController_OnNewSceneAdded(IParcelScene scene)
+        {
+            currentScene = scene as ParcelScene;
```
mikhail-dcl, line 43: Casting violates "Liskov Substitution" principle: we should operate with interfaces instead and do not rely on the underlying implementation

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/LoadingScreen/Scripts/LoadingScreenPercentageController.cs
```diff
@@ -0,0 +1,79 @@
...
+                    currentScene.sceneLifecycleHandler.sceneResourcesLoadTracker.OnResourcesStatusUpdate -= ResourcesStatusUpdate;
+                    currentScene.sceneLifecycleHandler.OnStateRefreshed -= Scene_OnStateRefreshed;
+                    currentScene = null;
+                    break;
+            }
+        }
+
+        private void ResourcesStatusUpdate()
+        {
+            loadingScreenPercentageView.SetLoadingPercentage((int)currentScene.sceneLifecycleHandler.sceneResourcesLoadTracker.loadingProgress);
```
mikhail-dcl, line 68: I wonder what happens if the scene fails to load; should not it be somehow handled?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/LoadingScreen/Scripts/LoadingScreenTipsController.cs
```diff
@@ -63,6 +63,9 @@ private async UniTask IterateTipsAsync()
 
         public void StopTips()
         {
+            if (!tipsView.gameObject.activeSelf) return;
+
+            tipsView.gameObject.SetActive(false);
             disposeCts.Cancel();
```
mikhail-dcl, line 69:
```
disposeCts.Cancel();
disposeCts?.Dispose();
```

Checking for nullity on the second line makes no sense as NRE can occur on the first one.
Also considering how you use it I would recommend checking the token for nullity instead of `activeSelf` so you will get rid of `?` on both lines.

```
if (disposeCts ==  null) return;
```

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/LoadingScreen/Scripts/LoadingScreenPercentageView.cs
```diff
@@ -0,0 +1,23 @@
...
+using UnityEngine.UI;
+
+namespace DCL.LoadingScreen
+{
+    /// <summary>
+    /// View responsible of showing the corresponding loading percentage and message provided by LoadingScreenPercentageController
+    /// </summary>
+    public class LoadingScreenPercentageView : MonoBehaviour
+    {
+
```
mikhail-dcl, line 12: Apply clean-up pls to get rid of redundant black lines.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/LoadingScreen/Scripts/LoadingScreenPercentageController.cs
```diff
@@ -0,0 +1,79 @@
...
+using System;
+using UnityEngine;
+
+namespace DCL.LoadingScreen
+{
+    /// <summary>
+    /// Loading screen percentage updater. The responsibility of this class is to update the loading bar only
+    /// listening to the destination scene
+    /// </summary>
+    public class LoadingScreenPercentageController : IDisposable
```
mikhail-dcl, line 11: I really like this approach with small granular Controllers and Views. 👍 ❤️

mikhail-dcl (changes requested):
Overall looks good, just few concerns to address

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/LoadingScreen/Scripts/LoadingScreenTipsController.cs
dalkia: Makes sense. Done

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/LoadingScreen/Scripts/LoadingScreenPercentageController.cs
dalkia: Found a better event that I could suscribe without casting :)

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/LoadingScreen/Scripts/LoadingScreenPercentageController.cs
dalkia: Using the scene `74,47` for testing a failed scene, I discovered that there is no scene fail handling at all. 

In [kernel](https://github.com/decentraland/kernel/blob/bdbc933ea49af847ef3adac1ec17c30e25b26adf/packages/shared/world/SceneWorker.ts#L339) there is a SCENE_FAIL event called on timeout. But besides reporting that a scene failed loading, it does nothing.

I dont think that scene failing handling should be relative to this PR or even this epic at all. We should create a whole new ticket for SCENE_FAIL handling, since currently no backup system is set up.

WDYT?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/LoadingScreen/Scripts/LoadingScreenTipsView.cs
```diff
@@ -10,8 +10,8 @@ namespace DCL.LoadingScreen
     /// </summary>
     public class LoadingScreenTipsView : MonoBehaviour
     {
-        [SerializeField] private TMP_Text tipsText;
-        [SerializeField] private Image tipsImage;
+        [SerializeField] internal TMP_Text tipsText;
```
mikhail-dcl, line 13: Remember that you could write it like this:
```
[field: SerializeField]
internal TMP_Text tipsText { get; private set; }
```
to keep fields closed

mikhail-dcl (approved):
There are no problems in the code itself but rather in the flow: unless we apply some considerable refactoring to the Kernel flow of events, it will look like a callback hell.

PR #3912, opened 2023-01-05
Added OnFinished Culling and culling restart on CullingController

PalauiDCL:
## PR
Adds simple functionality for Restarting the culling controller and invoking an event when a culling loop finishes

### DIFF
--- a/unity-renderer/Assets/Rendering/Culling/CullingController.cs
+++ b/unity-renderer/Assets/Rendering/Culling/CullingController.cs
@@ -311,7 +311,6 @@ namespace DCL.Rendering
             while (true)
             {
                 bool shouldCheck = objectPositionsDirty || playerPositionDirty;
-
                 playerPositionDirty = false;
                 objectPositionsDirty = false;
 
--- a/unity-renderer/Assets/Rendering/Culling/CullingController.cs
+++ b/unity-renderer/Assets/Rendering/Culling/CullingController.cs
@@ -5,6 +5,7 @@ using UnityEngine;
 using UnityEngine.Rendering;
 using UniversalRenderPipelineAsset = UnityEngine.Rendering.Universal.UniversalRenderPipelineAsset;
 using static DCL.Rendering.CullingControllerUtils;
+using System;
 
 namespace DCL.Rendering
 {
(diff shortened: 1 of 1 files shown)


unity-renderer/Assets/Rendering/Culling/CullingController.cs
```diff
@@ -91,6 +96,17 @@ public void Start()
             StartInternal();
         }
 
+        /// <summary>
+        /// Starts culling update coroutine, if it was already running it restarts.
+        /// The coroutine will keep running until Stop() is called or this class is disposed.
+        /// </summary>
+        public void Restart()
```
lorux0, line 103: this new added method is not being called?

unity-renderer/Assets/Rendering/Culling/CullingController.cs
```diff
@@ -18,30 +19,33 @@ public class CullingController : ICullingController
...
         private float timeBudgetCount = 0;
         private bool resetObjectsNextFrame = false;
         private bool playerPositionDirty;
         private bool objectPositionsDirty;
         private bool running = false;
 
-        // Cache to avoid allocations when getting names
-        private readonly HashSet<Shader> avatarShaders = new HashSet<Shader>();
-        private readonly HashSet<Shader> nonAvatarShaders = new HashSet<Shader>();
+        public event ICullingController.DataReport OnDataReport;
```
lorux0, line 42: these new events are not being used

lorux0 (changes requested):

unity-renderer/Assets/Rendering/Culling/CullingController.cs
PalauiDCL: @lorux0 I decided not to modify them, if you check the changes, you will see that all of them existed already, I just reordered them in order to comply with our code guidelines, so, in this case, there are no changes at all on this code, just reorder.

unity-renderer/Assets/Rendering/Culling/CullingController.cs
PalauiDCL: This will be called from the loading system to stop calculating GameObjects that belong to a different position (And will not be needed anymore), right now is not being called, I am just adding the capability of doing so as requested in the new loading screen

lorux0 (approved):

PR #3921, opened 2023-01-06
fix: gltfast texture re-size for different platforms

Kinerius:
## What does this PR change?

While importing textures with gltfast we were not resizing textures for platform-specific settings. 
This should reduce the memory usage for WebGL

## How to test the changes?

The platform should be more stable and crash less when roaming around the world and teleporting to different places using only GLTFAST and disabling asset bundles


mikhail-dcl (approved):

kwssbocian (approved):
✔️No crashes or issues with stability were observed with glTFast enabled and AssetBundles disabled.
✔️Memory usage was lower overall when compared to production.

PR #3973, opened 2023-01-11
fix: (sdk7) add global inputs and process multiple input in one tick

leanmendoza:
## What does this PR change?

Fix https://github.com/decentraland/sdk/issues/546
Fix https://github.com/decentraland/sdk/issues/550

This PR changes the approach to calculate and emit commands (for pointer events), and to calculate the hover events.

The previous approach saved the last button pressed on each iteration, this made a corner case for button combination in a single iteration (only one button wins). In the current approach, each button has its state saved and the system calculates the change.

For hover events, in addition to the previous problem, the hover was strongly coupled to the SINGLE button behavior. In the current approach, the calculation is a single function of:
1. The current button array state
2. Previous raycast hit
3. Current raycast hit

The last change here is that now global events are emitted (even when there is no raycast), and they are broadcast to all scenes as the previous sdk6 did.

TIP to review: read the [entire file](https://github.com/decentraland/unity-renderer/blob/33e387e0574dd711e5604358f80464d5610530ce/unity-renderer/Assets/DCLPlugins/ECS7/Systems/PointerInputSystem/ECSPointerInputSystem.cs) instead of the git diff (you can read first the previous implementation). 

## How to test the changes?

You can use the BasicInteraction scene from `sdk-test-scenes`:
https://sdk-test-scenes.decentraland.zone/?realm=LocalPreview&position=72%2C-7&renderer-branch=fix/global-and-multiple-button-events&LOS=1
For some reason, the Materials are not working in dev, so instead of coloring the boxes and spheres, they are disappearing.

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/DataStore/DataStore_ECS7.cs
```diff
@@ -32,7 +34,11 @@ public class Hit
         public readonly BaseDictionary<int, BaseRefCountedCollection<object>> pendingSceneResources = new BaseDictionary<int, BaseRefCountedCollection<object>>();
         public readonly BaseDictionary<long, GameObject> shapesReady = new BaseDictionary<long, GameObject>();
         public bool isEcs7Enabled = false;
-        public PointerEvent lastPointerInputEvent = new PointerEvent();
         public RaycastEvent lastPointerRayHit = new RaycastEvent();
+
+        // Should the max buttonId be defined by WebInterface?
+        private const int MAX_BUTTON = (int)WebInterface.ACTION_BUTTON.ACTION_6;
+        public bool[] buttonState = new bool[MAX_BUTTON];
+        public bool[] lastButtonState = new bool[MAX_BUTTON];
```
pbosio, line 42: this should be moved to `ECSPointerInputSystem` state, ideally we should use data store only for stuffs that are needed from outside sdk7 responsibility

unity-renderer/Assets/Scripts/MainScripts/DCL/DataStore/DataStore_ECS7.cs
```diff
@@ -32,7 +34,11 @@ public class Hit
         public readonly BaseDictionary<int, BaseRefCountedCollection<object>> pendingSceneResources = new BaseDictionary<int, BaseRefCountedCollection<object>>();
         public readonly BaseDictionary<long, GameObject> shapesReady = new BaseDictionary<long, GameObject>();
         public bool isEcs7Enabled = false;
-        public PointerEvent lastPointerInputEvent = new PointerEvent();
         public RaycastEvent lastPointerRayHit = new RaycastEvent();
+
+        // Should the max buttonId be defined by WebInterface?
+        private const int MAX_BUTTON = (int)WebInterface.ACTION_BUTTON.ACTION_6;
+        public bool[] buttonState = new bool[MAX_BUTTON];
```
pbosio, line 41:
```suggestion
        public readonly bool[] buttonState = new bool[Enum.GetValues(typeof(WebInterface.ACTION_BUTTON)).Length];
```

unity-renderer/Assets/DCLPlugins/ECS7/Systems/PointerInputSystem/ECSPointerInputSystem.cs
```diff
@@ -56,242 +47,179 @@ private enum InputHitType
...
+
+                state.lastHoverFeedback.hasValue = true;
+                state.lastHoverFeedback.sceneNumber = colliderScene.sceneData.sceneNumber;
+                state.lastHoverFeedback.scene = colliderData.scene;
+                state.lastHoverFeedback.entityId = colliderData.entity.entityId;
+
+                // Nothing to do here, we were not hitting anything, and we still are hitting nothing
+            } else {
+
             }
```
pbosio, line 192:
```suggestion
            } 
```

pbosio (changes requested):
looks good! 
some small changes requested

pbosio (approved):

unity-renderer/Assets/DCLPlugins/ECS7/Systems/PointerInputSystem/ECSPointerInputSystem.cs
```diff
@@ -56,242 +47,178 @@ private enum InputHitType
...
+
+                    // update
+                    prevState[i] = curState[i];
                 }
             }
-            else // no entity hit
+
+            // Check if the hovered entity has changed with three options:
+
+            // 1) We were hitting a collider A and now we're hitting a collider B
```
Kinerius, line 121: Could we get those steps extracted into methods? so we can reduce the visual complexity of this whole Update function.

Kinerius: ### Example
```C#
if (IsColliderDifferent(state))
{
    HandleColliderChanged(state);
} 
if (IsColliderMissing(state))
{ 
    HandleMissingCollider(state);
}
if (IsColliderAvailable(state))
{
    HandleAvailableCollider(state)
} 
```

Kinerius (approved):
Nice refactor!

kwssbocian (approved):
Issues appear to be fixed.
✔️Global and targeted action spheres/boxes respond correctly to the issued commands and disappear upon having the action triggered.
✔️Global inputs are registered regardless of direction the player is looking in.
✔️Multiple simultaneous inputs are correctly being registered.

PR #3976, opened 2023-01-11
hotfix: names & lands repetitive requests when the passport is opened

lorux0:
## What does this PR change?

The profile updating event is only triggered when the profile equality check fails. This fixes:
- many pagination buttons over the names section in the passport
- many GET requests for lands and names while the passport is opened
- passport being reloaded for no reason

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=fix/names-lands-repetitive-requests
2. Go to a crowded place (ideally 10+ users around)
3. Open the http requests tab in the developer console
4. The friends and names request should be triggered only once. Example: `/lambdas/nfts/names/:id` or `/lambdas/nfts/lands`
5. Let the user to change a wearable, name, anything in the profile. A request should be triggered again.
6. Go to the names section, the pagination "dots" should be displayed correctly
7. Open your backpack, start making changes in the wearables, check that you can apply all the changes
8. Open your profile menu, change your name, description, etc. Check that all changes are applied

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/UserProfile/UserProfile.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/UserProfile/UserProfile.cs
@@ -1,12 +1,12 @@
+using DCL;
+using DCL.Helpers;
+using Decentraland.Renderer.KernelServices;
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using DCL;
-using DCL.Helpers;
-using DCL.Interface;
-using Decentraland.Renderer.KernelServices;
+using UnityEditor;
 using UnityEngine;
-using Environment = System.Environment;
+using Environment = DCL.Environment;
 
 [CreateAssetMenu(fileName = "UserProfile", menuName = "UserProfile")]
 public class UserProfile : ScriptableObject //TODO Move to base variable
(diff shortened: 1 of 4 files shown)


sandrade-dcl (approved):
Although the source of this problem is in the kernel side and it should be fixed there, it's also a good practice in this case to ensure not to trigger the user profile's `OnUpdate` event from the Unity side if the information received is exactly the same.

I have tested it and it fix several known issues, I think it worth it to release it as a hotfix.

Good work!

unity-renderer/Assets/Scripts/MainScripts/DCL/UserProfile/UserProfile.cs
```diff
@@ -55,9 +55,11 @@ public void UpdateData(UserProfileModel newModel)
     {
         if (newModel == null)
         {
-            model = null;
+            model = new UserProfileModel();
```
eordano, line 58: This line feels really wrong. Could you explain what is happening here? Why are we creating a new model? Isn't line 47 in charge of this? Also, shouldn't we check if `model` is not-null before assigning a new model to it?

lorux0: If the current model is null, like it was in the old code, it will throw exceptions when trying to update it again. Im instantiating it again so its a valid object with default values, so the flow doesnt break. Line 47 is just an initialization, but it was never getting instantiated again after nulling the model.

kwssbocian (approved):
Issue appears to be fixed.
✔️Inspecting another user's passport no longer results in multiple HTTP requests being made for the same feature. Now for every new user 4 requests are issues: for lands, names, wearables and emotes. Upon repeated inspection of the passport for a user that has not made any changes to their profile only 2 requests: for wearables and emotes are being made.

PR #4006, opened 2023-01-12
fix: Clicking on BUY on NFTs in the Profile does not open the right link

### LINKED ISSUE #3991: Clicking on BUY on NFTs in the Profile does not open the right link
When I click on a NFT inside the Profile, and click on the BUY button it sometimes takes me to the Marketplace front page instead of the specific item.

There is no clear steps to reproduce this error.

sandrade-dcl:
## What does this PR change?
Fixes #3991

When we clicked on the BUY button of some NFTs inside the Profile, it sometimes takes us to the Marketplace front page instead of the specific item.

### Problem
- In order to get the necessary information to build the marketplace's url for an NFT (basically we need the collectionId + the tokenId of the NFT), when the passport opens, we send a query to TheGraph (to both networks ETHEREUM and POLYGON) to obtain the list of owned NFTs belonging to the user that we're loading in the passport. As result we always have these 2 pre-loaded lists with the TheGraph information: `ownedNftCollectionsL1`  and `ownedNftCollectionsL2`.
- When we click on the BUY button of any NFT item loaded in the passport, we try to find that NFT in the pre-loaded lists mentioned above. If we find it, we get the the collectionId + the tokenId from there and we build the url to redirect to the correct page in the marketplace.
- The problem is that these pre-loaded lists only can have a maximum of 100 entries because the queries to TheGraph have this maximum as default, so we can find cases where an user can own more than 100 NFTs and if we click on the BUY button of one of them and it is not found in none of the pre-loaded lists, we cannot build the correct url because we don't have the collectionId + the tokenId. In these cases we are redirectioning to the landing page of the marketplace.

### Solution
In the cases where we click on the BUY button of a NFT whose information we do not have in the pre-loaded TheGraph lists, we will send another query to TheGraph to obtain the information of that specific NFT and to be able to build the correct marketplace's url.

### Important
In the future, when we use the new lambdas to obtain the wearables/emotes info (as we are already doing with the names/lands information), we will be able to obtain the collectionId + the tokenId directly in our `WearableItem` model, in the same request, and we won't have to use a paralel request to TheGraph to obtain extra information of the NFTs.

## How to test the changes?
1. Go to: https://play.decentraland.zone/?renderer-branch=fix/buy-in-profile-does-not-open-the-right-link
2. Open the passport for several users.
3. Check that the BUY buttons of all wearable and emote items redirect to the correct marketplace url.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Passport/Passport/WebInterfacePassportApiBridge.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Passport/Passport/WebInterfacePassportApiBridge.cs
@@ -12,7 +12,7 @@ namespace DCl.Social.Passports
 
         public async UniTask<List<Nft>> QueryNftCollectionsEthereum(string userId)
         {
-            List<Nft> nftList = new List<Nft>();
+            List<Nft> nftList = null;
             await DCL.Environment.i.platform.serviceProviders.theGraph.QueryNftCollections(userId, NftCollectionsLayer.ETHEREUM)
                      .Then((nfts) => nftList = nfts);
 
@@ -21,16 +21,16 @@ namespace DCl.Social.Passports
 
         public async UniTask<Nft> QueryNftCollectionEthereum(string userId, string urn)
         {
-            List<Nft> nftList = new List<Nft>();
+            List<Nft> nftList = null;
             await DCL.Environment.i.platform.serviceProviders.theGraph.QueryNftCollectionsByUrn(userId, urn, NftCollectionsLayer.ETHEREUM)
                      .Then((nfts) => nftList = nfts);
 
-            return nftList.Count > 0 ? nftList[0] : null;
+            return nftList?.Count > 0 ? nftList[0] : null;
         }
 
         public async UniTask<List<Nft>> QueryNftCollectionsMatic(string userId)
         {
-            List<Nft> nftList = new List<Nft>();
+            List<Nft> nftList = null;
             await DCL.Environment.i.platform.serviceProviders.theGraph.QueryNftCollections(userId, NftCollectionsLayer.MATIC)
                      .Then((nfts) => nftList = nfts);
 
@@ -39,11 +39,11 @@ namespace DCl.Social.Passports
 
         public async UniTask<Nft> QueryNftCollectionMatic(string userId, string urn)
         {
-            List<Nft> nftList = new List<Nft>();
+            List<Nft> nftList = null;
             await DCL.Environment.i.platform.serviceProviders.theGraph.QueryNftCollectionsByUrn(userId, urn, NftCollectionsLayer.MATIC)
                      .Then((nfts) => nftList = nfts);
 
-            return nftList.Count > 0 ? nftList[0] : null;
+            return nftList?.Count > 0 ? nftList[0] : null;
         }
     }
 }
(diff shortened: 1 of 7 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Passport/Passport/WebInterfacePassportApiBridge.cs
```diff
@@ -21,6 +19,15 @@ await DCL.Environment.i.platform.serviceProviders.theGraph.QueryNftCollections(u
             return nftList;
         }
 
+        public async UniTask<Nft> QueryNftCollectionEthereum(string userId, string urn)
+        {
+            List<Nft> nftList = new List<Nft>();
```
AjimenezDCL, line 24: This is allocating memory that it's not used, it's gonna be assigned in the `.Then` anyway.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Passport/Passport/WebInterfacePassportApiBridge.cs
```diff
@@ -21,6 +19,15 @@ await DCL.Environment.i.platform.serviceProviders.theGraph.QueryNftCollections(u
             return nftList;
         }
 
+        public async UniTask<Nft> QueryNftCollectionEthereum(string userId, string urn)
```
AjimenezDCL, line 22: This needs to support cancellation

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Passport/Passport/WebInterfacePassportApiBridge.cs
```diff
@@ -29,5 +36,14 @@ await DCL.Environment.i.platform.serviceProviders.theGraph.QueryNftCollections(u
 
             return nftList;
         }
+
+        public async UniTask<Nft> QueryNftCollectionMatic(string userId, string urn)
```
AjimenezDCL, line 40: also needs cancellation suppot.

kwssbocian (approved):
Fix implemented correctly.
✔️Tested different NFT wearables for various users and in all cases clicking the "Buy" button on the wearable inside the passport resulted in the correct page of Marketplace being opened.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Passport/Passport/WebInterfacePassportApiBridge.cs
sandrade-dcl: Done in https://github.com/decentraland/unity-renderer/pull/4006/commits/b2664bebab0dbfe34c26f10fa95feba65e77b41b. Thanks!!

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Passport/Passport/WebInterfacePassportApiBridge.cs
sandrade-dcl: Done in https://github.com/decentraland/unity-renderer/pull/4006/commits/b2664bebab0dbfe34c26f10fa95feba65e77b41b. Thanks!!

AjimenezDCL (approved):
Good job! is looking good!

PR #4032, opened 2023-01-16
fix: settings should not be cleared when user logs out

### LINKED ISSUE #3999: Settings should not be cleared when user logs out
Settings should be local to the machine the user is using, and independent of the avatar.

Acceptance Criteria Journey:

1. Load into the platform with a wallet.
2. Open the Settings menu ('P').
3. Change any of the settings.
4. Log out using the "Logout" button on the avatar's menu (top right)
5. Log into the platform again with the same wallet
6. Access the Settings menu
7. See the same options

Alternative journeys to test: Wallet, then Guest; Guest, then Wallet; and Guest, logout, and Guest again.

popuz:
## What does this PR change?
Closes #3999 

### Problem 1
Settings were saved only when closing the UI. So it wasn't saved in case if user pressed sign-in (Guest) or log-out (Wallet) button before closing settings UI.

**Solution:**
- Since I didn't found any exit-point for the WebGL application to put the saving logic there, I directly subscribed to the SingIn and LogOut events to solve just this particular cases - it is inside `ProfileHUDController.cs` constructor and `Dispose` methods.
- I also added `Settings.Save(`) to `Main.ApplicationWantsToQuit()`, method, but it will work only in Standalone client.

### Problem 2
 `HideUI`, `ShowAvatarNames` and `CameraControl` userSettings were not saved in any scenario.

**Solution:** added respective logic in `PlayerPrefsGeneralSettingsRepository` and adjusted `HideUIControlController`

### WALLET scenario
1. Go to: https://play.decentraland.zone/?renderer-branch=fix/save-user-setttings-on-logout
2. Load into the platform with a wallet.
3. Open the Settings menu ('P').
4. Change any of the settings.
5. Log out using the "Logout" button on the avatar's menu (top right), before closing the UI
6. Log into the platform again with the same wallet
7. Access the Settings menu
8. See settings being saved

### GUEST scenario
same as previous, but in step 5 press "Sing-in" button instead.

### Alternative journeys to test: 
- Wallet, then Guest; 
- Guest, then Wallet; 
Note: settings are shared between Wallets and Guests on the same PC.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Settings/PlayerPrefsGeneralSettingsRepository.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Settings/PlayerPrefsGeneralSettingsRepository.cs
@@ -17,9 +17,7 @@ namespace DCL.SettingsCommon
         public const string INVERT_Y_AXIS = "InvertYAxis";
         public const string SKYBOX_TIME = "skyboxTime";
         public const string FIRST_PERSON_CAMERA_FOV = "firstPersonCameraFOV";
-
         public const string SHOW_AVATAR_NAMES = "showAvatarNames";
-        public const string HIDE_UI = "hideUI";
         public const string CAMERA_MOVEMENT_MODE = "cameraMovementMode";
 
         private readonly IPlayerPrefsSettingsByKey settingsByKey;
@@ -61,7 +59,6 @@ namespace DCL.SettingsCommon
             settingsByKey.SetBool(INVERT_Y_AXIS, currentSettings.invertYAxis);
             settingsByKey.SetFloat(SKYBOX_TIME, currentSettings.skyboxTime);
             settingsByKey.SetFloat(FIRST_PERSON_CAMERA_FOV, currentSettings.firstPersonCameraFOV);
-            settingsByKey.SetBool(HIDE_UI, currentSettings.hideUI);
             settingsByKey.SetBool(SHOW_AVATAR_NAMES, currentSettings.showAvatarNames);
             settingsByKey.SetBool(CAMERA_MOVEMENT_MODE, currentSettings.leftMouseButtonCursorLock);
         }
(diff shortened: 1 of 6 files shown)


kwsskulinski (approved):
Issue is fixed. Settings are not being erased after logout or changing to guest/wallet.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Settings/PlayerPrefsGeneralSettingsRepository.cs
```diff
@@ -59,6 +61,9 @@ public void Save()
             settingsByKey.SetBool(INVERT_Y_AXIS, currentSettings.invertYAxis);
             settingsByKey.SetFloat(SKYBOX_TIME, currentSettings.skyboxTime);
             settingsByKey.SetFloat(FIRST_PERSON_CAMERA_FOV, currentSettings.firstPersonCameraFOV);
+            settingsByKey.SetBool(HIDE_UI, currentSettings.hideUI);
+            settingsByKey.SetBool(SHOW_AVATAR_NAMES, currentSettings.showAvatarNames);
+            settingsByKey.SetBool(CAMERA_MOVEMENT_MODE, currentSettings.leftMouseButtonCursorLock);
```
sandrade-dcl, line 66: **Question**: Have you confirmed that it is a requirement to store these 3 settings (`HIDE_UI`, `SHOW_AVATAR_NAMES` and `CAMERA_MOVEMENT_MODE`)? It sounds to me like we were not saving some of them on purpose. For instance, for me it's a normal behavior that the "HIDE UI" feature always initializes deactivated as default at the beginning.

Before applying the change, I would ask product to be sure.

sandrade-dcl (approved):
The code is fine! I'm approving but please, take a look to the comment I left before applying the changes. Thanks!

AjimenezDCL (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/Settings/PlayerPrefsGeneralSettingsRepository.cs
popuz: Good point, @sandrade-dcl. `CAMERA_MOVEMENT_MODE` is new feature, so it is supposed to be saved. I We also agreed with @eordano, that `SHOW_AVATAR_NAMES ` should be saved as well (imagine someone who don't like it - he will need to go to settings on every enter to disable it). Thus, only `HIDE_UI` will not be saved. I will push fix soon.

PR #4033, opened 2023-01-16
fix: chat not opening from notification when the friend was not previously loaded

lorux0:
## What does this PR change?

Fixes clicking on chat notifications not opening the chat window with that user.

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=fix/open-dm-from-noti
2. Take two friend users, both in different locations in the world, far away from each other
3. Do not open any friends window with user A
4. Write a private message from user B to user A
5. User A should receive a DM notification
6. Click the notification
7. The chat window should be opened as expected

## Our Code Review Standards


kwsskulinski (approved):
Issue is fixed. As described clicking on the notification opens the private chat.

davidejensen (approved):

PR #4035, opened 2023-01-16
chore: build fails if some file exceeds maximum size

menduz:
## What does this PR change?

Makes the build process fail if the generated assets are bigger than 42MB. We can observe it fails with:

```
The file /tmp/workspace/unity-renderer/browser-interface/dist/unity.data exceeds the maximum cacheable file size: 69.67MB
The file /tmp/workspace/unity-renderer/browser-interface/dist/unity.wasm exceeds the maximum cacheable file size: 79.09MB
```

After the file size is reduced, the main branch should be merged to this PR to make it pass. It will be ready to merge then.


kuruk-mm (changes requested):
We should check after we compress with brotli 🤔 

I don't who is compressing it with brotli

PR #4042, opened 2023-01-17
fix: locked mouse on scene timeout

dalkia:
## What does this PR change?

Kernel was not using the DeactivateRendering method at all. 

With the new decoupled loading screen, the usage was reintroduced, but involved unexpected behaviour, like the locked mouse on scene timeout.

This PR stops using the DeactivateRendering method, so as to copy the current behaviour. Its not ideal, but it should be corrected while we take out responsibilities from kernel; as well as rethinking how the RenderingController class is used.

## Our Code Review Standards


dalkia:
Just an experiment, didnt solve the issue

PR #4057, opened 2023-01-18
Fix: Green blockers should not be visible unless the scene is being loaded

0xD-Fabio:
## What does this PR change?

This PR fix 2 related issues:
1. the problem was green blockers appearing on edge parcels even when no content was being loaded. Now the green blockers appear on any empty or non-empty parcel only while the content is still being loaded.
2. the problem where the user could select from the settings panel a load radius of 1. This problem caged the user in a 1 by 1 square parcel, without any possibility to escape. Now the minimum radius is 2.
...

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=fix/not-loading-green-blockers
2. Set the loading radius in the settings to 2, to easily see green-blockers when exploring.
3. Explore parcels outside Genesis plaza to check if green blockers spawn correctly
4. Teleport to a remote area where there are a lot of empty scenes, and check if green-blockers appear only if there is something loading.
5. Try changing the radius from the setting, and see if you can select 1 anymore.

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ParcelScene.cs
```diff
@@ -336,7 +336,18 @@ private void OnDrawGizmosSelected()
             }
         }
 
-        public IDCLEntity GetEntityById(string entityId) { throw new System.NotImplementedException(); }
+        public IDCLEntity GetEntityById(string entityId)
```
AjimenezDCL, line 339: we are deprecating the string usage for entities and scenes. Can you give it a look where this `GetEntityById` method is called and try to use the long instead? It's probably not much more difficult since before it wasnt even implemented.

AjimenezDCL (approved):
Approving, but take a look at my comment to see if we can remove the string usage.

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ParcelScene.cs
0xD-Fabio: Thanks, fixed!

dalkia (approved):
Tested in world environment, works as expected!

kwssbocian (approved):
Fixes implemented correctly.
✔️Setting scenes load radius to 2 and walking around various scenes in the world resulted in the green wall blockers being shown only when a nearby scene in range is still loading and not for unloaded scenes outside of range.
✔️Parcels outside Genesis Plaza appeared correctly.
✔️Minimum allowed scenes load radius setting has been increased from 1 to 2.
Additionally no changes have been observed regarding behavior of green blockers in Worlds where they've been configured to appear for example 2 tiles away from spawn point.

PR #4071, opened 2023-01-19
fix: avatar shape with null user id failing to load

Kinerius:
## What does this PR change?

Fixed a few null reference exceptions caused by the userId being null when loading avatar shapes.

Also fixed the nameplate background being visible when the player name is null. 

## How to test the changes?

https://play.decentraland.zone/?renderer-branch=fix/avatar-shape-name&realm=redeemer.dcl.eth

The avatar in the mirror should be visible and the nameplate of those avatars should be invisible.

## Our Code Review Standards


kwsskulinski (approved):
Issue is fixed. There is no nameplates and characters load without problems.

mikhail-dcl (approved):

PR #4084, opened 2023-01-19
chore: Refactor Wearables Catalog (Part 1 of 2)

### LINKED ISSUE #4014: Refactor Wearables Catalog (Part 1)
- `CatalogController` is an extreme mess and does not follow any good practices
- We are going to switch from Kernel-driven lambdas to Renderer once, we already have a generic system for that and two implementations. Thus we can call it our new standard. 
- It's much easier to design `Wearables Catalog` from scratch than trying to refactor the current one.
- This task is not for UI changes, thus they should be minimized

**Acceptance Criteria**
- [x] Design a new `Wearables Catalog Service`:
   - [x] Implement 2 new services:
      - [x] `LambdasWearablesCatalogService`
      - [x] `WebInterfaceWearablesCatalogService`
   - [x] Implement a new `WearablesCatalogService` service that will choose to use the lambdas services or the web interface one depending on the received configuration flag from kernel.
   - [x] All API that supports pagination should be exposed as `LambdaResponsePagePointer`
   - [x] All API that Is currently not migrated to `lamd2` (the same `Kernel` consumes) should be exposed as `async UniTask` requests
   - [x] There should be no `static` API provided by `Wearables Catalog Service`.
   - [x] `wearableCatalog` should be still filled to remain compatibility
   - [x] Think of a cache of `LambdaResponsePagePointer`s on the level of `Service` to maintain the current logic of avoiding repeated requests
- [x] Introduce backward compatibility:
   - [x] Convert `public static Promise<WearableItem[]> RequestOwnedWearables(string userId)` to an extension to retrieve all pages at once and expose as `UniTask<WearableItem[]>`
   - [x] If API for third-party collections will be ready to the moment, convert `CatalogController.RequestThirdPartyWearablesByCollection` in a similar fashion as well

sandrade-dcl:
## What does this PR include?
Fix #4014 

Deprecate the current `CatalogController` class to re-do it from scratch.

### Implementation's details
- **Implementation of `WearablesCatalogService` (`IWearablesCatalogService`)**: It will choose to use the lambdas services or the web interface one depending on the `urlParamsForWearablesDebug` flag received in the kernel's configuration:
    - **Implementation of `LambdasWearablesCatalogService` (`IWearablesCatalogService`)**: This service implements a direct way of getting wearables sending the requests directly to lambdas.
    - **Implementation of `WebInterfaceWearablesCatalogService` (`IWearablesCatalogService`)**: This service keeps the same logic of the old CatalogController (but managed with UniTasks instead of Promises) so all the wearables requests will pass through kernel. It will be deprecated once we move all the kernel's logic related to requesting wearables to Unity.
- **Pagination support for owned wearables**: The `RequestOwnedWearablesAsync` method implemented in `LambdasWearablesCatalogService` is prepared to support pagination through the new lambdas.
- **All requests implemented as `UniTasks`**: All requests in both `LambdasWearablesCatalogService` and `WebInterfaceWearablesCatalogService` are implemented as async `UniTasks` and supporting `CancellationTokens`.
- **Backward compatibility**: The new service has been implemented to be completely compatible with the current wearables logic we have in the old `CatalogController`.

### Next step of this refactor (Part 2 of 2) that will be implemented in #4132
- **Change all the static invocations used with `CatalogController` by calls to the new WearablesCatalogService**.
- **Make everything related to wearables (Backpack, Passport, etc) work as before this refactor**.
- **Tests implementation**.

## How to test it?
This first part of the refactor is not still testeable by QA, all the changes are related to the creation of the new **Wearables Catalog** service but they are not yet being implemented to replace the current behavior. The intention of this PR is to review and approve the new architecture.

## Our Code Review Standards


mikhail-dcl (changes requested):
I processed `LambdasWearablesCatalogService` and `Proxy` only so far, I have 28 comments

unity-renderer/Assets/DCLServices/WearablesCatalogService/WearablesWebInterfaceBridge.cs
```diff
@@ -0,0 +1,13 @@
+﻿using DCL.Interface;
+
+namespace DCLServices.WearablesCatalogService
+{
+    public class WearablesWebInterfaceBridge
+    {
+        public virtual void RequestWearables(string ownedByUser, string[] wearableIds, string[] collectionIds, string context) =>
```
mikhail-dcl, line 7: Why these methods are virtual?

unity-renderer/Assets/DCLServices/Lambdas/LambdasService.cs
```diff
@@ -90,7 +89,7 @@ internal string GetUrl(string endPoint, params (string paramName, string paramVa
 
             if (urlEncodedParams.Length > 0)
             {
-                urlBuilder.Append('?');
+                urlBuilder.Append(urlBuilder.ToString().Contains('?') ? '&' : '?');
```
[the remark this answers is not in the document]
sandrade-dcl: Done in: https://github.com/decentraland/unity-renderer/pull/4084/commits/0e9bdad0bf1e34457084996299476fbdc9fd59ed

unity-renderer/Assets/DCLServices/WearablesCatalogService/IWearablesCatalogService.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Collections.Generic;
+using System.Threading;
+
+namespace DCLServices.WearablesCatalogService
+{
+    public interface IWearablesCatalogService : IService
+    {
+        BaseDictionary<string, WearableItem> WearablesCatalog { get; }
+
+        UniTask<WearableItem[]> RequestOwnedWearablesAsync(string userId, int pageNumber, int pageSize, CancellationToken ct);
```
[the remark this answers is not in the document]
sandrade-dcl: Done in: https://github.com/decentraland/unity-renderer/pull/4084/commits/836bbf4c85228700ccb3643d64dbb53624d30087

unity-renderer/Assets/DCLServices/WearablesCatalogService/WearablesCatalogService.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        public void Initialize()
+        {
+            kernelConfig.EnsureConfigInitialized().Then(config => OnKernelConfigChanged(config, null));
+            kernelConfig.OnChange += OnKernelConfigChanged;
+        }
+
+        public void Dispose()
+        {
+            if (kernelConfig != null)
```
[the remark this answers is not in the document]
sandrade-dcl: Done in: https://github.com/decentraland/unity-renderer/pull/4084/commits/42ca04255b73afe011b91769b081f294a30f1dac

unity-renderer/Assets/DCLServices/WearablesCatalogService/WearablesCatalogService.cs
```diff
@@ -0,0 +1,97 @@
...
+                kernelConfig.OnChange -= OnKernelConfigChanged;
+
+            wearablesCatalogServiceInUse?.Dispose();
+        }
+
+        public UniTask<WearableItem[]> RequestOwnedWearablesAsync(string userId, int pageNumber, int pageSize, CancellationToken ct) =>
+            wearablesCatalogServiceInUse.RequestOwnedWearablesAsync(userId, pageNumber, pageSize, ct);
```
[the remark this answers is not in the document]
sandrade-dcl: Done in: https://github.com/decentraland/unity-renderer/pull/4084/commits/7a5d7d1e685fa758ec7d23bd3b25c89cc72f7876

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,285 @@
...
+            {
+                ownerWearablesPagePointers[userId] = new (
+                    WEARABLES_BY_OWNER_END_POINT.Replace("{userId}", userId),
+                    pageSize, ct, this);
+            }
+
+            var pageResponse = await ownerWearablesPagePointers[userId].GetPageAsync(pageNumber, ct);
+            AddWearablesToCatalog(pageResponse.response.wearables.ToArray());
+
+            return pageResponse.response.wearables.ToArray();
```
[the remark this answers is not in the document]
sandrade-dcl: Done in: https://github.com/decentraland/unity-renderer/pull/4084/commits/4fd3bac3e0b0aef0ec7d4831f1bbb49c16496796

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,285 @@
...
+        public void Dispose()
+        {
+            serviceCts.SafeCancelAndDispose();
+            serviceCts = null;
+            Clear();
+        }
+
+        public async UniTask<WearableItem[]> RequestOwnedWearablesAsync(string userId, int pageNumber, int pageSize, CancellationToken ct)
+        {
+            if (!ownerWearablesPagePointers.ContainsKey(userId))
```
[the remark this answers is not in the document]
sandrade-dcl: Done in: https://github.com/decentraland/unity-renderer/pull/4084/commits/4fd3bac3e0b0aef0ec7d4831f1bbb49c16496796

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,285 @@
...
+            {
+                ownerWearablesPagePointers[userId] = new (
+                    WEARABLES_BY_OWNER_END_POINT.Replace("{userId}", userId),
+                    pageSize, ct, this);
+            }
+
+            var pageResponse = await ownerWearablesPagePointers[userId].GetPageAsync(pageNumber, ct);
+            AddWearablesToCatalog(pageResponse.response.wearables.ToArray());
```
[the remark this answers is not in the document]
sandrade-dcl: Done in: https://github.com/decentraland/unity-renderer/pull/4084/commits/4fd3bac3e0b0aef0ec7d4831f1bbb49c16496796

unity-renderer/Assets/DCLServices/WearablesCatalogService/IWearablesCatalogService.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        UniTask<IReadOnlyList<WearableItem>> RequestOwnedWearablesAsync(string userId, int pageNumber, int pageSize, CancellationToken ct);
+        UniTask<IReadOnlyList<WearableItem>> RequestBaseWearablesAsync(CancellationToken ct);
+        UniTask<IReadOnlyList<WearableItem>> RequestThirdPartyWearablesByCollectionAsync(string userId, string collectionId, CancellationToken ct);
+        UniTask<IReadOnlyList<WearableItem>> RequestWearablesAsync(string[] wearableIds, CancellationToken ct);
+        UniTask<WearableItem> RequestWearableAsync(string wearableId, CancellationToken ct);
+        void AddWearablesToCatalog(IReadOnlyList<WearableItem> wearableItems);
+        void RemoveWearablesFromCatalog(IEnumerable<string> wearableIds);
+        void RemoveWearablesInUse(IEnumerable<string> wearablesInUseToRemove);
+        void EmbedWearables(IEnumerable<WearableItem> wearables); // This temporary until the emotes are in the content server
```
lorux0, line 20: perhaps `[Obsolete("Will be removed in the future, when emotes are in the content server")]` ?

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,285 @@
...
+            return pageResponse.response.wearables;
+        }
+
+        public async UniTask<IReadOnlyList<WearableItem>> RequestThirdPartyWearablesByCollectionAsync(string userId, string collectionId, CancellationToken ct)
+        {
+            if (!thirdPartyCollectionPagePointers.TryGetValue((userId, collectionId), out var pagePointer))
+            {
+                thirdPartyCollectionPagePointers[(userId, collectionId)] = pagePointer = new (
+                    WEARABLES_BY_THIRD_PARTY_COLLECTION_END_POINT
+                       .Replace($"{userId}", userId)
```
lorux0, line 98: careful here, you are trying to replace the value of user id. You should remove the `$` to replace `"{userId}"` instead

unity-renderer/Assets/DCLServices/WearablesCatalogService/WearablesCatalogServiceProxy.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        public void Dispose()
+        {
+            kernelConfig.OnChange -= OnKernelConfigChanged;
+            wearablesCatalogServiceInUse?.Dispose();
+        }
+
+        public async UniTask<IReadOnlyList<WearableItem>> RequestOwnedWearablesAsync(string userId, int pageNumber, int pageSize, CancellationToken ct)
+        {
+            await UniTask.WaitUntil(() => isInitialized, cancellationToken: ct);
```
lorux0, line 53: instead of waiting, you could throw a `NonInitializedException`.. would that add too much complexity to handle it? same for the rest

unity-renderer/Assets/DCLServices/WearablesCatalogService/WebInterfaceWearablesCatalogService.cs
```diff
@@ -0,0 +1,374 @@
...
+                awaitingWearableTasks.Add(wearableId, taskResult);
+
+                // We accumulate all the requests during the same frames interval to send them all together
+                pendingWearablesToRequest.Add(wearableId);
+            }
+            else
+                taskResult = awaitingWearableTasks[wearableId];
+
+            ct.RegisterWithoutCaptureExecutionContext(() => taskResult.TrySetCanceled());
+            return await taskResult.Task;
```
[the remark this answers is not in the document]
lorux0: cancellation is handled in the line above: `ct.RegisterWithoutCaptureExecutionContext(() => taskResult.TrySetCanceled());`. Should it be different?

unity-renderer/Assets/DCLServices/WearablesCatalogService/WearablesCatalogServiceProxy.cs
```diff
@@ -0,0 +1,116 @@
...
+        public void Clear() =>
+            wearablesCatalogServiceInUse?.Clear();
+
+        private void OnKernelConfigChanged(KernelConfigModel currentKernelConfig, KernelConfigModel previous)
+        {
+            kernelConfig.OnChange -= OnKernelConfigChanged;
+
+            if (currentKernelConfig.urlParamsForWearablesDebug)
+            {
+                WebInterfaceWearablesCatalogService.Instance.Initialize(new WearablesWebInterfaceBridge(), wearablesCatalog);
```
lorux0, line 102: i think that the `WearableWebInterfaceBridge` should be injected into the proxy

unity-renderer/Assets/DCLServices/WearablesCatalogService/WearablesWebInterfaceBridge.cs
```diff
@@ -0,0 +1,13 @@
+﻿using DCL.Interface;
+
+namespace DCLServices.WearablesCatalogService
+{
+    public class WearablesWebInterfaceBridge
```
lorux0, line 5: I dont find a real purpose of this class at this point. Perhaps after you make the test coverage

lorux0 (changes requested):
Are you planning to do the test coverage later?

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,285 @@
...
+            if (!thirdPartyCollectionPagePointers.ContainsKey((userId, collectionId)))
+            {
+                thirdPartyCollectionPagePointers[(userId, collectionId)] = new (
+                    WEARABLES_BY_THIRD_PARTY_COLLECTION_END_POINT
+                       .Replace($"{userId}", userId)
+                       .Replace("{collectionId}", collectionId),
+                    1, ct, this);
+            }
+
+            var pageResponse = await thirdPartyCollectionPagePointers[(userId, collectionId)].GetPageAsync(1, ct);
```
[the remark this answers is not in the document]
sandrade-dcl: True! I forgot to adapt this request to pagination! Done in https://github.com/decentraland/unity-renderer/pull/4084/commits/56da2d152140706808ea855d25d7399da47bd9d3

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,285 @@
...
+
+                    // We accumulate all the requests during the same frames interval to send them all together
+                    pendingWearablesToRequest.Add(wearableId);
+                }
+                else
+                    awaitingWearableTasks.TryGetValue(wearableId, out taskResult);
+
+                return taskResult != null ? await taskResult.Task : null;
+            }
+            catch (Exception e) when (e is not OperationCanceledException) { throw; }
```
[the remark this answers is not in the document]
sandrade-dcl: What do you mean? In this case I only want to throw the exception when it is NOT an `OperationCanceledException`. How would you simplify that?

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,285 @@
...
+                return wearable;
+            }
+
+            ct.ThrowIfCancellationRequested();
+
+            try
+            {
+                UniTaskCompletionSource<WearableItem> taskResult;
+
+                if (!awaitingWearableTasks.ContainsKey(wearableId))
```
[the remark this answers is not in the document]
sandrade-dcl: Done in: https://github.com/decentraland/unity-renderer/pull/4084/commits/b14620e5872722252095b797e0f1b326c8d28770

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,285 @@
...
+        private async UniTaskVoid CheckForSendingPendingRequestsAsync(CancellationToken ct)
+        {
+            while (!ct.IsCancellationRequested)
+            {
+                await UniTask.DelayFrame(FRAMES_TO_CHECK_FOR_SENDING_PENDING_REQUESTS, cancellationToken: ct);
+
+                if (pendingWearablesToRequest.Count <= 0)
+                    continue;
+
+                string[] wearablesToRequest = pendingWearablesToRequest.ToArray();
```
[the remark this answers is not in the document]
sandrade-dcl: Sorry I don't understand, what do you mean with "use pool"?

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,285 @@
...
+
+            var pageResponse = await thirdPartyCollectionPagePointers[(userId, collectionId)].GetPageAsync(1, ct);
+            AddWearablesToCatalog(pageResponse.response.wearables.ToArray());
+
+            return pageResponse.response.wearables.ToArray();
+        }
+
+        public async UniTask<WearableItem[]> RequestWearablesAsync(string[] wearableIds, CancellationToken ct)
+        {
+            LambdaResponsePagePointer<WearableResponse> wearablesPagePointers = new (
```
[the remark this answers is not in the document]
sandrade-dcl: Done in: https://github.com/decentraland/unity-renderer/pull/4084/commits/a8456acf661785ce3195075cdccc1c993d82cdc9

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,285 @@
...
+                    1, ct, this);
+            }
+
+            var pageResponse = await thirdPartyCollectionPagePointers[(userId, collectionId)].GetPageAsync(1, ct);
+            AddWearablesToCatalog(pageResponse.response.wearables.ToArray());
+
+            return pageResponse.response.wearables.ToArray();
+        }
+
+        public async UniTask<WearableItem[]> RequestWearablesAsync(string[] wearableIds, CancellationToken ct)
```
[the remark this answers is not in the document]
sandrade-dcl: Done in: https://github.com/decentraland/unity-renderer/pull/4084/commits/36c908371b169a9b93eadcc4227018140f340a83

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,285 @@
...
+            lambdasService.Ref.Get<WearableResponse>(
+                "",
+                endPoint,
+                REQUESTS_TIME_OUT_SECONDS,
+                ATTEMPTS_NUMBER,
+                cancellationToken,
+                LambdaPaginatedResponseHelper.GetPageSizeParam(pageSize),
+                LambdaPaginatedResponseHelper.GetPageNumParam(pageNumber));
+
+        private string GetWearablesQuery(string[] wearableIds)
```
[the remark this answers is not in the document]
sandrade-dcl: Done in: https://github.com/decentraland/unity-renderer/pull/4084/commits/36c908371b169a9b93eadcc4227018140f340a83

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,285 @@
...
+        {
+            StringBuilder wearableIdsQuery = new StringBuilder();
+
+            foreach (string wearableId in wearableIds)
+                wearableIdsQuery.Append($"wearableId={wearableId}&");
```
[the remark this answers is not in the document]
sandrade-dcl: Done in: https://github.com/decentraland/unity-renderer/pull/4084/commits/12f7bcecdd12a634166fc192a5eb3ba819adf0ca

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,285 @@
...
+                    taskResult = new UniTaskCompletionSource<WearableItem>();
+                    awaitingWearableTasks.Add(wearableId, taskResult);
+
+                    // We accumulate all the requests during the same frames interval to send them all together
+                    pendingWearablesToRequest.Add(wearableId);
+                }
+                else
+                    awaitingWearableTasks.TryGetValue(wearableId, out taskResult);
+
+                return taskResult != null ? await taskResult.Task : null;
```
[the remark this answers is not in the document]
sandrade-dcl: Done in: https://github.com/decentraland/unity-renderer/pull/4084/commits/7818a1b81cca3ed738b8d9e06e48691de11433f9

unity-renderer/Assets/DCLServices/WearablesCatalogService/WearablesCatalogService.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        private void OnKernelConfigChanged(KernelConfigModel currentKernelConfig, KernelConfigModel previous)
+        {
+            kernelConfig.OnChange -= OnKernelConfigChanged;
+
+            if (currentKernelConfig.urlParamsForWearablesDebug)
+            {
+                WebInterfaceWearablesCatalogService.Instance.Initialize(new WearablesWebInterfaceBridge(), wearablesCatalog);
+                wearablesCatalogServiceInUse = webInterfaceWearablesCatalogService;
+                lambdasWearablesCatalogService?.Dispose();
```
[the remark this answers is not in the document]
sandrade-dcl: It's true that I don't need to dispose `lambdasWearablesCatalogService` in this case (so I have removed it), but I need to dispose the `webInterfaceWearablesCatalogService` in the `else` because I want to destroy its associated component that was instantiated in the scene (in the `Main` game object) to be able to receive messages from kernel. I decided to put this `destroy` operation inside its own  `Dispose` method.

unity-renderer/Assets/DCLServices/WearablesCatalogService/WearablesCatalogService.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            this.lambdasWearablesCatalogService = lambdasWearablesCatalogService;
+            this.webInterfaceWearablesCatalogService = webInterfaceWearablesCatalogService;
+            this.wearablesCatalog = wearablesCatalog;
+            this.kernelConfig = kernelConfig;
+        }
+
+        public void Initialize()
+        {
+            kernelConfig.EnsureConfigInitialized().Then(config => OnKernelConfigChanged(config, null));
```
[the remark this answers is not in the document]
sandrade-dcl: 1. I think making `EnsureConfigInitialized` async is something that is out of the scope of this PR as it is being used also by other systems. I would move it to an independent `tech debt` task.
2. I decided to listen only one time to `kernelConfig.OnChange` because we never thought of making this "proxy" as something that could change of service dynamically in real time. Once we receive the kernel config at the beginning, the proxy choose the service to use and that's it. Why keep the subscription in this case?

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,285 @@
...
+            AddWearablesToCatalog(pageResponse.response.wearables.ToArray());
+
+            return pageResponse.response.wearables.ToArray();
+        }
+
+        public async UniTask<WearableItem[]> RequestThirdPartyWearablesByCollectionAsync(string userId, string collectionId, CancellationToken ct)
+        {
+            if (!thirdPartyCollectionPagePointers.ContainsKey((userId, collectionId)))
+            {
+                thirdPartyCollectionPagePointers[(userId, collectionId)] = new (
```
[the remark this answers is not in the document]
sandrade-dcl: Hmmm... it's true. Any suggestion for a caching strategy here? The current `CatalogController` is not currently caching requests. In order to not to go too far off plan, maybe we should focus this refactor to keep the current behavior as it is and think on caching improvements for future steps. Wdyt?

unity-renderer/Assets/DCLServices/WearablesCatalogService/WearablesWebInterfaceBridge.cs
sandrade-dcl: I left them as `virtual` in order to be able to mock them in the tests (that are not still implemented)

unity-renderer/Assets/DCLServices/WearablesCatalogService/WearablesCatalogServiceProxy.cs
sandrade-dcl: Mikhail suggested me to wait instead of throwing exceptions here.

unity-renderer/Assets/DCLServices/WearablesCatalogService/WearablesWebInterfaceBridge.cs
sandrade-dcl: It was exactly for ease the tests (they're not still implemented). My idea was to mock this class in the tests.

unity-renderer/Assets/DCLServices/WearablesCatalogService/WearablesCatalogService.cs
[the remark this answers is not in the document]
mikhail-dcl: >I think making EnsureConfigInitialized async is something that is out of the scope of this PR as it is being used also by other systems. I would move it to an independent tech debt task.

You can already await it as it is a `Promise` and inherits `CustomYieldInstruciton`, you don't need to modify anything out of the scope of the task

>I decided to listen only one time to kernelConfig.OnChange because we never thought of making this "proxy" as something that could change of service dynamically in real time. Once we receive the kernel config at the beginning, the proxy choose the service to use and that's it. Why keep the subscription in this case?

Currently, you will call `OnKernelConfigChanged` twice: from the promise and from `OnChange` so the question is why it is called twice, not once?

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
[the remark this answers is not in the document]
mikhail-dcl: `catch (OperationCanceledException) { }`

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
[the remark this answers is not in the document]
mikhail-dcl: Check `ListPool<T>` and `PoolUtils`

unity-renderer/Assets/DCLServices/WearablesCatalogService/WearablesWebInterfaceBridge.cs
mikhail-dcl: I think it's a proper moment to introduce some tests. Otherwise, you will carry potential bugs to the next PR that is connected to the usage of the API only.

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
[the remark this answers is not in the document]
mikhail-dcl: 1. You can't leave it as is because it will not function properly for users, from the users perspective it's just a bug
2. Caching is a part of the task, and it's already here but just implemented not properly
3. With the API chosen you can't remove caching as it will lead to recreation of `PagePointer` every time a page is requested, even within the same view navigation. If you chose to expose `PagePointer` instead you could do simple caching on the view level from `Open` to `Close` (Disposing `PagePointer` on view close, and reusing the results while the view is open)
4. If I were to implement it I would introduce a generic cache with a lifetime for its entries, e.g. `PagePointer` is kept no longer than X seconds, and then on a subsequent request if the lifetime is exceeded it's recreated. We could go one step further and introduce tiers via Feature Flags Variants to A/B test UX. I would also limit the capacity for such cache according to LRU principle as it can become a problem.

unity-renderer/Assets/DCLServices/WearablesCatalogService/WebInterfaceWearablesCatalogService.cs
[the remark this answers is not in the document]
mikhail-dcl: I didn't notice that:
```
ct.RegisterWithoutCaptureExecutionContext(() => taskResult.TrySetCanceled());
```
It's incorrect as we reuse `taskResult` from the cache. Thus, we are attaching several independent `CancellationToken`'s to the same `CompletionSource`.
Instead, we should be attaching it to the task that projects onto `CompletionSource` (`taskResult.Task` returns a fresh copy every time is called).
But I would redesign the flow of waiting for `CompletionSources` as I proposed in one of the previous threads: it would eliminate this creation of `CompletionSource` and the `async` flow would look more natural/sequential

unity-renderer/Assets/DCLServices/WearablesCatalogService/WearablesCatalogService.cs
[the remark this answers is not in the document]
sandrade-dcl: > You can already await it as it is a Promise and inherits CustomYieldInstruciton, you don't need to modify anything out of the scope of the task

Ok, I will do it!

> Currently, you will call OnKernelConfigChanged twice: from the promise and from OnChange so the question is why it is called twice, not once?

Oh, I already know what did you mean... I didn't realize that I didn't need to subscribe to the `OnChange` event if I was already using `EnsureConfigInitialized` 👍 I will change it!

unity-renderer/Assets/DCLServices/WearablesCatalogService/WearablesWebInterfaceBridge.cs
sandrade-dcl: Yep, I agree. My idea was to start implementing tests just after applying all the feedback here and before starting the second part of the refactor 👍

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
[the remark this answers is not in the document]
sandrade-dcl: Done in: https://github.com/decentraland/unity-renderer/pull/4084/commits/09b6f3c9357cac935394f704164693e1db33d005

unity-renderer/Assets/DCLServices/WearablesCatalogService/IWearablesCatalogService.cs
sandrade-dcl: Done in: https://github.com/decentraland/unity-renderer/pull/4084/commits/b7714ff7b69fde03d8e16425229ce960d17f1136

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
sandrade-dcl: Good catch! I didn't relalize, thanks! Fixed in: https://github.com/decentraland/unity-renderer/pull/4084/commits/4077bbd1967325a7296de18bbf679dfa315bd344

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,285 @@
...
+            }
+            catch (Exception e) when (e is not OperationCanceledException) { throw; }
+        }
+
+        public void AddWearablesToCatalog(IReadOnlyList<WearableItem> wearableItems)
+        {
+            foreach (WearableItem wearableItem in wearableItems)
+            {
+                if (WearablesCatalog.ContainsKey(wearableItem.id))
+                    continue;
```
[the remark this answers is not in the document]
sandrade-dcl: In theory a wearable data shouldn't change as it is an NFT. Taking a look to the old `CatalogController` flow, it wasn't being replacing it neither. I prefer to keep the same flow for now.

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,285 @@
...
+            while (!ct.IsCancellationRequested)
+            {
+                await UniTask.Delay(TimeSpan.FromSeconds(TIME_TO_CHECK_FOR_UNUSED_WEARABLES), cancellationToken: ct);
+
+                if (wearablesInUseCounters.Count <= 0)
+                    continue;
+
+                List<string> wearablesToRemove = (from wearableInUse in wearablesInUseCounters
+                    where wearableInUse.Value <= 0
+                    select wearableInUse.Key).ToList();
```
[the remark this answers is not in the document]
sandrade-dcl: Done!

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,285 @@
...
+        {
+            WearablesCatalog = wearablesCatalog;
+        }
+
+        public void Initialize()
+        {
+            serviceCts = serviceCts.SafeRestart();
+
+            // All the requests happened during the same frames interval are sent together
+            CheckForSendingPendingRequestsAsync(serviceCts.Token).Forget();
```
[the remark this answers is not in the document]
sandrade-dcl: Done!

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,285 @@
...
+
+            // Check unused wearables (to be removed from our catalog) only every [TIME_TO_CHECK_FOR_UNUSED_WEARABLES] seconds
+            CheckForUnusedWearablesAsync(serviceCts.Token).Forget();
```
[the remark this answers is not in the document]
sandrade-dcl: Done!

unity-renderer/Assets/DCLServices/WearablesCatalogService/WearablesCatalogService.cs
```diff
@@ -0,0 +1,97 @@
...
+        public void Clear() =>
+            wearablesCatalogServiceInUse.Clear();
+
+        private void OnKernelConfigChanged(KernelConfigModel currentKernelConfig, KernelConfigModel previous)
+        {
+            kernelConfig.OnChange -= OnKernelConfigChanged;
+
+            if (currentKernelConfig.urlParamsForWearablesDebug)
+            {
+                WebInterfaceWearablesCatalogService.Instance.Initialize(new WearablesWebInterfaceBridge(), wearablesCatalog);
```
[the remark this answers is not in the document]
sandrade-dcl: Done!

unity-renderer/Assets/DCLServices/WearablesCatalogService/WearablesCatalogServiceProxy.cs
sandrade-dcl: Done!

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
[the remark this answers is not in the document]
sandrade-dcl: Finally, I have followed this approach to manage the cached pages, let's see what you think:

I have added a new boolean param `cleanCachedPages` in the signature of both paginated methods:
```
UniTask<IReadOnlyList<WearableItem>> RequestOwnedWearablesAsync(string userId, int pageNumber, int pageSize, bool cleanCachedPages, CancellationToken ct);
UniTask<IReadOnlyList<WearableItem>> RequestThirdPartyWearablesByCollectionAsync(string userId, string collectionId, int pageNumber, int pageSize, bool cleanCachedPages, CancellationToken ct);
```
Depending on that `cleanCachedPages` param, we will dispose the existing `PagePointer` and re-create it again. In this way we let the consumers to decide if they want to force the cleaning of the already cached pages or not before doing each request.

For example: When we open the Passport, we can request the first page of wearables using `cleanCachedPages = true`, but we can keep it as `cleanCachedPages = false` while we are navigating inside the open passport.

You can see the implementation here: https://github.com/decentraland/unity-renderer/pull/4084/commits/404bd835734608579c79f81b8c072d4f14f3c613

@mikhail-dcl wdyt?

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,308 @@
...
+
+            var pageResponse = await pagePointer.GetPageAsync(pageNumber, ct);
+            AddWearablesToCatalog(pageResponse.response.wearables);
+
+            return pageResponse.response.wearables;
+        }
+
+        public async UniTask<IReadOnlyList<WearableItem>> RequestBaseWearablesAsync(CancellationToken ct)
+        {
+            var serviceResponse = await lambdasService.Ref.Get<WearableResponse>(
```
mikhail-dcl, line 91: You should handle `pageResponse.success`

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,308 @@
...
+                "",
+                WEARABLES_BY_COLLECTION_END_POINT.Replace("{collectionId}", BASE_WEARABLES_COLLECTION_ID),
+                REQUESTS_TIME_OUT_SECONDS,
+                ATTEMPTS_NUMBER,
+                ct,
+                LambdaPaginatedResponseHelper.GetPageSizeParam(1),
```
mikhail-dcl, line 97: Is it ever paginated or you can delete these parameters?

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,308 @@
...
+            if (createNewPointer)
+            {
+                thirdPartyCollectionPagePointers[(userId, collectionId)] = pagePointer = new (
+                    WEARABLES_BY_THIRD_PARTY_COLLECTION_END_POINT
+                       .Replace("{userId}", userId)
+                       .Replace("{collectionId}", collectionId),
+                    pageSize, ct, this);
+            }
+
+            var pageResponse = await pagePointer.GetPageAsync(pageNumber, ct);
```
mikhail-dcl, line 128: You should handle `pageResponse.success`

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,308 @@
...
+
+        private async UniTask<WearableItem> SyncWearablesRequestsAsync(string newWearableId, CancellationToken ct)
+        {
+            if (!awaitingWearableTasks.TryGetValue(newWearableId, out var source))
+            {
+                pendingWearablesToRequest.Add(newWearableId);
+                awaitingWearableTasks[newWearableId] = source = new UniTaskCompletionSource<WearableItem>();
+            }
+
+            await UniTask.DelayFrame(FRAMES_TO_CHECK_FOR_SENDING_PENDING_REQUESTS, cancellationToken: ct);
```
mikhail-dcl, line 227: It can't work this way: you just skip a frame instead of synchronizing.
I proposed to `Yield(PostLateUpdate)` because `FRAMES_TO_CHECK_FOR_SENDING_PENDING_REQUESTS` = 1

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,308 @@
...
+                cancellationToken,
+                LambdaPaginatedResponseHelper.GetPageSizeParam(pageSize),
+                LambdaPaginatedResponseHelper.GetPageNumParam(pageNumber));
+
+        private async UniTask<WearableItem> SyncWearablesRequestsAsync(string newWearableId, CancellationToken ct)
+        {
+            if (!awaitingWearableTasks.TryGetValue(newWearableId, out var source))
+            {
+                pendingWearablesToRequest.Add(newWearableId);
+                awaitingWearableTasks[newWearableId] = source = new UniTaskCompletionSource<WearableItem>();
```
mikhail-dcl, line 224: One of the main reasons for my proposal was to avoid the creation of `UniTaskCompletionSource` per `WearableItem`. What you did here essentially means: We have a task that is related to `newWearableId` then we create one more `Task` inside the existing task to wait for the same `newWearableId`: it does not make much sense, does it?

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,308 @@
...
+                wearableIds.AddRange(pendingWearablesToRequest);
+                pendingWearablesToRequest.Clear();
+
+                (WearableResponse response, bool success) serviceResponse;
+
+                try
+                {
+                    serviceResponse = await lambdasService.Ref.Get<WearableResponse>(
+                        "",
+                        WEARABLES_BY_ID_END_POINT.Replace("{wearableIdsQuery}", GetWearablesQuery(wearableIds)),
```
mikhail-dcl, line 244: Don't pass Url encoded params as an `endPoint`, there is a separate argument for them

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,308 @@
...
+                var wearableIds = wearableIdsPool.GetList();
+                wearableIds.AddRange(pendingWearablesToRequest);
+                pendingWearablesToRequest.Clear();
+
+                (WearableResponse response, bool success) serviceResponse;
+
+                try
+                {
+                    serviceResponse = await lambdasService.Ref.Get<WearableResponse>(
+                        "",
```
mikhail-dcl, line 243: Don't omit `endPointTemplate` -> it is used for Sentry transactions. Just use `WEARABLES_BY_ID_END_POINT` because it is free from customized data

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,308 @@
...
+                using var wearableIdsPool = PoolUtils.RentList<string>();
+                var wearableIds = wearableIdsPool.GetList();
+                wearableIds.AddRange(pendingWearablesToRequest);
+                pendingWearablesToRequest.Clear();
+
+                (WearableResponse response, bool success) serviceResponse;
+
+                try
+                {
+                    serviceResponse = await lambdasService.Ref.Get<WearableResponse>(
```
mikhail-dcl, line 242: You should handle `pageResponse.success`

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,308 @@
...
+                        "",
+                        WEARABLES_BY_ID_END_POINT.Replace("{wearableIdsQuery}", GetWearablesQuery(wearableIds)),
+                        REQUESTS_TIME_OUT_SECONDS,
+                        ATTEMPTS_NUMBER,
+                        serviceCts.Token);
+                }
+                catch (Exception e)
+                {
+                    source.TrySetException(e);
```
mikhail-dcl, line 251: In your scheme, it is incorrect as you just set `exception` for one `CompletionSource`, what about other aggregated ones?

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,308 @@
...
+        }
+
+        public void EmbedWearables(IEnumerable<WearableItem> wearables)
+        {
+            foreach (WearableItem wearableItem in wearables)
+            {
+                WearablesCatalog[wearableItem.id] = wearableItem;
+
+                if (wearablesInUseCounters.ContainsKey(wearableItem.id))
+                    wearablesInUseCounters[wearableItem.id] = 10000; //A high value to ensure they are not removed
```
mikhail-dcl, line 196: Just set `int.MaxValue`

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,308 @@
...
+        }
+
+        public void RemoveWearablesInUse(IEnumerable<string> wearablesInUseToRemove)
+        {
+            foreach (string wearableToRemove in wearablesInUseToRemove)
+            {
+                if (!wearablesInUseCounters.ContainsKey(wearableToRemove))
+                    continue;
+
+                wearablesInUseCounters[wearableToRemove]--;
```
mikhail-dcl, line 185: It is the only place where you decrement references count. Thus, I don't see much sense of having a separate asynchronous `CheckForUnusedWearablesAsync`, you can remove wearables from the catalog right here, and as a bonus you will avoid allocation of:
```
var wearablesToRemove = from wearableInUse in wearablesInUseCounters
                    where wearableInUse.Value <= 0
                    select wearableInUse.Key;
```

mikhail-dcl (changes requested):

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
```diff
@@ -0,0 +1,308 @@
...
+            }
+
+            if (createNewPointer)
+            {
+                ownerWearablesPagePointers[userId] = pagePointer = new (
+                    WEARABLES_BY_OWNER_END_POINT.Replace("{userId}", userId),
+                    pageSize, ct, this);
+            }
+
+            var pageResponse = await pagePointer.GetPageAsync(pageNumber, ct);
```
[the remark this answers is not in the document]
sandrade-dcl: Done: https://github.com/decentraland/unity-renderer/pull/4084/commits/8b240ddbd614d8149c720b4b3b2e02086ba0d7ca

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
sandrade-dcl: Done: https://github.com/decentraland/unity-renderer/pull/4084/commits/8b240ddbd614d8149c720b4b3b2e02086ba0d7ca

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
sandrade-dcl: Done: https://github.com/decentraland/unity-renderer/pull/4084/commits/8b240ddbd614d8149c720b4b3b2e02086ba0d7ca

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
sandrade-dcl: Done: https://github.com/decentraland/unity-renderer/pull/4084/commits/8b240ddbd614d8149c720b4b3b2e02086ba0d7ca

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
sandrade-dcl: Yep, we can delete them. Done!

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
sandrade-dcl: Done!

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
sandrade-dcl: Done in: https://github.com/decentraland/unity-renderer/pull/4084/commits/5813c3cff8668732b635c988098655b706230923

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
sandrade-dcl: Done in: https://github.com/decentraland/unity-renderer/pull/4084/commits/62738dd3619fccb4409f386166de51b1e0cec8af

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
sandrade-dcl: You're completely right! I simply moved the old behavior to the new service and I didn't realize about this... good catch, it is a very good improvement. Done in: https://github.com/decentraland/unity-renderer/pull/4084/commits/489a8b92c614f7315161fd010b10554fd07dfc5c

unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs
sandrade-dcl: @mikhail-dcl I have apply the approach as you suggested initially but I need some help with a little detail. I understand why the first request is setting `lastRequestSource` to null BUT, because of this, it is making the rest of sent requests never ends. I mean, the current flow works in this way:
1. In the same frame we capture 3 requests.
https://github.com/decentraland/unity-renderer/blob/462934560cb767bd818cb7b5d631b93abbf09b33/unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs#L221

2. The first one creates a new `lastRequestSource`.
https://github.com/decentraland/unity-renderer/blob/462934560cb767bd818cb7b5d631b93abbf09b33/unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs#L225

3. The first one enters in the flow of `pendingWearablesToRequest.Count > 0` and then it caches the `lastRequestSource` into a local `source` and set `lastRequestSource` to null.
https://github.com/decentraland/unity-renderer/blob/462934560cb767bd818cb7b5d631b93abbf09b33/unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs#L228-L232

4. The first one send the request and keeps waiting.
https://github.com/decentraland/unity-renderer/blob/462934560cb767bd818cb7b5d631b93abbf09b33/unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs#L243-L248

5. The second and the third ones enter in the flow of `pendingWearablesToRequest.Count <= 0` and keep waiting for `lastRequestSource`.
https://github.com/decentraland/unity-renderer/blob/462934560cb767bd818cb7b5d631b93abbf09b33/unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs#L260-L261

6. The first request (who did the call to lambdas) finishes and return the result for that first wearable.
https://github.com/decentraland/unity-renderer/blob/462934560cb767bd818cb7b5d631b93abbf09b33/unity-renderer/Assets/DCLServices/WearablesCatalogService/LambdasWearablesCatalogService.cs#L265

8. But the other 2 wearables keep waiting indefinitely because `lastRequestSource` was set to null and will never resolves.

I have tried and if we don't set `lastRequestSource` to NULL, it works... but it would provoke a wrong behavior if we order more requests while the first one is still awaiting.

I can think of no other way to solve this than by managing a `UniTaskCompletionSource` by wearableId. If you come up with another solution, I'll be happy to try it.

sandrade-dcl: Done!

sandrade-dcl:
@mikhail-dcl could you review it? I already applied all the feedback from you and the only missing part would be the test coverage that I will be doing it in a separate PR. If we could approve and close this huge block, I would appreciate it 🙏 Thanks!

mikhail-dcl (approved):
Thank you for all the changes, now it looks good to me. Good job! 👍

lorux0 (approved):

PR #4144, opened 2023-01-27
Add video error load texture

PalauiDCL:
## Video Load Texture

When an error is thrown while loading a video we are currently showing this image: https://www.figma.com/file/xCXWcdGUXB0NpH5dtkRUBq/VideoLoadingFeedback?node-id=1%3A321&t=n0NXF3BEaeqUP5gp-1

The current approach has been using resources, but we could use an asset bundle in a different task to have this texture outside the build.

Aspects to keep in mind:
- Using a texture for this seems not optimal, in the future we should think of adding a canvas to show the error message and use less memory for it, (In this case we would need the sprites only). This change would require a lot more work though.
- Because of how the texture is handled, the texture is upside down (Flipped vertically).
- The message always adapts to the center.
- The extra space outside the centered message is filled with the background color of the passed texture.
- A further optimization for the texture is possible, but the code would become more difficult to understand. During the PR process an opinion on this aspect will be very welcome.

## QA
- Once a video can not be displayed by an error, a texture showing that there was an error should be present.
- If there is no way to reproduce a video error, I can force it from code for QA to test.


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/Plugins/WebVideoPlayer.cs
```diff
@@ -155,12 +157,49 @@ public float GetDuration()
...
+            return plugin.GetState(videoPlayerId);
         }
 
         public void Dispose()
         {
             plugin.Remove(videoPlayerId);
         }
+
+
+        private void UpdateTextureConservingAspectRatio(Texture2D textureToCenter, bool fillBackground)
```
PalauiDCL, line 169: Possible optimization of texture size: we can make the system force the texture passed to be understood as a square texture, this way, the texture does not need to be square and a good portion of the background can be removed.
The code would add a new parameter `bool useAsSquaredTexture`
When calculating the aspect ratio, if `useAsSquaredTexture` is `true`, only width would be used.

PalauiDCL:
image

Loading texture

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/Plugins/WebVideoPlayer.cs
```diff
@@ -37,19 +41,18 @@ public void Update()
...
                     string newError = plugin.GetError(videoPlayerId);
+                    if (newError == lastError)
+                        break;
 
-                    if ( newError != lastError )
-                    {
-                        lastError = newError;
-                        Debug.LogError(lastError);
-                    }
+                    UpdateTextureConservingAspectRatio(Resources.Load<Texture2D>("Textures/VideoFailed"), true);
```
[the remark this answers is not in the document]
PalauiDCL: Good catch, I modified it and added constants for the paths

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/Plugins/WebVideoPlayer.cs
```diff
@@ -29,6 +31,8 @@ public WebVideoPlayer(string id, string url, bool useHls, IVideoPluginWrapper pl
             this.plugin = plugin;
             this.url = url;
             plugin.Create(id, url, useHls);
+
+            UpdateTextureConservingAspectRatio(Resources.Load<Texture2D>("Textures/VideoLoading"), true);
```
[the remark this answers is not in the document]
PalauiDCL: Yes, I don't know if we should add it somehow during this task or in a posterior one, but there is no doubt that this does not need to be in the build

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/Plugins/WebVideoPlayer.cs
```diff
@@ -29,28 +33,30 @@ public WebVideoPlayer(string id, string url, bool useHls, IVideoPluginWrapper pl
...
 
                     break;
                 case VideoState.READY:
                     if (!isReady)
                     {
                         isReady = true;
-
-                        texture = plugin.PrepareTexture(videoPlayerId);
+                        //texture = plugin.PrepareTexture(videoPlayerId);
+                        SetAsError();
```
Kinerius, line 59: Is this intended? Wont this always show an error?

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/Plugins/WebVideoPlayer.cs
```diff
@@ -61,8 +67,8 @@ public void Update()
 
                     break;
                 case VideoState.PLAYING:
-                    if (visible)
-                        plugin.TextureUpdate(videoPlayerId);
+                    //if (visible)
+                    //    plugin.TextureUpdate(videoPlayerId);
```
Kinerius, line 71: Same here

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/Plugins/WebVideoPlayer.cs
```diff
@@ -155,12 +161,59 @@ public float GetDuration()
 
         public VideoState GetState()
         {
-            return (VideoState)plugin.GetState(videoPlayerId);
+            return plugin.GetState(videoPlayerId);
+        }
+
+        public void SetAsLoading()
+        {
+            UpdateTextureConservingAspectRatio(Resources.Load<Texture2D>(VIDEO_LOADING_PATH), true);
```
Kinerius, line 169: We should be loading that resource only once and cache the result, this is also a good opportunity to use addressable and remove that texture from the Resources folder

Kinerius (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/Plugins/WebVideoPlayer.cs
PalauiDCL: Addressables is the idea, it was commented also by Ashley, right now as I am forcing the error for QA, this PR should be ignored by devs ^^

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Video/Plugins/WebVideoPlayer.cs
PalauiDCL: As stated before, this is a test for QA, last changes will be reverted after approval

Kinerius (approved):

PR #4154, opened 2023-01-27
feat: block unfriend confirm

lorux0:
## What does this PR change?

Displays a new confirmation dialog before unfriending or blocking another user.

## How to test the changes?

1. Go to: https://play.decentraland.zone/?renderer-branch=feat/block-unfriend-confirm
2. Open the friend list
3. Click the three dots, open the contextual menu and then hit unfriend. A dialog should be opened
4. Open the contextual menu and hit block. A dialog should be opened
5. Open the passport of a friend, click unfriend or block. A dialog should be opened
6. Open a private chat window with a friend
7. Click three dots at the header, hit unfriend or block. A dialog should be opened

## Our Code Review Standards


kwsskulinski (approved):
Issue is fixed. The confirmation dialog is present in all mentioned cases.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Passport/Passport/PlayerInfo/PassportPlayerInfoComponentController.cs
```diff
@@ -238,10 +243,16 @@ private async UniTaskVoid AcceptFriendRequestAsync(CancellationToken cancellatio
         private void BlockUser()
         {
             if (ownUserProfile.IsBlocked(currentPlayerId)) return;
-            ownUserProfile.Block(currentPlayerId);
-            view.SetIsBlocked(true);
-            passportApiBridge.SendBlockPlayer(currentPlayerId);
-            socialAnalytics.SendPlayerBlocked(friendsController.IsFriend(currentPlayerId), PlayerActionSource.Passport);
+
+            dataStore.notifications.ConfirmationPopup.Set(ConfirmationPopupData.CreateBlockUserData(
```
AjimenezDCL, line 247: We have a whole notification system that allows customization, it's implemented as a queue and all the fancy stuff. Cannot we use it instead?

AjimenezDCL (approved):
Really solid PR, I just wonder if we can avoid the new ConfirmationPopup and reuse the notification system

RominaMarchetti (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Passport/Passport/PlayerInfo/PassportPlayerInfoComponentController.cs
lorux0: I think that the current notification system does not scale properly when you need new properties on a new type of notification and neither complies correctly with MVC/service/bridges architecture. In comparison, im adding properties for cancel button text and the cancel callback, and i don't need all the exposed properties in the notification's model. To support more kind of properties, the notification system model would have to be converted in some kind of god object, which i consider a smell pattern.
I think it is better to have specific type of notification implementations instead of a broad generic one. What do you think?

PR #4155, opened 2023-01-27
fix: negative loading value in loading screen

dalkia:
## What does this PR change?

Adds a check to avoid negative values in the ResourcesLoadTracker progress method

## How to test the changes?

1. Go to position=84,-15
2. See that you dont get a -200& loading

## Our Code Review Standards


kwssbocian (changes requested):
Loading screen when teleporting to 84,-15 still shows -200% progress.
image

dalkia:
Hey @kwssbocian ! Please try again, but use `https://play.decentraland.zone/?explorer-branch=fix%2Fnegative-percentage-on-loading-screen` as default url. Since there was a recent merge, I think that the current configuration is not grabbing the correct branch

kwssbocian (approved):
✔️Changes implemented correctly. After the fix now the loading screen no longer shows progress at -200%.

sandrade-dcl (approved):

AjimenezDCL (approved):

dalkia:
> Approved but I'm worried we are not fixing the underlying issue, why are we even getting negative values?

TBH, I had to drop it due to a matter of priorities.

The logical return of loading progress would be
```
int totalComponents = resourcesNotReady.Count + resourcesReady;
return totalComponents - resourcesReady / totalComponents * 100f
```

That was my first try to change it, because the current calculation which compares `(resourcesReady - pendingResourcesCount)  / resourcesReady` does not make sense to me to know what is the current state of loading.

But, when I changed to the one I thought logical, the system started malfunctioning. The value got resetted when loading was finished. This caused the timeout to be fired because when settling the position, the tracker was returning that the scene was not loaded when it was.

I started going down the rabbit hole, but I was almost half a day in. I could assign the time to check what is wrong with the tracker, but I had other priorities more related to sprint goals.

This check would solve the issue. Actually, the first option was to clamp it, but at least this one looked more aligned with the tracker.

AjimenezDCL:
> > Approved but I'm worried we are not fixing the underlying issue, why are we even getting negative values?
> 
> TBH, I had to drop it due to a matter of priorities.
> 
> The logical return of loading progress would be
> 
> ```
> int totalComponents = resourcesNotReady.Count + resourcesReady;
> return totalComponents - resourcesReady / totalComponents * 100f
> ```
> 
> That was my first try to change it, because the current calculation which compares `(resourcesReady - pendingResourcesCount) / resourcesReady` does not make sense to me to know what is the current state of loading.
> 
> But, when I changed to the one I thought logical, the system started malfunctioning. The value got resetted when loading was finished. This caused the timeout to be fired because ,when settling the position, the tracker was returning that the scene was not loaded when it was.
> 
> I started going down the rabbit hole, but I was almost half a day in. I could assign the time to check what is wrong with the tracker, but I had other priorities more related to sprint goals.
> 
> This check would solve the issue. Actually, the first option was to clamp it, but at least this one looked more aligned with the tracker.

Sure! let's have this issue tracked somewhere and merge this as mitigation

PR #4163, opened 2023-01-30
fix: global-scenes didn't have sdk7 in their serializable settings

leanmendoza:
## What does this PR change?
Close https://github.com/decentraland/sdk/issues/581 and add a test.

## How to test the changes?
Steps in the playground: 
1. Go to the playground with the next query params added: https://chore-support-dcl-explorer.sdk-playground.pages.dev/?&sdk-branch=chore/use-dcl-explorer&explorer-branch=fix/sdk7-global-scenes
2. When all is ready, click on Run and it should work
3. Check that is currently not working without this branch: https://chore-support-dcl-explorer.sdk-playground.pages.dev/?&sdk-branch=chore/use-dcl-explorer&explorer-branch=dev


eordano:
Can we not have static variables named "sdk7", and instead favor "runtimeVersion == 7"?

cc: @menduz

menduz:
The scene.json uses runtime version.

This is an internal struct

leanmendoza:
> It should be named `runtime7`, not `sdk7`, we keep carrying this error further...

I agree but can we defer it to an issue? This is a quick particular fix, and the internal naming comes from the kernel (when it creates the GlobalScene) and inherits the naming in the internal struct of unity (which is used in other places). 

Please react with 👍 and I'll create the issue `Move naming from 'sdk7' to runtime7 in the explorer repo`

kwsskulinski (approved):
Issue seems to be fixed. Playground works as intended.

AjimenezDCL (approved):
Regardless the on-going discussion, the PR looks solid. Approving from my side, I let you all decide what to do with the naming

PR #4168, opened 2023-01-30
fix: emotes performance degradation

Kinerius:
## What does this PR change?

Disabled the Animation component of loaded emotes to avoid the processing overhead

## Benchmark

Performance increase scales with the amount of NFT emotes that were loaded
In the next example there's a great amount of them and the overhead can be greatly reduced. 

### Prod
image

### This branch
image

## How to test the changes?

- Check that base emotes and NFT emotes work correctly.
- Check that you can see another avatar rather than yourself being animated correctly while their use an emote.

## Our Code Review Standards


lorux0 (approved):

kwsskulinski (approved):
Issue is fixed. Emotes animate correctly.

PR #4211, opened 2023-02-01
refactor: drop Airdrop system from browser interface and more

eordano:
Looking to increase readability of the code, this is a first approach to make `entryPoints` more readable. 

This PR:

* Standardizes the pattern of waiting for a redux store condition by observing on it with a function called `storeCondition`
* Drops all `Airdrop` related things
* Runs `make lint-fix` on all the code
* Splits up the `loadWebsiteSystems` and `startPreview` out of `packages/entryPoints/index.ts`

It also includes the changes from #4212 :

* Moves all of the `atomicHelper` components to a "lib" folder. The rationale is that it's easier to understand that "lib" will contain helper functions than understanding what "atomic" means.
* Drops those that are not in use internally in this package
* Drops the use of `namespace` in parcel limitations
* Drops the use of the deprecated value `visibleRadius`

Partially achieves:
* Standardizing the use of `base` to require files
* Enforcing the use of `import type` over `import`

The folders under `lib` signal the domain of the functionality provided by files found in that folder:
* `math` for Vector2, Vector3, and similar types
* `javascript` contains functions to make the code more idiomatic that could be used in any project 
  * `data-structures` can also be used in any javascript context
  * `encoding` uses the Buffer shim for the browser 
* `browser` contains functionality that only works on a browser context
* `decentraland` is for functions specific to Decentraland
* `redux` is for helper functions that interact with redux


browser-interface/packages/unity-interface/initializer.ts
```diff
@@ -138,7 +138,7 @@ export async function initializeUnity(options: KernelOptions['rendererOptions'])
   // wait until the renderer is fully loaded before returning, this
   // is important because once this function returns, it is assumed
   // that the renderer will be ready
-  await ensureUnityInterface()
+  await storeCondition(isRendererInitialized)
```
menduz, line 141: This is beautiful thanks

browser-interface/packages/entryPoints/hlsLoader.ts
```diff
@@ -1,7 +1,7 @@
 // ensure that Hls exist in the global context. To enable http live streaming in the browsers
-import _hls from "hls.js"
-declare var globalThis: any
-if (typeof globalThis.Hls === "undefined") {
+import _hls from 'hls.js'
+declare let globalThis: any
```
kuruk-mm, line 3:
```suggestion
declare const globalThis: any
```

kuruk-mm (approved):
It looks great!

PR #4212, opened 2023-02-01
refactor: move `atomicHelpers` to `lib`

eordano:
This PR:

* Moves all of the `atomicHelper` components to a "lib" folder. The rationale is that it's easier to understand that "lib" will contain helper functions than understanding what "atomic" means.
* Drops those that are not in use internally in this package
* Drops the use of `namespace` in parcel limitations
* Drops the use of the deprecated value `visibleRadius`

Partially achieves:
* Standardizing the use of `base` to require files
* Enforcing the use of `import type` over `import`

The folders under `lib` signal the domain of the functionality provided by files found in that folder:
* `math` for Vector2, Vector3, and similar types
* `javascript` contains functions to make the code more idiomatic that could be used in any project 
  * `data-structures` can also be used in any javascript context
  * `encoding` uses the Buffer shim for the browser 
* `browser` contains functionality that only works on a browser context
* `decentraland` is for functions specific to Decentraland
* `redux` is for helper functions that interact with redux


menduz (approved):

browser-interface/test/unit/positionThings.test.tsx
```diff
@@ -1,6 +1,6 @@
 import { expect } from 'chai'
 import { pickWorldSpawnpoint } from 'shared/world/positionThings'
-import { gridToWorld } from '../../packages/atomicHelpers/parcelScenePositions'
+import { gridToWorld } from "lib/decentraland/parcels/gridToWorld"
```
kuruk-mm, line 3:
```suggestion
import { gridToWorld } from 'lib/decentraland/parcels/gridToWorld'
```

kuruk-mm (approved):

PR #4219, opened 2023-02-02
Web Request Layer new API

PalauiDCL:
## What does this PR change?

This PR is aimed to refactor the WebRequestController extra layer by deprecating the old system, with the old system still working but in obsolete status, more frequent merges will be able to be made and the task can be divided into smaller bits.

The step for this PR:
- Create the new API and make old API obsolete [x]

For each step to be accepted all PlayTests must be correct.

Further steps will be implemented once this PR is accepted

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestController.cs
@@ -1,5 +1,7 @@
+using Cysharp.Threading.Tasks;
 using System;
 using System.Collections.Generic;
+using System.Threading;
 using UnityEngine;
 using UnityEngine.Networking;
 
@@ -13,7 +15,7 @@ namespace DCL
         private IWebRequestAudioFactory audioClipWebRequestFactory;
         private IPostWebRequestFactory postWebRequestFactory;
 
-        private readonly List<WebRequestAsyncOperation> ongoingWebRequests = new ();
+        private readonly List<WebRequestAsyncOperation> ongoingWebRequests = new();
 
         private WebRequestController(
             IWebRequestFactory getWebRequestFactory,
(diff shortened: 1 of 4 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestController.cs
```diff
@@ -51,12 +53,164 @@ public static WebRequestController Create()
...
+                {
+                    foreach (var item in headers)
+                        request.SetRequestHeader(item.Key, item.Value);
+                }
+
+                if (downloadHandler != null)
+                    request.downloadHandler = downloadHandler;
+
+                UnityWebRequestAsyncOperation asyncOp = request.SendWebRequest();
+                while (!asyncOp.isDone)
```
mikhail-dcl, line 175: You can replace the loop with `await asyncOp.WithCancellation`

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestController.cs
```diff
@@ -51,12 +53,164 @@ public static WebRequestController Create()
...
+                {
+                    await UniTask.Yield();
+                    if (cancellationToken.IsCancellationRequested)
+                    {
+                        request.Abort();
+                        break;
+                    }
+                }
+
+                if (asyncOp.webRequest == null)
```
mikhail-dcl, line 185: I doubt it is possible, is it?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestController.cs
```diff
@@ -51,12 +53,164 @@ public static WebRequestController Create()
...
+                    }
+                }
+                else
+                {
+                    onFail?.Invoke(asyncOp.webRequest);
+                    return asyncOp.webRequest;
+                }
+            }
+
+            return null;
```
mikhail-dcl, line 210: `null` in the `async` flow does not make much sense. Imagine that consumers will need to check it for nullity every time.
Consider introducing an exception if attempts are exceeded

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestController.cs
```diff
@@ -51,12 +53,164 @@ public static WebRequestController Create()
...
+            int timeout = 0,
+            CancellationToken cancellationToken = default)
+        {
+            audioClipWebRequestFactory.SetAudioType(audioType);
+            return await SendWebRequest(audioClipWebRequestFactory, url, null, onSuccess, onfail, requestAttemps,
+                timeout, cancellationToken);
+        }
+        
+
+        private async UniTask<UnityWebRequest> SendWebRequest<T>(
```
mikhail-dcl, line 148: I kinda agree that we don't need `disposeOnCompleted` here as it can be handled by consumers simply by wrapping in `using`. But was it intentional?

mikhail-dcl (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestController.cs
PalauiDCL: It was 100% intentional

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestController.cs
PalauiDCL: Agree, feels more natural, will be adding those changes

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/WebRequest/WebRequestController.cs
PalauiDCL: This should be never reachable code, not a single edge case should be able to reach there, that's why I think throwing an exception is a good idea, something really bad happened if we reach it, will change it.

mikhail-dcl (approved):
Looks good to me, thanks

PR #4221, opened 2023-02-02
fix: handle wearable timeouts

AjimenezDCL:
Handle properly wearable resolving timeouts


mikhail-dcl (approved):

kwsskulinski (approved):
Issue is fixed. No crashes and avatar is always visible.

PR #4240, opened 2023-02-03
fix: mitigate VR client not working

eordano:
The current hypothesis is that this is a race condition generated by the merge of `browser` and `renderer`. Judging by Rollbar metrics, the issue was not in production two days ago -- for the time being, this is making the VR client packed by [Ong](https://www.onginnovations.com/team) unusable for 100% of their users.


browser-interface/packages/entryPoints/index.ts
```diff
@@ -64,7 +64,7 @@ globalThis.DecentralandKernel = {
     }
 
     if (!Hls.isSupported()) {
-      throw new Error('HTTP Live Streaming is not supported in your browser')
+      // throw new Error('HTTP Live Streaming is not supported in your browser')
```
kuruk-mm, line 67: As a hotfix I will totally remove it. And we can add a url parameter to not check it.

kuruk-mm (approved):

PR #4254, opened 2023-02-03
Material Transition Controller Removal

### LINKED ISSUE #3119: Spike: Shader transition investigation
We want to drop the use of the `MaterialTransitionController.cs` and replace it with a shader. The objective of this spike is to analyze if its possible to create a transition effect using just shaders.

The MTC is a script that could be added to a Shape to generate a transition; this involved creating new transition materials which get destroyed at the end of the lifecycle. This was not optimal and lead to memory leaks in different corner cases.

We want to see if its possible to add the effect to the default material in DCL; and using just a function call that access that material to create the effect without the need of adding, removing or destroying anything. 

As a reference, here is the effect using the MTC on every object that appears: 
https://play.decentraland.zone/?renderer-branch=feat%2Fmtc-on-base-shape&realm=baldr&position=-12%2C0&island=Inwcf

In the absence of the Technical Artist the output of this spike will be:
- [x] Remove the transition effect
- [ ] Create a separate ticket for creating a new one

### LINKED ISSUE #4295: Remove Material Transition Controller
From the issue: https://app.zenhub.com/workspaces/workstream-core-experience-63aaac346e1c8d00104586e8/issues/gh/decentraland/unity-renderer/3119

This task will be used as a substitute as it was decided that the Material Transition Controller needs to be removed.

PalauiDCL:
fixes #4295 

## What does this PR do?

As commented with @Kinerius. he and @AjimenezDCL concluded that the best option was to remove the controller and clean all the affected code and related variables.

## QA
There is no material transition any more when showing loaded elements, (Sometimes called "loading visual feedback")
There should be no other changes from develop.


eordano (approved):
Approved. The green blockers should be enough "transition" for the time being, the extra performance and stability gained make up for the strange appearing of assets behind the green walls, we can revisit this decision later.

eordano:
Heads up, some rendering/scripting artifact is happening with the tram (I guess the model wasn't loaded well) in the Tram Line (50,50)

image

Kinerius (approved):
LFG

kwsskulinski (approved):
Issue is fixed. No problems have been found.

PR #4260, opened 2023-02-04
fix: close connect wallet modal when pressing ESC

eordano:
Steps to reproduce:

* Log in as guest, without wallet
* Click on a user
* You should see a "Connect your wallet" modal blocking the passport view
* Press Escape

What should happen with this fix:
* Both the Passport and the "Connect Wallet" modals get hidden

Error happening in production:
* The Passport is closed, but the "Connect Wallet" modal is still open


kwsskulinski (approved):
Issue is fixed. Wallet and Passport modals close after pressing ESC.

Kinerius (approved):

PR #4296, opened 2023-02-07
refactor: GPU Skinning Throttler

### LINKED ISSUE #4174: Refactor GPUSkinningThrottler
## Problem

- Each avatar has its own `GPUSkinningThrottler`
- Each instance of `GPUSkinningThrottler` creates an infinite Task (while true) that lasts until it's canceled.
- The algorithm waits for the `PostLateUpdate` to update the animations, it does this by creating another task wich has an `AttachExternalCancellation` which is slow and allocates memory.

### WebGL
- Create a `GPUSkinningThrottlerService` so a single instance of this object exists within our services.
- Inject the reference of the service into the avatar
- Register/Unregister the avatar accordingly based on its lifetime.
- Iterate through all avatars and update their `IGPUSkinning`
- Avoid using `AttachExternalCancellation`
- This can be a coroutine

### Desktop
- If the target platform is desktop, using Tasks is probably the right call since those can be sent to a thread. 
- Check if running the whole iteration on a thread is fast enough, avoid creating too many tasks, as it might be counterproductive. 

### Acceptance criteria
- Reduce the overhead produced by the throttler itself to nearly 0 ms.
- Reduce the memory allocation of the throttler itself to nearly 0 kb.

PalauiDCL:
Fixes #4174 

## What does this PR change?

This PR modifies how the GPU skinning throttler is accessed and centralizes it.
- Now the Throttler is a Service: `GPUSkinningThrottlerService` that can be accessed from the ServiceLocator.
- For WebGL we use a coroutine to control the flow of the throttler
- For Desktop we use an UniTask.

## QA
This throttler affects avatars only, so we must make sure that avatars behave the same as in `dev`
No changes should be observable


unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/AvatarShape/AvatarShape.cs
```diff
@@ -105,7 +105,7 @@ private void Awake()
                 visibility,
                 avatarLOD,
                 new SimpleGPUSkinning(),
-                new GPUSkinningThrottler(),
+                Environment.i.serviceLocator.Get<IGPUSkinningThrottlerService>(),
```
Kinerius, line 108: You can define 
`private Service<IGPUSkinningThrottlerService> skinningThrottlerService;`
and use `skinningThrottlerService.Ref` on this line instead

unity-renderer/Assets/Scripts/MainScripts/DCL/AvatarSystem/AvatarFactory.cs
```diff
@@ -21,7 +21,7 @@ public AvatarFactory(ServiceLocator serviceLocator)
                 visibility,
                 lod,
                 new SimpleGPUSkinning(),
-                new GPUSkinningThrottler(),
+                Environment.i.serviceLocator.Get<IGPUSkinningThrottlerService>(),
```
Kinerius, line 24: Same strategy here with `Service<IGPUSkinningThrottlerService>`

unity-renderer/Assets/Scripts/MainScripts/DCL/AvatarSystem/AvatarFactory.cs
```diff
@@ -41,7 +41,7 @@ IVisibility visibility
...
```
Kinerius, line 44: and here :D

unity-renderer/Assets/Rendering/Utils/GPUSkinning/GPUSkinningThrottlerService.cs
```diff
@@ -0,0 +1,101 @@
...
+    private IEnumerator ThrottleUpdate()
+    {
+        WaitForEndOfFrame wait = new WaitForEndOfFrame();
+
+        yield return wait;
+        while (!stopAsked)
+        {
+            foreach (KeyValuePair<IGPUSkinning, int> entry in gpuSkinnings)
+            {
+                if (Time.frameCount % entry.Value == 0)
```
Kinerius, line 77: I don't see the `value` of this Dictionary being ever changed, how are we counting frames for each `IGPUSkinning`?

unity-renderer/Assets/Rendering/Utils/GPUSkinning/GPUSkinningThrottlerService.cs
```diff
@@ -0,0 +1,101 @@
...
+        GPUSkinningThrottlerService service = new GPUSkinningThrottlerService();
+        if (initializeOnSpawn)
+            service.Initialize();
+        return service;
+    }
+
+
+    private IEnumerator ThrottleUpdate()
+    {
+        WaitForEndOfFrame wait = new WaitForEndOfFrame();
```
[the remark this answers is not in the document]
PalauiDCL: As this is only used in one of the versions (WebGL), I see the generation of the variable inside marginally more optimal.
Also, the loop only starts once per run usually.

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/AvatarShape/AvatarShape.cs
PalauiDCL: I will change it for this and the next ones ^^

unity-renderer/Assets/Rendering/Utils/GPUSkinning/GPUSkinningThrottlerService.cs
PalauiDCL: image
This method enables the modification, it is working as expected in the tests.

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/AvatarShape/AvatarShape.cs
AjimenezDCL: Im not so sold in the usage of a service here. This seems like a really specific system, does it make sense to make it  available for the whole platform?

Cannot we inject an instance created in the CoreComponentsPlugin to AvatarShape?

AjimenezDCL (changes requested):

unity-renderer/Assets/Rendering/Utils/GPUSkinning/GPUSkinningThrottlerService.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            Debug.LogError("GPUSkinningThrottlerService: Initialize called while stop was asked");
+            return;
+        }
+
+#if UNITY_WEBGL && !UNITY_EDITOR
+        CoroutineStarter.Start(ThrottleUpdate());
+#else
+        ThrottleUpdateAsync();
+#endif
```
AjimenezDCL, line 26: Why are we using coroutines? WebGL can handle unitasks as well.

AjimenezDCL (dismissed):

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/AvatarShape/AvatarShape.cs
AjimenezDCL: @Kinerius My only concern is having a service widely know in the environment to only being used by a specific system. To be honest this is not egregious or even a dirty implementation.

PS. You can reply without having to edit my comment hahaha

unity-renderer/Assets/Rendering/Utils/GPUSkinning/GPUSkinningThrottlerService.cs
AjimenezDCL: So far we have not found a case where a coroutine solves something a task cannot, as a rule of thumb always go first for the task approach first.

Kinerius: Yeah I agree, lets do a Task only routine

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/AvatarShape/AvatarShape.cs
Kinerius: What do you suggest then? I wanted to avoid having a static instance or a singleton decoupled from our services. 
`CoreComponentsPlugin` is more for SDK features

AjimenezDCL: I had in mind sharing the throttler between avatar shapes and leave the player controller and the previews out of this. Therefore we could have the instance in the `CoreComponentPlugin`.  But yeah, having a service, although I find it uglier, makes sense to wrap them all.

Then what about this, remove the static accessor in AvatarShape and inject the dependency in CoreComponentPlugin. Ideally only the plugins should have access to the static environment and the sub-systems should have their needs injected.

PalauiDCL:
Blocked due difference of opinions on how this task should be resolved:
- By a service
- By a more localized solution

Currently is solved as a service, is working and passing tests.

Kinerius (approved):

kwssbocian (approved):
✔️Changes implemented correctly. No issues with avatar appearance observed.

unity-renderer/Assets/Rendering/Utils/GPUSkinning/GPUSkinningThrottlerService.cs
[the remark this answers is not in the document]
mikhail-dcl: It was converted to `UniTask`, closing the thread

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/AvatarShape/AvatarShape.cs
mikhail-dcl: I think resolving `skinningThrottlerService` from `Awake` is a bigger problem than the global availability of the service itself considering that we don't have a mechanism of scopes.

mikhail-dcl: I changed it to use Avatar factory instead: now it looks much better imo

leanmendoza (approved):

kwssbocian (approved):
✔️After introduced changes to issues were observed.

PR #4309, opened 2023-02-07
feat: click own avatar opens passport

### LINKED ISSUE #3954: Open my on profile clicking on myself

lorux0:
## What does this PR change?

`Fixes #3954 `

The passport should be opened when clicking over your avatar.
Only available in 3rd person.
The outline effect has been added.
Works behind a feature flag: `click_own_avatar_passport`

## How to test the changes?

1. Go to: https://play.decentraland.zone/?explorer-branch=feat/click-own-passport
2. Switch to 3rd person camera
3. Hover the mouse over your avatar
4. The outline effect should be visible
5. Click over your avatar
6. The passport should be opened with your own profile
7. Check that the rest of the world features keeps working as expected in 1st and 3rd person camera, specially clicking & hovering other objects with the mouse in scenes
8. Go to: https://play.decentraland.org/?explorer-branch=feat/click-own-passport
9. The feature should be disabled

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/PlayerAvatarController/PlayerAvatarController.cs
```diff
@@ -228,20 +248,40 @@ private async UniTaskVoid LoadingAvatarRoutine(UserProfile profile, Cancellation
...
             };
+
             DataStore.i.player.ownPlayer.Set(player);
 
             enableCameraCheck = true;
             avatarCollider.gameObject.SetActive(true);
             CommonScriptableObjects.rendererState.RemoveLock(this);
             DataStore.i.common.isPlayerRendererLoaded.Set(true);
+
+            var entity = new DecentralandEntity
```
AjimenezDCL, line 267: I have taken a look below and it seems like it's almost ready to be compatible with null entities, we are only missing support in `CollidersManager.AddOrUpdateEntityCollider`. 

Can you try adding it there and avoid this entity? I'm approving regardless because maybe you will find more instances of usage of this entity down the rabbit hole that Im missing, if so, we can leave it as it is.

AjimenezDCL (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/PlayerAvatarController/PlayerAvatarController.cs
lorux0: I've uploaded the support for null entities. However the change could be kind of risky. I did find 46 usages of the entity. Many cases supports nulls, but there are corner cases that do not. On this particular case, i think it will not bring major issues, since only affects the pointer events. Will wait for QA's approval, so we dont introduce any major issue

kwssbocian (changes requested):
✔️On WebGL own avatar gets highlighted when moused over and clicking on it causes the passport to open. However feature appears to be unresponsive to the feature flag `explorer-click_own_avatar_passport' as on WebGL it does not have to be included for the feature to appear.
❌Feature does not work on Desktop regardless of the inclusion of the feature flag.
❌Additionally PR introduced a regression that has broken highlighting of other avatars. Mousing over them does not cause them to get outlined.

kwssbocian (approved):
✔️After the implemented fixes the features work correctly. Avatar highlighting, passport opening and the feature flags are working without issues and according to the specification.

PR #4321, opened 2023-02-08
feat: ignore messages sent in the request event

Julieta11:
In order to include the sent message in the request event in the chat history, we have decided to send the message to the room of the friendship ([check PR](https://github.com/decentraland/social-service/pull/140)). 

**This PR prevents the message from being notified as a new message at the moment the friend request is sent, since the users are not yet friends. However, if the request is accepted, the message if any will be notified as such.**


mikhail-dcl:
I removed myself as I am not to evaluate this

jmoguilevsky (approved):

PR #4322, opened 2023-02-08
refactor: big cleanup

eordano:
This should be mostly non-controversial:

* browser-interface/packages/shared/comms/peers.ts
  Changes in performance and aesthetics
* moved `logger` to `lib/logger` instead of `shared`
* `getIs` -> `is` can be added to the standard convention for redux selectors (as is in Java)
* moved `shared/profiles/utils` and `transformations` to `lib/decentraland/profiles`
* added `type` to many `imports`
* prefer absolute path from root rather than `../../../../../this`
* renamed `USER_AUTHENTIFIED` to `USER_AUTHENTICATED`
* introduced `waitFor` to wait on a store condition when inside of a saga


hugoArregui (approved):
regarding comms: LGTM. The other stuff is a bit above my pay grade :joy:

browser-interface/packages/kernel-web-interface/dcl-crypto.ts
```diff
@@ -0,0 +1,52 @@
...
+export type AuthLink = {
+  type: AuthLinkType
+  payload: string
+  signature: Signature
+}
+
+/**
+ * @public
+ */
+export enum AuthLinkType {
```
menduz, line 37: All the items in this file can be imported from `@dcl/schemas`

browser-interface/packages/kernel-web-interface/index.ts
```diff
@@ -0,0 +1,171 @@
...
+   */
+  hasStoredSession(address: string, networkId: number): Promise<{ result: boolean; profile?: any }>
+}
+
+/**
+ * @public
+ */
+export interface IDecentralandKernel {
+  initKernel(options: KernelOptions): Promise<KernelResult>
+}
```
menduz, line 171: I think most types in this file are repeated

browser-interface/packages/shared/profiles/sagas.ts
```diff
@@ -525,7 +526,7 @@ export async function generateRandomUserProfile(userId: string, reason: string):
   }
 
   if (!profile) {
-    profile = backupProfile(userId)
+    profile = randomProfile(userId)
```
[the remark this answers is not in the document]
eordano: Yes, this is a move-only change: check out https://github.com/decentraland/unity-renderer/blob/refactor/cleanup/browser-interface/packages/lib/decentraland/profiles/generateRandomUserProfile.ts

browser-interface/packages/shared/comms/debug.ts
```diff
@@ -1,174 +0,0 @@
-import { defaultLogger } from 'shared/logger'
```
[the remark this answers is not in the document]
eordano: I removed this file because it was used on `V2` only -- and no other files were importing from it!

browser-interface/packages/kernel-web-interface/dcl-crypto.ts
eordano: Agreed, but I'd rather keep this as "move-only" -- this Pull request is already big enough, I'll create another issue

browser-interface/packages/kernel-web-interface/index.ts
eordano: Same as above:  I'd rather keep this as "move-only" -- this Pull request is already big enough, I'll create another issue

browser-interface/packages/kernel-web-interface/index.ts
```diff
@@ -0,0 +1,171 @@
...
+   */
+  SIGN_UP = 'SIGN_UP',
+  WAITING_PROFILE = 'WAITING_PROFILE',
+  COMPLETED = 'COMPLETED'
+}
+
+/**
+ * @public
+ */
+export type DecentralandIdentity = AuthIdentity & {
```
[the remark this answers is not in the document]
eordano: Same as above: Id rather keep this as "move-only" -- this Pull request is already big enough, I'll create another issue

PR #4334, opened 2023-02-09
chore: faster parallel builds

### LINKED ISSUE #4329: [Spike] Remove dependency of `browser-interface` from `unity-renderer`
It looks like the build of `browser-interface` depends on the `unity-renderer` artifacts. This could be avoided, as `test-js-interface` does not depend on it. If this is about publishing the package, it should be done in a further step, as a `publish-*` phase.

eordano:
Closes #4329

image


kuruk-mm (approved):
Looks great!
to merge it, we need to change the `branch protection` settings with the name of the new steps

browser-interface/build.js
```diff
@@ -168,12 +168,7 @@ function createWorker(entry, outfile) {
 }
 
 async function compileJs() {
-  ensureFileExists(DIST_PATH, 'unity.loader.js')
-  const injectUnityPath = path.resolve(DIST_PATH, 'unity.loader.js')
-
-  for (let file of Object.values(generatedFiles)) {
-    ensureFileExists(DIST_PATH, file)
-  }
```
menduz, line 176: Every check has a story...

I'd keep the ensureFileExists, as a sanity check that the build finished succesfully

menduz: Nevermind, just saw it was moved to the next file

menduz (approved):

PR #4339, opened 2023-02-10
fix: Sentry script busting the cache

eordano:
[Offending code](https://github.com/decentraland/unity-renderer/blob/1632eb12d26eb5c7454690c2453b0430190ee217/scripts/src/inject-sentry-params.ts#L53)

Also, leaving the `.txt` with checksums as circleci artifacts for easier debugging of these kind of issues later

### DIFF
--- a/scripts/hash-unity-files.sh
+++ b/scripts/hash-unity-files.sh
@@ -7,6 +7,7 @@ find ./unity-renderer -type f \
     \( -not -path '*Library*' \) \
     \( -not -path '*browser-interface*' \) \
     \( -not -path '*node_modules*' \) \
+    \( -not -path '*Sentry/SentryOptions.asset' \) \
     \( -iname \*.unity \
     -o -iname \*.sh \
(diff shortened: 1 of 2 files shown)


scripts/hash-unity-files.sh
```diff
@@ -7,6 +7,7 @@ find ./unity-renderer -type f \
     \( -not -path '*Library*' \) \
     \( -not -path '*browser-interface*' \) \
     \( -not -path '*node_modules*' \) \
+    \( -not -path '*Sentry/SentryOptions.asset' \) \
```
kuruk-mm, line 10: Looks like this was deleted on https://github.com/decentraland/unity-renderer/pull/4334

kuruk-mm (approved):

scripts/hash-unity-files.sh
eordano: My bad, sorry!

PR #4381, opened 2023-02-13
feat: (sdk7 crdt) add append command and use for sending pointer events

leanmendoza:
# What? 
- Add `APPEND` command to CRDT protocol, storing as grow only set
- Send pointer event with `APPEND`


protocol-gen/package.json
```diff
@@ -15,7 +15,7 @@
     "typescript": "^4.2.3"
   },
   "dependencies": {
-    "@dcl/protocol": "1.0.0-4114477251.commit-ccb88d6",
+    "@dcl/protocol": "https://sdk-team-cdn.decentraland.org/@dcl/protocol/branch//dcl-protocol-1.0.0-4168098189.commit-e75240a.tgz",
```
menduz, line 18: New version already published

menduz (approved):

menduz (approved):
Tested and working as expected with its counterpart in SDK

PR #4382, opened 2023-02-13
feat: hardcode catalyst selection

eordano:
Because `eth` contract calls are pretty slow (particularly over metamask), the function to get the list of catalysts from the DAO contract was taking ~10 seconds out of every boot. This hard-codes the values into the code for a faster boot, without removing the logic to actually fetch the catalysts' list from the DAO contract.

### DIFF
--- a/browser-interface/packages/shared/web3.ts
+++ b/browser-interface/packages/shared/web3.ts
@@ -1,9 +1,9 @@
-import { ETHEREUM_NETWORK, ethereumConfigurations } from 'config'
-import { defaultLogger } from 'lib/logger'
-import { CatalystNode, GraphResponse } from './types'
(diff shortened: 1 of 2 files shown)


browser-interface/packages/shared/web3.ts
```diff
@@ -280,13 +280,42 @@ const catalystABI = [
...
+ * least through metamask)
+ */
 export async function fetchCatalystNodesFromDAO(): Promise<CatalystNode[]> {
   if (!requestManager.provider) {
     throw new Error('requestManager.provider not set')
   }
 
   const net = await getAppNetwork()
 
+  if (HARDCODED_CATALYST_LIST) {
```
menduz, line 294: I think this code is _never_ called, it is a fallback in the case peer.decentraland is down

eordano: In my tests it was :/

menduz (approved):

PR #4383, opened 2023-02-13
Refactor/catalyst comms

eordano:
This PR removes "dao" and renames it to "catalystSelection", pushing all "realm" stuff to the "realm" folder. It also drops the cache to localstorage, because it was not being used...


eordano:
Did not pass QA -- needs more work.

> Regarding the PR there seems to be an issue - it's getting stuck on the initial loading. First it was getting stuck on the loading of explorer-website (before it allows to choose either wallet or guest login), but now it's always getting stuck on "downloading the latest version of Decentraland". Cross-checking with `dev` and `main`, the issue occurs only on this branch

PR #4384, opened 2023-02-14
feat: parallel requests

eordano:
This initializes systems in parallel, saving a few seconds on initialization.

Test plan:
Ensure the rendering initializes correctly under the following scenarios, no difference between Web and Desktop
* New user, guest, no metamask
* New user, with metamask
* Existing user with a profile with metamask
* Existing user, with guest account and no metamask


browser-interface/packages/lib/browser/persistentStorage.ts
```diff
@@ -1,9 +1,7 @@
-import type { PersistentAsyncStorage } from 'kernel-web-interface'
-
 declare let window: any
 
-class PersistentLocalStorage implements PersistentAsyncStorage {
-  storage: any
+class PersistentLocalStorage {
```
kuruk-mm, line 3: Any reason that we're not using the interface? We could create a function that returns the interface and avoid using the `class`.

Basically, to avoid the following code above.
```
      setPersistentStorage(options.kernelOptions.persistentStorage as any)
```

I would like to have strong-typing

kuruk-mm (changes requested):

kwsskulinski (approved):
Rendering works ad intended. No differences between versions.

PR #4388, opened 2023-02-14
fix: setRemoteDescription failing on stable

eordano:
Tackles https://rollbar.com/decentraland/World/items/72012/ in a not-so-elegant way


hugoArregui (approved):
@eordano this look good to me, but we didn't work on voice chat, I think this is @kuruk-mm works

Issue #4393, opened 2023-02-14
spike: browser interface refactor proposal


2fd:
## Replace lodash

replace [lodash.uniqBy](https://github.com/decentraland/unity-renderer/blob/dev/browser-interface/packages/unity-interface/UnityInterface.ts#L269) with a function like the following can remove lodash from the build

```
Array.from(new Map(wearables.map(wearable=> [wearable.id, wearable])).values())
```

<img width="1420" alt="Screenshot 2023-02-14 at 08 33 21 copy" src="https://user-images.githubusercontent.com/208789/218726018-1bd7d56a-7e83-452b-831c-b4d0458fb42f.png">

## Dedupe bn.js

setting the global version of `bn.js` at `~4` will duplicate 6 of 8 modules

| current | suggested |
|--------|------------|
| <img width="569" alt="Screenshot 2023-02-14 at 08 37 09" src="https://user-images.githubusercontent.com/208789/218727423-7a659f5b-392f-422f-9d3b-8e83096078f1.png"> | <img width="574" alt="Screenshot 2023-02-14 at 08 36 21" src="https://user-images.githubusercontent.com/208789/218727470-b1644dce-0b21-40e0-8ea0-76a2739f5f39.png"> |

<img width="952" alt="Screenshot 2023-02-14 at 08 39 03" src="https://user-images.githubusercontent.com/208789/218727588-2541a6a9-9fe9-4095-8a02-07a498e5b22f.png">

## Fix `matrix-js-sdk` at `^21.0.0`

Version 22 of `matrix-js-sdk` includes the module `matrix-widget-api` that is not in use.

<img width="131" alt="Screenshot 2023-02-14 at 08 49 01" src="https://user-images.githubusercontent.com/208789/218730184-e87a2d32-2f23-461f-be17-84799a379b25.png">

## Move SDK workers to rollout

Currently, the SDK runtimes are included in the build even when they are your own package and are loaded as a worker.

<img width="229" alt="Screenshot 2023-02-14 at 08 41 57" src="https://user-images.githubusercontent.com/208789/218728426-595d5f4b-46f5-42c1-a406-8e84982dfcc8.png">

Moving them out of the build will reduce the bundle size, and loading them directly from the CDN using rollouts will prevent unnecessary reloads if they don't change and give the SDK team more flexibility to send new versions to production.

## Move HLS.js outside the build

[...]

PR #4395, opened 2023-02-14
fix: do not depend on lodash

eordano:
Partial fix for: https://github.com/decentraland/unity-renderer/issues/4393


menduz (approved):
is it also added as a dependency? or it is a transitive dep?

2fd (approved):

2fd:
❤️

PR #4400, opened 2023-02-14
chore: adaptations to make preview.html work and optimize downloads from the CDN using brotli

menduz:
Enables using custom renderer and explorer from preview mode and playground.

Also enables brotli compression of assets when downloading explorer directly from any of Decentraland's CDNs

Test steps:

Make sure it loads in the test links


pbosio (approved):

eordano (approved):
Works on my machine™

PR #4401, opened 2023-02-14
feat: sdk7 allow media permission

pbosio:
add required permission for external media urls.
Textures, Video, AudioStream.

permission must be added to `scene.json`
```
  "requiredPermissions": [
    "ALLOW_MEDIA_HOSTNAMES"
  ]
```
and hostnames must be added to scene's allow list
```
  "allowedMediaHostnames": [
    "somehost.com",
    "otherhost.xyz"
  ]
```

related to https://github.com/decentraland/sdk/issues/64
also fixes videos deployed with scene not working https://github.com/decentraland/sdk/issues/605


wacaine:
@pbosio will this work with video streams? An m3u8 (video stream) contains URLs to the actual video snippets + other stuff like subtitles, etc. that are under random CDN urls.   How can we predict that while white listing?


example

```
#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="audio-medium",NAME="audio",AUTOSELECT=YES,DEFAULT=YES,CHANNELS="2",URI="https://57vod-adaptive.akamaized.net/exp=1659580719~acl=%2F0e9760aa-5ab2-4acc-9a38-30b1b7256
```

menduz:
Yes will, it should work! I think it should only filters the "src" of the video and not the internal hls streams

@pbosio do you concur?

pbosio:
> Yes will, it should work! I think it should only filters the "src" of the video and not the internal hls streams
> 
> @pbosio do you concur?

That's correct.
@wacaine it will only affect the `scr` set in the video component. 
the url of the m3u8 manifest, in your example. urls inside the manifest are not being checked

unity-renderer/Assets/DCLPlugins/ECS7/Utils/Scene/Media.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                if (allowedDomains[i] == uri.Host)
+                    return true;
+            }
+        }
+
+        return false;
+    }
+
+    public static bool TrySetMediaUrl(string inputUrl, ContentProvider sceneContentProvider,
```
AjimenezDCL, line 39: Maybe I'm not understanding the method, but I would remove the `Set` from the name. It  looks like just getting the output url.

AjimenezDCL (approved):
Small request, good job!

unity-renderer/Assets/DCLPlugins/ECS7/Utils/Scene/Media.cs
pbosio: Yeah you are right, don't know why did I put "Set" instead of "Get" 🤷‍♂️
changed here https://github.com/decentraland/unity-renderer/pull/4401/commits/e708091992e57c4333122cf3468e84a68ea879ea

PR #4408, opened 2023-02-15
fix: `EmotesCatalogService` null reference exception on `Dispose`

pbosio:
`EmbedEmotes` is called on `finally`
which means that if task is cancelled it will try to use a ScriptableObject that might not be loaded
throwing a null reference exception


kuruk-mm (approved):

unity-renderer/Assets/DCLServices/EmotesCatalog/EmotesCatalogService.cs
```diff
@@ -31,11 +31,17 @@ public EmotesCatalogService(IEmotesCatalogBridge bridge, IAddressableResourcePro
...
+
         try
         {
             addressableCTS = new CancellationTokenSource();
             embeddedEmotesSO = await addressableResourceProvider.GetAddressable<EmbeddedEmotesSO>("EmbeddedEmotes.asset", addressableCTS.Token);
         }
+        catch (OperationCanceledException e)
+        {
+            taskCancelled = true;
+        }
```
sandrade-dcl, line 44: Another option, in order to avoid to create a new boolean variable, would be:

```
try
{
	...
}
catch (OperationCanceledException) { return; }
catch (Exception e)
{
	...
}

EmbedEmotes();
```

sandrade-dcl (approved):

unity-renderer/Assets/DCLServices/EmotesCatalog/EmotesCatalogService.cs
pbosio: AFAIK `finally` would be called regardless of the `return` in the `catch`

sandrade-dcl: Yes, for this reason I remove the `finally` block in my suggested alternative. Anyway, this was only another way to do the same as you already have. Feel free to use what you prefer.

pbosio: oops! sry, didn't notice that :)
I didn't want to remove the finally because of my lack of context regarding this class, but if you say I can remove it, I'll remove it 😎

AjimenezDCL (approved):

unity-renderer/Assets/DCLServices/EmotesCatalog/EmotesCatalogService.cs
```diff
@@ -50,10 +51,8 @@ private async UniTaskVoid InitializeAsyncEmbeddedEmotes()
             DisposeCT();
             InitializeAsyncEmbeddedEmotes();
         }
-        finally
-        {
-            EmbedEmotes();
-        }
+
+        EmbedEmotes();
```
mikhail-dcl, line 55: I think `EmbedEmotes()` should be moved after `embeddedEmotesSO`, otherwise, it will be called several times, won't it? I know it is not your code but can we improve it?

mikhail-dcl (changes requested):

unity-renderer/Assets/DCLServices/EmotesCatalog/EmotesCatalogService.cs
pbosio: sorry, it is already merged.
you should add an issue to the explorer board for that, also I noticed that there is no test coded for the "retry" logic, so it might be worth to add that coverage to the issue

mikhail-dcl: Moved to #4486

PR #4410, opened 2023-02-15
feat: spawn in middle of scene by default

eordano:
This prevents awkward interactions with greenblockers if scenes around have not been loaded. Scene owners are free to fix their spawn point to this position anyways!

Test plan:
- Enter a scene with a defined start position (for example, genesis plaza). Ensure that scene defined spawn point is used
- Enter a scene with no spawn position defined (for example, 46,74). Ensure that you spawn in the middle of the scene, and not strictly in the edge of a corner (a relative 0,0 of the scene)
- Enter a scene with a 0,0 spawn position defined (for example, aetheria at 110,100). Ensure that you spawn strictly in the corner

### DIFF
--- a/browser-interface/packages/shared/world/positionThings.ts
+++ b/browser-interface/packages/shared/world/positionThings.ts
@@ -107,10 +107,13 @@ function pickSpawnpoint(land: Scene): InstancedSpawnPoint {
     spawnPoints = [{
       position: {
         x: [1, 15],
-        y: [1, 15],
+        y: [0, 0],
         z: [1, 15]
       },
     }]
+  }
+  if (!spawnPoints) {
+    throw new Error(`Invalid spawnpoint definition`)
   }
 
   // 1 - default spawn points
--- a/browser-interface/packages/shared/world/positionThings.ts
+++ b/browser-interface/packages/shared/world/positionThings.ts
@@ -103,7 +103,7 @@ export function pickWorldSpawnpoint(land: Scene): InstancedSpawnPoint {
 
 function pickSpawnpoint(land: Scene): InstancedSpawnPoint {
   let spawnPoints = land.spawnPoints
-  if (!spawnPoints || spawnPoints.length === 0) {
+  if (!Array.isArray(spawnPoints) || spawnPoints.length === 0) {
     spawnPoints = [{
       position: {
         x: [1, 15],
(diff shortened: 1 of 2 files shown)


browser-interface/packages/shared/world/positionThings.ts
```diff
@@ -103,16 +101,26 @@ export function pickWorldSpawnpoint(land: Scene): InstancedSpawnPoint {
   }
 }
 
-function pickSpawnpoint(land: Scene): InstancedSpawnPoint | undefined {
-  if (!land || !land.spawnPoints || land.spawnPoints.length === 0) {
-    return undefined
+function pickSpawnpoint(land: Scene): InstancedSpawnPoint {
+  let spawnPoints = land.spawnPoints
+  if (!spawnPoints || spawnPoints.length === 0) {
```
menduz, line 106:
```suggestion
  if (!spawnPoints || !Array.isArray(spawnPoints) || spawnPoints.length === 0) {
```

kuruk-mm (approved):
Looks great! We should highlight this on the release notes with the following documentation https://docs.decentraland.org/creator/development-guide/scene-metadata/#spawn-location

browser-interface/packages/shared/world/positionThings.ts
eordano: Maybe even...
```suggestion
  if (!Array.isArray(spawnPoints) || spawnPoints.length === 0) {
```

nearnshaw:
@eordano 
Hey, I'm not convinced this change will cause more good than harm.

1) It's not great to start in a corner, but it's much much worse to start stuck inside a collider and not be able to move from there
2) Most every scene created via the SDK in the last couple of years has an explicitly defined spawn point at 0,0,0, simply because all of the templates come with that spawn point by default and most creators don't bother to change that. That means that they won't be affected by this change.
3) Builder scenes don't have an explicit spawn point, and have no mechanism to set one. So they will always be random
4) Old abandoned scenes will likely not change their spawn points, and keep being random

I think the end result of this will be that any new scene with potentially high traffic won't be affected, but when spawning into random parts of the map with old content players will end up stuck inside colliders and have a much worse experience than they currently do

Perhaps a better solution is to explicitly change the spawn location in all our template scenes, and at least lazy creators will have a better default.

PR #4433, opened 2023-02-16
feat: new map renderer map controller

AjimenezDCL:


unity-renderer/Assets/DCLServices/MapRendererV2/MapCameraController/MapCameraControllerInternal.cs
```diff
@@ -0,0 +1,70 @@
...
+using DCLServices.MapRendererV2.MapLayers;
+using System;
+using UnityEngine;
+using Object = UnityEngine.Object;
+
+namespace DCLServices.MapRendererV2.MapCameraController
+{
+    internal class MapCameraController : IMapCameraControllerInternal
+    {
+        private static readonly Vector2 ZOOM_CAMERA_SIZE = new (150, 400); // TODO Evaluate if we need to expose this in the constructor
```
mikhail-dcl, line 11: I believe, we need to adapt the render texture size and aspect ratio to the requested size of the region, what do you think?

mikhail-dcl (changes requested):
I think it's time to connect `RentMapCameraController`, `IMapCameraController RentCamera(in MapCameraInput cameraInput)` and these class together and see if we miss anything

mikhail-dcl (approved):

PR #4446, opened 2023-02-17
fix: disabling skybox camera when fullscreen ui is visible

aixaCode:
## What does this PR change?

When showing fullscreen UI there is not point in rendering anything under. We already disable main camera, but skybox was still rendering

## How to test the changes?

Open and close main menu. There should not be any visible change, skybox should be visible.

## Our Code Review Standards


kwsskulinski (approved):
Skybox is always visible.

PR #4455, opened 2023-02-17
fix: textures memory leak

pbosio:
Textures set to Materials were never disposed until the scene was unloaded.

- Material was waiting for texture to finish loading before setting it reference, which lead to the race condition of changing the texture for that material before the previous texture was loaded, leaving an orphan texture.
fixed by setting material reference to texture regardless of the texture download state.
- Texture were never disposed, even if they were removed from all materials.
fixed by disposing textures when reference count equals 0

note: texture are not removed from component manager after disposed, given the texture nature of disposable component means that the texture component should exist even if there are no reference to it.

fixes https://github.com/decentraland/sdk/issues/607

how to test:
https://play.decentraland.org/?explorer-branch=fix/texture-leak
go around the world, textures and videos should be loading correctly


Kinerius (approved):
Nice!

lorux0 (approved):

kwsskulinski (approved):
Textures load correctly. No issues detected.

PR #4457, opened 2023-02-24
feat: logger for why reconnection is attempted

eordano:
This adds the `getStack` function to `lib` to debug stack traces. It also joins multiple `select` calls into a single one (`reconnectionState`)


browser-interface/packages/shared/comms/sagas.ts
```diff
@@ -394,13 +393,12 @@ function* handleCommsReconnectionInterval() {
       USER_AUTHENTICATED: take(USER_AUTHENTICATED),
       timeout: delay(1000)
     })
+    // TODO: Why are we not doing `if (reason === undefined) continue`?
```
kuruk-mm, line 396: It should make sense to add the 'if (reason === undefined) continue'. If you want to add it will be ok. Or just leave the comment to thing about it in the future.

kuruk-mm (approved):

browser-interface/packages/shared/comms/sagas.ts
eordano: Another PR!

PR #4463, opened 2023-02-27
feat: drop bouncy castle

eordano:
Create a cert locally using this: https://stackoverflow.com/questions/42786986/how-to-create-a-valid-self-signed-x509certificate2-programmatically-not-loadin


aixaCode:
Can you give a bit more context about the changes here? :)

kuruk-mm:
> Can you give a bit more context about the changes here?

We're using [BouncyCastle.NET](https://github.com/bcgit/bc-csharp) for creating a self-signed certificate which is used to connect to the Unity WebSocket Server through SSL.

This PR should remove the [BouncyCastle.NET](https://github.com/bcgit/bc-csharp) dependency and use something native from .NET, removing a huge dependency of the project.

kuruk-mm (approved):

eordano:
> Can you give a bit more context about the changes here?

Some early feedback:
- This reduces the build times by ~1 minute
- Reduces the build size by ~1Mb (compressed: [before](https://cdn.decentraland.org/@dcl/explorer/1.0.85138-20230301155511.commit-28c8090/unity.wasm.br), [after](https://renderer-artifacts.decentraland.org/branch/feat/drop-bouncy/unity.wasm.br))

PR #4476, opened 2023-02-28
refactor: single select to increase efficiency of profiles

eordano:
The `fetchProfile` function makes 5 calls to `yield select`. This makes the fetching of profiles more efficient by only issuing one `select` effect.


jmoguilevsky:
Great readability refactor!

jmoguilevsky (approved):

PR #4501, opened 2023-03-01
feat: mutual friends

Julieta11:
Retrieving the addresses of the friends that are shared between the authenticated user and the user specified in the request.


jmoguilevsky (approved):

browser-interface/packages/renderer-protocol/inverseRpc/services/mutualFriendsService.ts
```diff
@@ -0,0 +1,22 @@
+import {
+  GetMutualFriendsResponse,
+  MutualFriendsKernelServiceDefinition
+} from '@dcl/protocol/out-ts/decentraland/renderer/kernel_services/mutual_friends_kernel.gen'
+import { RpcServerPort } from '@dcl/rpc'
+import * as codegen from '@dcl/rpc/dist/codegen'
+import { RendererProtocolContext } from '../context'
+import { getMutualFriends } from 'shared/friends/sagas'
+
+export function registerMutualFriendsKernelService(port: RpcServerPort<RendererProtocolContext>) {
```
agusaldasoro, line 10: I'm not sure that I understand the flow of this registration, who is calling it?

browser-interface/packages/renderer-protocol/inverseRpc/services/mutualFriendsService.ts
```diff
@@ -0,0 +1,22 @@
+import {
+  GetMutualFriendsResponse,
+  MutualFriendsKernelServiceDefinition
+} from '@dcl/protocol/out-ts/decentraland/renderer/kernel_services/mutual_friends_kernel.gen'
+import { RpcServerPort } from '@dcl/rpc'
```
braianj, line 5: just a small contribution. maybe better something like 
`import { RpcServerPort } from '@dcl/rpc/dist/types'`

browser-interface/packages/renderer-protocol/inverseRpc/services/mutualFriendsService.ts
Julieta11: It's called from [here](https://github.com/decentraland/unity-renderer/blob/c56c628d39c23d0eeea4b721b0d5b1dbefd148c2/browser-interface/packages/renderer-protocol/inverseRpc/rpcServer.ts#L24-L30), and as far as I understand, it's added to a registration list. Each registered service is then used to map ports to their corresponding function implementations.

Issue #1807, opened 2022-01-20
Change default port of the WebSocketCommunication


kuruk-mm:
Currently, to debug we use port 5000. In Mac M1 that port is used by the ControlCenter and there are some programs that used it too according to wikipedia:

![image.png](https://images.zenhubusercontent.com/60647b38cc4b79842895cb82/49eb48eb-5d55-4fb6-9913-6d2b0afae433)
Reference: https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers

We have a system to detect available ports. But when we're debugging without SSL, we only are allowed to use port 5000. So there's no other option.

I propose to change the default port to 7666, which is not listed on the Wikipedia webpage.

Remember to change here:

https://github.com/decentraland/definitions/blob/108887b87546cb59b8b90764ec2e9b4548d1a25e/src/domains.ts#L16

https://github.com/decentraland/definitions/blob/108887b87546cb59b8b90764ec2e9b4548d1a25e/src/domains.ts#L34

popuz:
@kuruk-mm, when changing `startPort` from **5000** to **7666**, should I also change `endPort` from **5100** to **7766**? (Line 88 of `WebSocketCommunication` constructor).
![image.png](https://images.zenhubusercontent.com/631628793fb7f60ae36ae984/9765b1ce-1ae0-461a-81d1-b8005fab2734)

PR #4526, opened 2023-03-03
fix: load self-signed cert from base64 and not a file, and add a comm…

kuruk-mm:
…and line to disable SSL on Desktop by Command Line

The self-signed certificate from file was not being loaded on the build.

The new parameter `--without-ssl` should work for debugging.

## Our Code Review Standards


m3taphysics (approved):

eordano (approved):

unity-renderer/Assets/Desktop/Scripts/MainScripts/DCL/MainDesktop/MainDesktop.cs
```diff
@@ -68,7 +68,9 @@ protected override void InitializeCommunication()
 
 #if UNITY_EDITOR
                 withSSL = DebugConfigComponent.i.webSocketSSL;
-                startPort = 5000;
+                startPort = 7666;
```
eordano, line 71: I like the change, `5000` is common, is this used somewhere else?

unity-renderer/Assets/Desktop/Scripts/MainScripts/DCL/CommandLineParserDesktop/CommandLineParserUtils.cs
```diff
@@ -26,6 +27,9 @@ public static void ParseArguments()
                             i++; // shift
                             startPort = int.Parse(arguments[i]);
                             break;
+                        case "--without-ssl":
```
eordano, line 30: Might I suggest

```suggestion
                        case "--no-ssl":
```
as an attempt to simplify the command line usage?

unity-renderer/Assets/Desktop/Scripts/MainScripts/DCL/MainDesktop/MainDesktop.cs
eordano, line 71: I like the change, `5000` is too common, is this used somewhere else by default where we should change it?

eordano (approved):
Non-blocking comments, approved

unity-renderer/Assets/Desktop/Scripts/MainScripts/DCL/MainDesktop/MainDesktop.cs
kuruk-mm: We have already decided to use 7666 instead of 5000; this was missing. The Launcher is set to 7666, so this change has no production change. It's mainly for easier profiling and debugging.

More details about the use: https://github.com/decentraland/unity-renderer/issues/1807

PR #4537, opened 2023-03-06
fix: wait for initial position to render remote avatar

### LINKED ISSUE #4142: [QA] Some of the avatars are loaded partially below the floor in the Genesis Plaza
Description:
Some of the avatars are loaded partially underground. This issue occurs only in Genesis Plaza. There is no issue with loading in WonderMine and ICE Poker.

Reproduction rate: 
5/5

Steps to reproduce:    
1. Login to the Decentraland (guest or wallet).
2. Enter the Genesis Plaza.
3. Observe the issue.
 
Results:    
Some of the avatars are loaded slightly below the ground.
 
Expected results:
All of the avatars are loaded at the proper height.
 
Notes:    
- Issue occurs for both production and dev branches.

dalkia:
## What does this PR change?

Fixes #4142

Starts sending position and rendering to players only when the rendering state is true. (or, in other words, when the loading screen is out)

To achieve this, the ActivateRendering logic is finally passed to renderer. For now, the ACK messages are still there. 

Finally, it decouples the dependency on the `loadingScreen`. 

## How to test the changes?

You need two DCL players to test this. Keep in mind that other players may still appear half way through, and this is due that they dont have the proper sending of information in their builds. 

Also, its easier to understand the fix if you first understand what is broken. I recommend to do the following to see the bug. I recommend to test in a low populated realm (like `&realm=hera`) to avoid the noise from other avatars

Situation 1:

1. Client 1: Enter DCL normally. Get to Genesis Plaza
2. Client 2: Create a new user. Get to the avatar creation scene. Leave it there, minimize the tab
3. Client 1: Go to the floor ground. There should be an avatar guest through the floor around the bar. This is a bug
4. Client 2: Complete the avatar creation. Get to to the loading screen, and minimize.
5. Client 1: The guest avatar is still stuck on the ground.
6. Client 2: Go back to the window. Wait for loading screen to be gone
7. Client 1: Finally, the guest avatar is on the correct place. But probably, it does not have the correct settings (avatar name or wearables chosen)

Situation 2:

1. Client 1: Enter DCL normally. Get to Genesis Plaza
2. Client 2: With the new client you created in Situation 1, enter DCL. Get to the loading screen, and minimize. Dont let it finish
3. Client 1: Check that the Client 2 is stuck middle ground on the upper section of genesis plaza

Both situations are fixed with this PR. 

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
```diff
@@ -157,17 +162,13 @@ public override IEnumerator ApplyChanges(BaseModel newModel)
...
-            // To deal with the cases in which the entity transform was configured before the AvatarShape
-            if (!initializedPosition && scene.componentsManagerLegacy.HasComponent(entity, CLASS_ID_COMPONENT.TRANSFORM))
-            {
-                initializedPosition = true;
-                OnEntityTransformChanged(entity.gameObject.transform.localPosition,
-                    entity.gameObject.transform.localRotation, true);
-            }
+            yield return new UnityEngine.WaitUntil(PositionCorrectlyInitialized);
+            
+            Debug.Log("DEBUG: " + entity.gameObject.transform.localPosition);
```
mikhail-dcl, line 167: Please remove the log

mikhail-dcl (changes requested):
Looks good to me

unity-renderer/Assets/Scripts/MainScripts/DCL/Components/Avatar/AvatarShape.cs
dalkia: Thanks Misha! Sorry for the silly mistake

kwssbocian (approved):
✔️Testing on an empty realm resulted in the other avatars always being correctly placed at ground level during spawn at Genesis Plaza.

mikhail-dcl (approved):

PR #4561, opened 2023-03-08
refactor: try to improve entrypoint readability

eordano:
Brings back Hls check -- VR client is broken for other reasons


browser-interface/packages/entryPoints/index.ts
```diff
@@ -25,137 +21,164 @@ import type { RootState } from 'shared/store/rootTypes'
...
+}
+
+function ensureHLSCapability() {
+  if (!Hls || !Hls.isSupported) {
+    throw new Error('HTTP Live Streaming did not load')
+  }
+
+  if (!Hls.isSupported()) {
+    throw new Error('HTTP Live Streaming is not supported in your browser')
+  }
```
kuruk-mm, line 83: `Hls.isSupported()` is a function according Hls.js.

So this code should be like:
```suggestion
  if (!Hls) {
    throw new Error('HTTP Live Streaming did not load')
  }

  if (!Hls.isSupported()) {
    throw new Error('HTTP Live Streaming is not supported in your browser')
  }
```

I didn't find Hls.isSupported as a variable.

kuruk-mm (approved):

browser-interface/packages/entryPoints/index.ts
eordano: I believe this is testing if the function exists so as to not call `undefined`

PR #4593, opened 2023-03-09
chore: add frame time by profiler recorder

### LINKED ISSUE #4431: Add frame time CPU and GPU measurements
### Description
In the project, there are multiple ways of reporting fps and frame time. We should standardize the approach and use ProfilerRecorder API instead of `LinealBufferFPSCounter` which provides not consistent results.
For supported platforms, it would be great to have GPU render time

### Acceptance Criteria
- [x] Update all places using fps and frame time (`FPSDisplay`, `PerformanceMetricsController`, `PerformanceMeterController`)
- [x] Add render time (`FPSDisplay`, `PerformanceMetricsController`, `PerformanceMeterController`)

### Notes
To keep the data compatible with the metrics we still need to send frame time in the same format in  WebInterface.PerformanceReportPayload (frame time is multiplied by 1000)

For collecting data we can use Unity API
```
cpuTimeRecorder = ProfilerRecorder.StartNew(ProfilerCategory.Internal, "Main Thread", 15);
gpuTimeRecorder = ProfilerRecorder.StartNew(ProfilerCategory.Internal, "Render Thread", 15);
```

We also have LinealBufferHiccupCounter that is using frame time to detect hiccups and to keep the compatibility we can feed it with the frame time, and no changes are needed.

popuz:
## What does this PR change?
Closes #4431 

- Replaced old FPS counting approach by one from ProfilerRecorder API
- samples sent to backend now use `unscaledDeltaTime` instead of `deltaTime`
- Move all `ProfilerRecorder` usages in one service class

Allocations reduction:
- Removed dynamic allocations in the `InputController.Update()`
- Removed dynamic call of `Enum.GetValues` in `ECSPointerInputSystem.Update`

## How to test the changes?

0. Launch the explorer

### Counter:
1. Run /showfps in the chat window
2. Observe FPS and frame time counting "correctly" (jump in Ice Poker to see the difference in performance or adjust graphic settings for that)

### InputControlelr
1. Press ALT+T and verify that chat recording started
2. Press B+0-9 and verify that respective emotes is triggered by this shortcut

Devs only: would be nice to check that hiccups counting works correctly.

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Debugging/Performance/PerformanceMeterController/PerformanceMeterController.cs
```diff
@@ -325,26 +292,26 @@ private void ReportData()
         /// </summary>
         private int CalculatePerformanceScore()
         {
-            double desiredFrameTime = Settings.i.qualitySettings.Data.fpsCap ? 1000/30.0 : 1000/60.0;
-            double frameScore = Mathf.Min((float)(desiredFrameTime/ averageFrameTime), 1); // from 0 to 1
-            double hiccupsScore = 1 - (float) totalHiccupFrames / samples.Count; // from 0 to 1
+            double desiredFrameTime = Settings.i.qualitySettings.Data.fpsCap ? 1000 / 30.0 : 1000 / 60.0;
```
menduz, line 295: Usually to detect stuttering, desiredFrameRate is dynamic and depends on the actual measurements of each machine, but this may fall outside of this PR. What I've seen in the while is the `desiredFrameRate = avg(frames(from p40 to p60))` https://www.vortez.net/articles_pages/frame_time_analysis,2.html

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Debugging/Performance/LinealBufferHiccupCounter.cs
```diff
@@ -1,67 +1,41 @@
...
         {
-            if (IsHiccup(counter.values[counter.end]))
+            float lastFrameTimeInSec = profilerRecordsService.LastFrameTimeInSec;
+
+            if (IsHiccup(lastFrameTimeInSec))
             {
-                hiccupsCountInBuffer -= 1;
-                hiccupsSum -= counter.values[counter.end];
+                HiccupsCountInBuffer -= 1;
+                HiccupsSum -= lastFrameTimeInSec;
```
menduz, line 25: Please add tests to ensure this works exactly like the previous one. I'd even suggest to revert this specific piece, the new implementation is not doing the same at plain sight (can't confirm since tests are no longer there). The previous one used a O(1) circular buffer and removed the oldest measurement from the buffer. This one is removing the newest.

The correct behavior of this buffer is crucial for all of our performance dashboards, unless it has serious problems I'd prevent touching it. If we change the measurement method suddenly all of our new data is not comparable with the previous.

The rest of the PR looks sensible though

menduz (changes requested):

Kinerius (approved):
Left a comment, everything else looks good 👌

menduz (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Debugging/Performance/LinealBufferHiccupCounter.cs
popuz: after our discussion and small investigation, I realized that these measurements are not used in the backend. Because
`counter.values` array is never send to the backend and we send `PerformanceMetricsController.encodedSamples` array instead. As for hiccups, as you mentioned, we calculate them differently on the backend (however, we still send these ones 🤔 maybe to support `old_hiccups` chart, I don't know). Anyway, I brought back old `counter.values` array for hiccups counting.

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Debugging/Performance/PerformanceMeterController/PerformanceMeterController.cs
popuz: extracted it [here](https://github.com/decentraland/unity-renderer/issues/4613)

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/Debugging/Performance/ProfilerRecordsService.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        public void StopRecordGCAllocatedInFrame()
+        {
+            if (isRecordingAdditionalProfilers)
+                return;
+
+            gcAllocatedInFrameRecorder.Dispose();
+        }
+
+        private float GetRecorderFrameAverage(ProfilerRecorder recorder)
```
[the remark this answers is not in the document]
popuz: This one is only used in a specific scenario - when debug window is open (for example via `/showfps` command in chat). Currently the buffer set to 15 and needed only to prevent stuttering of the value in the view (we can increase it to 30). With 15 the overhead is minimal (taking into account it was recorded with DeepProfiler overhead)
image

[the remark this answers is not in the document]
popuz: `unsafe` version was more performant, so I changed to it (and also added comments for awarness)

Benchmark:
- for 200 samples with `List`
image
- versus 200 samples on `unsafe`
image

kwssbocian (approved):
✔️Frame timer is present in the /showfps panel. No issues observed regarding its functionality.

dalkia (approved):

PR #4594, opened 2023-03-10
fix: build profiling documentation and fixes

m3taphysics:
## What does this PR change?

- Updates specific documentation on profiling local builds
- Fixes some errors in the build scripts on windows

## Our Code Review Standards

### DIFF
--- a/docs/how-to-profile-a-local-unity-build-remotely.md
+++ b/docs/how-to-profile-a-local-unity-build-remotely.md
@@ -25,4 +25,4 @@
 
 - Recommended setup: Only CPU, Rendering and Memory modules on display; Disable VSync and "Others" from CPU module.
 
-- On WebGL the process of building via the editor `bakes` the `local ip` into the build itself. If this changes, it will no longer work and you will have to build again. This only happens when you have "AutoConnect To Profiler" ticked in unity.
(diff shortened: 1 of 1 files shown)


docs/how-to-profile-a-local-unity-build-remotely.md
```diff
@@ -25,3 +24,5 @@ codeUrl: resolveWithBaseUrl('unity.wasm')
 - If the data is not transmitted automatically, fiddle with toggling/untoggling the record button in the Profiler until the samples start arriving at Unity
 
 - Recommended setup: Only CPU, Rendering and Memory modules on display; Disable VSync and "Others" from CPU module.
+
+- On WebGL the process of building via the editor `bakes` the `local ip` into the build itself. If this changes, it will no longer work and you will have to build again.
```
eordano, line 28:
```suggestion
- On WebGL the process of building via the editor bakes-in the `local ip` into the build itself. This means that the profiling build will strictly only work if the machine that compiled it has the same IP. If your IP changes, profiling will no longer work and you will have to build again.
```

eordano (approved):

docs/how-to-profile-a-local-unity-build-remotely.md
AjimenezDCL: I would also specify that the "ip baking" only occurs when building with `Autoconnect to profiler` is ticked.

PR #4607, opened 2023-03-13
fix: memory alloc fixes

m3taphysics:
## What does this PR change?

- Removes some simple allocations in SDK7

## Our Code Review Standards


unity-renderer/Assets/DCLPlugins/ECS7/Systems/ScenesUiSystem/ECSScenesUiSystem.cs
```diff
@@ -18,6 +18,8 @@ public class ECSScenesUiSystem : IDisposable
         private readonly BaseList<IParcelScene> loadedScenes;
         private readonly BaseVariable<bool> loadingHudVisibleVariable;
 
+        private static HashSet<IParcelScene> scenesToSort = new HashSet<IParcelScene>();
```
menduz, line 21: Please add comments why this change was required to prevent it from rolling back. Comment-less changes are fragile

unity-renderer/Assets/DCLPlugins/ECS7/Systems/PointerInputSystem/ECSPointerInputSystem.cs
```diff
@@ -86,7 +85,7 @@ private static void Update(State state)
                 if (curState[i] != prevState[i])
                 {
                     PointerEventType pointerEventType = curState[i] ? PointerEventType.PetDown : PointerEventType.PetUp;
-                    InputAction inputAction = (InputAction)inputActionEnum[i];
+                    InputAction inputAction = (InputAction)_inputActionEnums[i];
```
pbosio, line 88: I think this could be just replaced by `inputAction inputAction = (InputAction)i;` and avoid the `Enum.GetValues` completely
@leanmendoza could you confirm?

unity-renderer/Assets/DCLPlugins/ECS7/Systems/ScenesUiSystem/ECSScenesUiSystem.cs
pbosio, line 21: If we want to remove the allocation of the hashset, I wouldn't add that field as static.
I think it would probably be better to set it as part of the class state and modify `ApplyParenting` to receive it as argument

m3taphysics: Because it allocates memory at runtime causing memory fragmentation, increased memory usage, garbage collection stalls and hiccups

unity-renderer/Assets/DCLPlugins/ECS7/Systems/PointerInputSystem/ECSPointerInputSystem.cs
m3taphysics: This should work I will wait for @leanmendoza though

unity-renderer/Assets/DCLPlugins/ECS7/Systems/ScenesUiSystem/ECSScenesUiSystem.cs
m3taphysics: What comment are you expecting?

menduz: Assuming there is a serial killer on the team, you need to prevent them to say "ugh, why is this static? I'll move it to a class property, it is not compatible with our standards" and then they introduce a regression in the allocations and want to come after you

Then something like `// This is a static property because ...`

pbosio: I wouldn't make that property as static, just part of the class instance state should be fine,
plus we avoid any "static property controversy" :)

m3taphysics: Agree! No idea why I didn't do it this way originally, good spot

m3taphysics: I agree with @pbosio it shouldn't be a static anyway!

mikhail-dcl (approved):

pbosio (approved):

PR #4621, opened 2023-03-14
feat: (preview mode) enable pipe the console .log and .error to the parent in a iframe

leanmendoza:
## What does this PR change?
This enables to pipe of the console calls to an iframe parent

Editor counterpart: https://github.com/decentraland/editor-sdk7/pull/12

## Why?
https://github.com/decentraland/sdk/issues/631

### DIFF
--- a/browser-interface/packages/lib/logger/logger.ts
+++ b/browser-interface/packages/lib/logger/logger.ts
@@ -53,6 +53,51 @@ function createDefaultLogger(type: 'kernel' | 'unity', prefix: string, subPrefix
   }
 }
 
+export function createForwardedLogger(type: 'kernel' | 'unity', prefix: string, subPrefix: string = ''): ILogger {
+  const targetWindow = globalThis.window.parent || globalThis.window
+  const kernelPrefix = `${type}:${prefix} `
+
+  function forwardMessage(type: string, args) {
+    targetWindow && targetWindow.postMessage({ type, payload: { args } }, '*')
+  }
+
+  return {
+    error(message: string | Error, ...args: any[]): void {
+      const computedArgs = [kernelPrefix, subPrefix, message, ...args]
+      if (typeof message === 'object' && message.stack) {
+        computedArgs.push(message.stack)
+      }
+
+      console.error.apply(null, computedArgs)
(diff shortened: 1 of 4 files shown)


browser-interface/static/preview.html
```diff
@@ -537,6 +537,23 @@
...
+            const targetWindow = window.parent || window
+            targetWindow && targetWindow.postMessage({ type, payload: args }, '*')
+            originalFunctionCallback(...args)
+          }
+        }
+
+        console.log('Pipe the console log and error enabled.')
+
+        globalThis.originalConsole = { log: globalThis.console.log, error: globalThis.console.error }
+        globalThis.console.log = pipe('log', globalThis.originalConsole.log)
```
menduz, line 554: I think should only pipe the scene's console from the dev adapter. Wdyt?

menduz (changes requested):

browser-interface/static/preview.html
leanmendoza: I think we'd miss a lot of info in the editor if we only handle the scene logs.

menduz: Isn't that the only thing that matters? There are toons of useless logs there, if we'd like to debug we can open vscode's own console to see the logs

leanmendoza: It **should be** the only thing that matters. But this enables the possibility to filter other things that can matter. I agree the Dev-Adapter approach is the ideal one

menduz (approved):
amazing thank you <3

browser-interface/packages/lib/logger/logger.ts
```diff
@@ -53,6 +53,51 @@ function createDefaultLogger(type: 'kernel' | 'unity', prefix: string, subPrefix
   }
 }
 
+export function createForwardedLogger(type: 'kernel' | 'unity', prefix: string, subPrefix: string = ''): ILogger {
+  const targetWindow = globalThis.window.parent || globalThis.window
```
eordano, line 57:
```suggestion
  const targetWindow = globalThis.window?.parent || globalThis.window
```

cazala (approved):

PR #4625, opened 2023-03-14
fix: disable offscreen culling for SkinnedMeshRenderer by feature flag

AjimenezDCL:
CullingController is messing up with the Offscreen Update of SkinnedMeshRenderer even with the CullingController disabled. 

Ideally objects should be constructed in a proper manner so the pivot doesnt stray too far away from the model (even when animated). 

Since we are evaluating removing the whole system I added a feature flag to allow the changes in `Offscreen Update` or not so we can A/B test it's implications in performance.

## Test
This is hard to test, you have to find a skinned mesh renderer (like the humanoids walking around genesis plaza). With `ENABLE_SMR_OFFSCREEN_UPDATE_STATUS` and the Detail Culling in Settings disabled luckily you will be able to reproduce that looking at the models from weird angles make them dissapear. On contrary, with not using `ENABLE_SMR_OFFSCREEN_UPDATE_STATUS` or explicitly using `DISABLE_SMR_OFFSCREEN_UPDATE_STATUS` this should never happen.

### DIFF
--- a/unity-renderer/Assets/Rendering/Culling/CullingController.cs
+++ b/unity-renderer/Assets/Rendering/Culling/CullingController.cs
@@ -17,6 +17,7 @@ namespace DCL.Rendering
     public class CullingController : ICullingController
     {
         private const string ANIMATION_CULLING_STATUS_FEATURE_FLAG = "animation_culling_status";
+        private const string SMR_OFFSCREEN_UPDATE_STATUS_FEATURE_FLAG = "smr_offscreen_update_status";
         private const bool DRAW_GIZMOS = false;
         internal List<CullingControllerProfile> profiles = null;
 
@@ -34,6 +35,7 @@ namespace DCL.Rendering
         private bool playerPositionDirty;
         private bool objectPositionsDirty;
         private bool running = false;
+        private bool offScreenUpdate = true;
 
         // Cache to avoid allocations when getting names
         private readonly HashSet<Shader> avatarShaders = new HashSet<Shader>();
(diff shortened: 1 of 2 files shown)


Kinerius (approved):

kwsskulinski (approved):
Issue with disappearing characters does not happen.

unity-renderer/Assets/Rendering/Culling/CullingController.cs
```diff
@@ -195,7 +198,8 @@ internal IEnumerator ProcessProfileWithEnabledCulling(CullingControllerProfile p
                     if (IsAvatarRenderer(mat))
                         shouldHaveShadow &= TestAvatarShadowRule(profile, distance);
 
-                    skr.updateWhenOffscreen = false;
+                    if (offScreenUpdate)
+                        skr.updateWhenOffscreen = false;
```
popuz, line 202: this looks a bit strange (but maybe I miss details) - when `offScreenUpdate` is enabled, then disable `updateWhenOffscreen` on skinned meshes. Is it because then it handled by some our system?

popuz (approved):

unity-renderer/Assets/Rendering/Culling/CullingController.cs
AjimenezDCL: `offScreenUpdate` defines if we are updating `UpdateWhenOffscreen` or not. If you have a better naming I'm all for it!

PR #4611, opened 2023-03-13
fix: race condition on PX

eordano:
## What does this PR change?
There is a reported race condition with wearables. This PR fixes it.

## How to test the changes?
1. Launch the explorer with this link: https://play.decentraland.org/?GLOBAL_PX=urn%3Adecentraland%3Aentity%3Abafkreihj3zxbsd4ekz5ifgpwhqzuglod4s46353yis333c3w74uasgjgpu%3FbaseUrl%3Dhttps%3A%2F%2Fsdk-content-server.decentraland.org%2Fipfs%2F&realm=baldr&position=58%2C52&island=baldr9kdc
2. Do it many times :sweat_smile: reproducibility of the race condition is ~2/5
3. Ensure this doesn't happen:

image

PR #4632, opened 2023-03-15
Align dev with Release 2023 03 14

mikhail-dcl:
## What does this PR change?

...

## How to test the changes?

1. Launch the explorer
2. ...

## Our Code Review Standards


browser-interface/Makefile
```diff
@@ -9,7 +9,7 @@ CWD = $(shell pwd)
 # SDK Files
 EMPTY_SCENES := public/empty-scenes/common
 
-build: empty-parcels
+build:
```
menduz, line 12: This is a required change, please do not remove it
```suggestion
build: empty-parcels
```

browser-interface/packages/shared/meta/sagas.ts
```diff
@@ -37,6 +42,38 @@ export function* metaSaga(): any {
   }
 
   yield put(metaConfigurationInitialized(merge))
+
+  yield fork(fetchInitialPortableExperiences)
+}
+
+function* fetchInitialPortableExperiences() {
```
menduz, line 49: This PR reverts an important fix for race conditions needed for the fashion week. Please do not include the reverted commit of https://github.com/decentraland/unity-renderer/pull/4611 in `dev`

menduz (changes requested):
This reverts the PR https://github.com/decentraland/unity-renderer/pull/4611 which is required to fix an important race condition for the FW next week, please do not include the reverted commit in `dev`

browser-interface/packages/shared/meta/sagas.ts
mikhail-dcl: We need to make a fix properly then as it does not work on Standalone according to QA

mikhail-dcl: So what's the plan @menduz @eordano ? We can do it via `dev` and then we can align `main` with `dev` or make a `hotfix`?

PR #4637, opened 2023-03-15
refactor: small allocation improvements

fabio-dee:
## What does this PR change?

Made minor improvements on memory allocations.

...

## How to test the changes?

Changes involve Ui. Would be wise to test if Ui (both menu and scenes like ICE Poker) behaves normally.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Prefabs/Resources/Bridges.prefab
+++ b/unity-renderer/Assets/Prefabs/Resources/Bridges.prefab
@@ -15,7 +15,6 @@ GameObject:
   - component: {fileID: 1314399029881169376}
   - component: {fileID: 3653805652683081651}
   - component: {fileID: 7701290377581287658}
-  - component: {fileID: 6402082374581292810}
   - component: {fileID: 607285373800589754}
   - component: {fileID: 2728691134734170003}
   m_Layer: 0
@@ -110,18 +109,6 @@ MonoBehaviour:
   m_Enabled: 1
   m_EditorHideFlags: 0
   m_Script: {fileID: 11500000, guid: 7382bddbf9f0d460683567eea269f211, type: 3}
-  m_Name: 
-  m_EditorClassIdentifier: 
---- !u!114 &6402082374581292810
-MonoBehaviour:
-  m_ObjectHideFlags: 0
-  m_CorrespondingSourceObject: {fileID: 0}
-  m_PrefabInstance: {fileID: 0}
-  m_PrefabAsset: {fileID: 0}
-  m_GameObject: {fileID: 6401575379531892522}
-  m_Enabled: 1
-  m_EditorHideFlags: 0
-  m_Script: {fileID: 11500000, guid: 7ab929cd616fbde4fa9b696c94c91e6e, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
 --- !u!114 &607285373800589754


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/UpdateEventHandler/UpdateDispatcher.cs
```diff
@@ -13,26 +13,44 @@ public class UpdateEventCollection
...
-                eventList.Remove(action);
-                eventHashset.Remove(action);
+                int count = eventList.Count;
+                for (int i = 0; i < count; i++)
+                {
+                    if (eventList[i] == action)
+                    {
+                        eventList.RemoveAt(i);
+                        eventHashset.Remove(action);
+                        break;
```
AjimenezDCL, line 36: We can return here instead of break. It's similar but I think it's more clear.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/UpdateEventHandler/UpdateDispatcher.cs
```diff
@@ -45,10 +63,21 @@ void Awake()
...
-            if ( !eventCollections.ContainsKey(eventType) )
+            bool eventTypeExists = false;
+            foreach (var key in eventCollections.Keys)
+            {
+                if (key == eventType)
+                {
+                    eventTypeExists = true;
+                    break;
+                }
+            }
```
AjimenezDCL, line 76: You can extract a `Contains` method here. It will also allow you to get rid of `eventTypeExsists`.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/UpdateEventHandler/UpdateEventHandler.cs
```diff
@@ -9,10 +9,11 @@ namespace DCL
     public class UpdateEventHandler : IUpdateEventHandler
     {
         private UpdateDispatcher dispatcher;
+        private const string UPDATE_DISPATCHER = "_UpdateDispatcher";
 
         public UpdateEventHandler ()
         {
-            var go = new GameObject("_UpdateDispatcher");
+            var go = new GameObject(UPDATE_DISPATCHER);
```
AjimenezDCL, line 16: We are never using the object name to find stuff.

I think this is only instantiated once, so it wont be a big deal, but we could do something like this:
```suggestion
            var go = new GameObject();
   #if UNITY_EDITOR
            go.name = UPDATE_DISPATCHER;
    #endif
```

AjimenezDCL (approved):
Approving with small suggestions

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/UpdateEventHandler/UpdateDispatcher.cs
0xD-Fabio: done :)

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/UpdateEventHandler/UpdateDispatcher.cs
```diff
@@ -22,17 +22,31 @@ public void Add(Action action)
...
+                        eventHashset.Remove(action);
+                        return;
+                    }
+                }
             }
 
             public bool Contains(Action action)
             {
-                return eventHashset.Contains(action);
+                int count = eventList.Count;
```
popuz, line 39: I would inline this variable into loop definition

popuz (approved):
Good one 💪 
Would be nice to have 2 screenshots of Profiler Hierarchy with deepProfiling for allocations comparison - before and after

PR #4639, opened 2023-03-15
fix: white image stuck in onboarding

### LINKED ISSUE #4629: [Onboarding scene] White screen on start
The onboarding scene is showing a white screen on start for a split of a second. 

Looks like the correct image [is loading late](https://decentralandteam.slack.com/archives/C01QVR7TJK1/p1678708121540229?thread_ts=1678705654.506769&cid=C01QVR7TJK1), and what we see is an empty raw image that is not ready to be shown

dalkia:
## What does this PR change?

Fixes #4629

To white image issue came from enabling a texture before it has been fetched. This PR reorganizes the call

The stuck situation came from changing realms at start, and generated race conditions with scenes not being unloaded properly. This PR takes you directly to the onboarding realm if the flag is on 

Also cleans old loading screen remanant code

## How to test the changes?

1. Go to zone (only place where flag is enabled) in incognito window. 
2. Create a user
3. Teleport out of the onboarding (you can use the genesis plaza button, or complete the onboarding)
4. Get to Genesis Plaza
5. Refresh
6. Check that the onboarding did not start again
7. With another already created account, check that you are not being taken to the onboarding

## Our Code Review Standards

### DIFF
--- a/browser-interface/packages/shared/dao/sagas.ts
+++ b/browser-interface/packages/shared/dao/sagas.ts
@@ -1,4 +1,11 @@
-import { ETHEREUM_NETWORK, PIN_CATALYST, PREVIEW, rootURLPreviewMode } from 'config'
+import {
+  ETHEREUM_NETWORK,
+  HAS_INITIAL_POSITION_MARK,
+  PIN_CATALYST,
+  PREVIEW,
+  RESET_TUTORIAL,
+  rootURLPreviewMode
+} from 'config'
 import { getFromPersistentStorage, saveToPersistentStorage } from 'lib/browser/persistentStorage'
 import defaultLogger from 'lib/logger'
 import { waitFor } from 'lib/redux'
@@ -9,7 +16,7 @@ import { BringDownClientAndReportFatalError } from 'shared/loading/ReportFatalEr
 import {
   getAddedServers,
   getCatalystNodesEndpoint,
-  getDisabledCatalystConfig,
+  getDisabledCatalystConfig, getFeatureFlagVariantName, getFeatureFlagVariantValue,
   getPickRealmsAlgorithmConfig
 } from 'shared/meta/selectors'
 import { SET_REALM_ADAPTER } from 'shared/realm/actions'
@@ -35,6 +42,10 @@ import { createAlgorithm } from './pick-realm-algorithm/index'
 import { getAllCatalystCandidates, getCatalystCandidatesReceived } from './selectors'
 import { Candidate, PingResult, Realm, ServerConnectionStatus } from './types'
 import { ask, ping } from './utils/ping'
+import {store} from "../store/isolatedStore";
+import {trackEvent} from "../analytics/trackEvent";
+import {logger} from "../../entryPoints/logger";
+import {getCurrentUserProfile} from "../profiles/selectors";
 
 const waitForExplorerIdentity = waitFor(getCurrentIdentity, USER_AUTHENTICATED)
 
@@ -142,6 +153,7 @@ function* selectRealm() {
   }
 
   const realm: string | undefined =
+    OnboardingTutorialRealm() ||
     // query param (dao candidates & cached)
     (yield call(qsRealm)) ||
     // preview mode
@@ -153,11 +165,35 @@ function* selectRealm() {
     // cached in local storage
     (yield call(getRealmFromLocalStorage, network))
 
+
   if (!realm) debugger
 
   console.log(`Trying to connect to realm `, realm)
 
   return realm
+}
+
+function OnboardingTutorialRealm(){
+  const profile = getCurrentUserProfile(store.getState())!
+  if(profile){
+    const NEEDS_TUTORIAL = RESET_TUTORIAL || !profile.tutorialStep
+    const NEW_TUTORIAL_FEATURE_FLAG = getFeatureFlagVariantName(store.getState(), 'new_tutorial_variant')
+    const IS_NEW_TUTORIAL_DISABLED =
+      NEW_TUTORIAL_FEATURE_FLAG === 'disabled' || NEW_TUTORIAL_FEATURE_FLAG === 'undefined' || HAS_INITIAL_POSITION_MARK
+    if(NEEDS_TUTORIAL && !IS_NEW_TUTORIAL_DISABLED){
+      try {
+        const realm: string | undefined = getFeatureFlagVariantValue(store.getState(), 'new_tutorial_variant')
+        if (realm) {
+          trackEvent('onboarding_started', { onboardingRealm: realm })
+          return realm
+        } else {
+          logger.warn('No realm was provided for the onboarding experience.')
+        }
+      } catch (err) {
+        console.error(err)
+      }
+    }
+  }
 }
 
 // load realm from local storage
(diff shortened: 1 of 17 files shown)


browser-interface/packages/shared/dao/sagas.ts
```diff
@@ -153,13 +167,44 @@ function* selectRealm() {
...
 }
 
+function OnboardingTutorialRealm(){
+  const profile = getCurrentUserProfile(store.getState())!
+  if(profile){
+    const NEEDS_TUTORIAL = RESET_TUTORIAL || !profile.tutorialStep
+    const NEW_TUTORIAL_FEATURE_FLAG = getFeatureFlagVariantName(store.getState(), 'new_tutorial_variant')
+    const IS_NEW_TUTORIAL_DISABLED =
+      NEW_TUTORIAL_FEATURE_FLAG === 'disabled' || NEW_TUTORIAL_FEATURE_FLAG === 'undefined' || HAS_INITIAL_POSITION_MARK
+      console.log("DEBUG" + profile.tutorialStep);
```
Kinerius, line 185: You probably want to remove this :D

Kinerius (approved):
LGTM!

browser-interface/packages/shared/dao/sagas.ts
```diff
@@ -160,6 +175,35 @@ function* selectRealm() {
   return realm
 }
 
+function* onboardingTutorialRealm() {
+  const profile = yield select(getCurrentUserProfile)
+  if (profile) {
+    const needs_tutorial = RESET_TUTORIAL || !profile.tutorialStep
+    const new_tutorial_feature_flag = getFeatureFlagVariantName(store.getState(), 'new_tutorial_variant')
```
menduz, line 182: Style and conventions:
* use camel case
* use yield select inside sagas
```suggestion
    const newTutorialFeatureFlag = yield select(getFeatureFlagVariantName, 'new_tutorial_variant')
```

dalkia: Correcting. Thanks!

browser-interface/packages/shared/dao/sagas.ts
dalkia: Removed!

browser-interface/packages/shared/dao/sagas.ts
```diff
@@ -160,6 +177,35 @@ function* selectRealm() {
...
+  const profile = yield select(getCurrentUserProfile)
+  if (profile) {
+    const needsTutorial = RESET_TUTORIAL || !profile.tutorialStep
+    const newTutorialFeatureFlag = yield select(getFeatureFlagVariantName, 'new_tutorial_variant')
+    const isNewTutorialDisabled =
+      newTutorialFeatureFlag === 'disabled' || newTutorialFeatureFlag === 'undefined' || HAS_INITIAL_POSITION_MARK
+    console.log('DEBUG' + profile.tutorialStep)
+    if (needsTutorial && !isNewTutorialDisabled) {
+      try {
+        const realm: string | undefined = getFeatureFlagVariantValue(store.getState(), 'new_tutorial_variant')
```
menduz, line 190:
```suggestion
        const realm: string | undefined = yield select(getFeatureFlagVariantValue, 'new_tutorial_variant')
```

browser-interface/packages/shared/dao/sagas.ts
```diff
@@ -216,6 +262,15 @@ function* cacheCatalystRealm() {
   const dao: string = yield select((state) => state.dao)
   const realmAdapter: IRealmAdapter = yield call(waitForRealm)
 
+  if (
+    isOnboarding &&
+    realmAdapter.about.configurations?.realmName !==
+      getFeatureFlagVariantValue(store.getState(), 'new_tutorial_variant')
+  ) {
+    store.dispatch(saveProfileDelta({ tutorialStep: 256 }))
```
menduz, line 270: using store.dispatch inside a saga is forbidden and breaks the code, use `yield put(saveProfileDelta({ tutorialStep: 256 }))` instead
```suggestion
    yield put(saveProfileDelta({ tutorialStep: 256 }))
```

browser-interface/packages/shared/dao/sagas.ts
```diff
@@ -216,6 +262,15 @@ function* cacheCatalystRealm() {
   const dao: string = yield select((state) => state.dao)
   const realmAdapter: IRealmAdapter = yield call(waitForRealm)
 
+  if (
+    isOnboarding &&
+    realmAdapter.about.configurations?.realmName !==
+      getFeatureFlagVariantValue(store.getState(), 'new_tutorial_variant')
```
menduz, line 268: * where does `isOnboarding` comes from?
* can you unwind this if into `if (two && consts)`?

dalkia: isOnboarding is a bool that determines we are doing the onboarding. I needed some way to detect that we were inOboarding when a realmChanged happen, so as to set the onboarding as complete. I could as well have started another async function that checks that, but not completely sure on how to do it.
That would have saved me the isOnboarding var

browser-interface/packages/shared/dao/sagas.ts
```diff
@@ -160,6 +176,34 @@ function* selectRealm() {
...
+function* onboardingTutorialRealm() {
+  const profile = yield select(getCurrentUserProfile)
+  if (profile) {
+    const needsTutorial = RESET_TUTORIAL || !profile.tutorialStep
+    const newTutorialFeatureFlag = yield select(getFeatureFlagVariantName, 'new_tutorial_variant')
+    const isNewTutorialDisabled =
+      newTutorialFeatureFlag === 'disabled' || newTutorialFeatureFlag === 'undefined' || HAS_INITIAL_POSITION_MARK
+    if (needsTutorial && !isNewTutorialDisabled) {
+      try {
+        const realm: string | undefined = yield select(getFeatureFlagVariantValue, 'new_tutorial_variant')
```
eordano, line 188: Isn't this the same as `newTutorialFeatureFlag` of line 183? Avoid doing multiple consecutive select calls if possible.

eordano (commented):
Code looks good, and should work. Tested that it works correctly.

I do thing though that we should strip out the "read catalyst from localstorage" part of the code. It's adding a lot of unnecessary complexity...

browser-interface/packages/shared/dao/sagas.ts
```diff
@@ -35,8 +44,14 @@ import { createAlgorithm } from './pick-realm-algorithm/index'
...
 import { Candidate, PingResult, Realm, ServerConnectionStatus } from './types'
 import { ask, ping } from './utils/ping'
+import { trackEvent } from '../analytics/trackEvent'
+import { logger } from '../../entryPoints/logger'
+import { getCurrentUserProfile } from '../profiles/selectors'
+import { saveProfileDelta } from '../profiles/actions'
 
 const waitForExplorerIdentity = waitFor(getCurrentIdentity, USER_AUTHENTICATED)
+let isCurrentlyInOnboarding = false
+let onboardingRealmName = ""
```
menduz, line 54: prevent using global variables, use selector instead

PR #4648, opened 2023-03-16
feat: spawn point based on teleport position and nearest spawn point

### LINKED ISSUE #4652: [MVFW] Teleporting to any road of the festival teleports you back to the default spawn point
We have the problem of selecting a random spawn point instead of deciding what's the closest one depending on our teleport coordinates

Kinerius:
## What does this PR change?

Closes #4652

When jumping to a scene we used to select a random spawn point defined by the scene.

Now we are selecting the spawn point that is nearest to our jump coordinates.
This means that if we have a scene that is 4x4 parcels and has a spawn point at each corner, the teleport positions will be:

| B 	| B 	| C 	| C 	|
|---:	|:---:	|:---:	|---	|
| B 	| B 	| C 	| C 	|
| A 	| A 	| D 	| D 	|
| A 	| A 	| D 	| D 	|

Being:
A: Lower-left
B: Top-Left
C: Top-Right
D: Lower-Right

## How to test the changes?

- Go to [MVFW Road scenes](https://play.decentraland.org/?explorer-branch=feat%2Fspawn-points&realm=baldr&position=-73%2C6)
- Open the map and click near the corners (that are inside the green box), if that parcel is a road, you will get teleported to the nearest spawn point (at the time I'm writing this, the scene has 4 spawn points, one at each corner )

## Our Code Review Standards

### DIFF
--- a/browser-interface/packages/shared/world/positionThings.ts
+++ b/browser-interface/packages/shared/world/positionThings.ts
@@ -153,15 +153,10 @@ function pickSpawnpoint(land: Scene, targetParcelPosition: Vector3, basePosition
 
   // 5 - If the final position is outside the scene limits, we zero it
   if (!DEBUG) {
-    const sceneBaseParcelCoords = land.scene.base.split(',')
-    const sceneBaseParcelWorldPos = gridToWorld(
-      parseInt(sceneBaseParcelCoords[0], 10),
-      parseInt(sceneBaseParcelCoords[1], 10)
-    )
     const finalWorldPosition = {
-      x: sceneBaseParcelWorldPos.x + finalPosition.x,
+      x: basePosition.x + finalPosition.x,
       y: finalPosition.y,
-      z: sceneBaseParcelWorldPos.z + finalPosition.z
+      z: basePosition.z + finalPosition.z
     }
 
     if (!isWorldPositionInsideParcels(land.scene.parcels, finalWorldPosition)) {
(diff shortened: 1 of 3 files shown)


browser-interface/packages/shared/world/positionThings.ts
```diff
@@ -185,3 +199,17 @@ function computeComponentValue(x: number | number[]) {
...
+{
+  const x = spawnPoint.position.x;
+  const y = spawnPoint.position.y;
+  const z = spawnPoint.position.z;
+
+  if (typeof x === "number" && typeof y === "number" && typeof z === "number")
+  {
+    return new Vector3(x, y ,z); 
+  } else {
+    return new Vector3(x[0], y[0], z[0]);
```
pbosio, line 213: why other indexes aren't being considered?
spawnpoint A : {x: [0,9], y: [0], z: [0,10]}
spawnpoint B : {x: 10, y: 0, z: 0}
if you want to teleport to {x: 6, y:0, z: 0}
spawnpoint B will be set as the spawnpoint but it should be spawnpoint A

Kinerius: In your example, you will be teleported at spawn point A since the parcel is 0,0 the position to teleport at is 0.0, parcels are 16x16 and you can't teleport to a specific position.

What we can change is to teleport to the middle of the parcel and use the center of the area spawn point to compare their position.

kwsskulinski (approved):
Issue is fixed no issues were found. Player teleports to the nearest spawn point,.

aixaCode (approved):

PR #4649, opened 2023-03-16
chore: increase scene timeout to 120

dalkia:
## What does this PR change?

Increases scene timeout to 2 minutes. This is a requirement for MVFW 2023, since there are a couple of scenes not loadings. The current scene shows whatever is in the scene once the timeout is reached, and this is a problem, since many times it shows uncompleted assets or even an empty scene alltogether. The increased timeout will give more time to users with slower connection.

The increasing or decreasing of this number is flawed, and its currently just a mitigation. A new PoC is in the works which would allow proper handling of download times and its corresponding realtime feedback. Also, it will include a user feddback and decision system, in which the user will be able to take action when the scene is not loading; like keep waiting or go where he came from.

More info on the [PoC](https://github.com/decentraland/unity-renderer/issues/4665) here

## Our Code Review Standards

### DIFF
--- a/browser-interface/packages/shared/world/SceneWorker.ts
+++ b/browser-interface/packages/shared/world/SceneWorker.ts
@@ -324,7 +324,7 @@ export class SceneWorker {
 
     sceneEvents.emit(SCENE_LOAD, signalSceneLoad(this.loadableScene))
 
-    const WORKER_TIMEOUT = 180_000
+    const WORKER_TIMEOUT = 120_000
     setTimeout(() => this.onLoadTimeout(), WORKER_TIMEOUT)
   }
 
--- a/browser-interface/packages/shared/world/SceneWorker.ts
+++ b/browser-interface/packages/shared/world/SceneWorker.ts
@@ -324,7 +324,7 @@ export class SceneWorker {
 
     sceneEvents.emit(SCENE_LOAD, signalSceneLoad(this.loadableScene))
 
-    const WORKER_TIMEOUT = 90_000 // ninety seconds to mars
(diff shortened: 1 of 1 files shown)


Kinerius (approved):

pbosio (approved):

browser-interface/packages/shared/world/SceneWorker.ts
```diff
@@ -329,7 +329,7 @@ export class SceneWorker {
 
     sceneEvents.emit(SCENE_LOAD, signalSceneLoad(this.loadableScene))
 
-    const WORKER_TIMEOUT = 90_000 // ninety seconds to mars
+    const WORKER_TIMEOUT = 120_000
```
menduz, line 332: Why is this needed?

dalkia: Further information [here](https://decentralandteam.slack.com/archives/C0292P4HJ93/p1678985200846279)

pbosio: @menduz I've heard from back channels that this change is needed for MFW and there is not enough time for a proper fix

menduz: Never change parameters like that without adding a comment explaining why it was changed, pleeease 🙏

Nothing prevents anyone to revert it and it only makes the codebase more fragile

// timeout is 120 because gltf are big and network is slow

Even that example would be helpful... Right now nobody knows why it was changed, what it fixes and what is the root cause of the problem.

Issue #4489, opened 2023-03-01
Spike: Investigate what could cause users to switch servers


m3taphysics:
### Description 
We have had reports of users switching servers (usually heimdallr) in the middle of a game. They are sometimes moved to a slightly different location, but something is causing the realm switch.

Refer to: [Slack link](https://decentralandteam.slack.com/archives/CS8SCNYUV/p1675980295082479)

### A/C's
- [ ] Investigate and understand what could cause a realm switch mid experience

PR #4671, opened 2023-03-17
fix: add tracking for error events

marianogoldman:
## What does this PR change?
Adds event tracking in 2 particular situations:
* When there is a comms error trying to send data to LiveKit.
* When receiving a disconnect event from the server.

This would capture the events and/or exceptions and allow us to determine why disconnections happen. This might (or might not) be related to https://github.com/decentraland/unity-renderer/issues/4489.

### DIFF
--- a/browser-interface/packages/shared/comms/adapters/LivekitAdapter.ts
+++ b/browser-interface/packages/shared/comms/adapters/LivekitAdapter.ts
@@ -41,7 +41,12 @@ export class LivekitAdapter implements MinimumCommunicationsAdapter {
         this.config.logger.log('remote participant left', _.identity)
       })
       .on(RoomEvent.Disconnected, (_reason: DisconnectReason | undefined) => {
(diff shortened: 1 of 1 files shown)


pentreathm (approved):

browser-interface/packages/shared/comms/adapters/LivekitAdapter.ts
```diff
@@ -41,7 +41,12 @@ export class LivekitAdapter implements MinimumCommunicationsAdapter {
         this.config.logger.log('remote participant left', _.identity)
       })
       .on(RoomEvent.Disconnected, (_reason: DisconnectReason | undefined) => {
-        this.config.logger.log('disconnected from room')
+        this.config.logger.log('disconnected from room', _reason)
+        trackEvent('error', {
```
marianogoldman, line 45: @menduz WDYT about this one? It might be a lot of data to track...

menduz (approved):

PR #4684, opened 2023-03-20
fix: error loading wearables for guest users in worlds

sandrade-dcl:
## What does this PR change?
Url param `DISABLE_force_to_request_wearables_through_kernel]` causing issues with wearables when entering world `menduz.dcl.eth` directly on startup.

## How to test the changes?
Clear cache, log in with "DISABLE_force_to_request_wearables_through_kernel" and "realm=menduz.dcl.eth" params. Guest arrives at the initial character creation screen with "Problem loading your wearables" error and a blank avatar. Wallet: (After a lengthy loading) logs in with "loading scene timeout" error message and a blue hologram-looking avatar missing its wearables.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/DCLServices/Lambdas/LambdasService.cs
+++ b/unity-renderer/Assets/DCLServices/Lambdas/LambdasService.cs
@@ -74,6 +74,7 @@ namespace DCLServices.Lambdas
             var urlBuilder = GenericPool<StringBuilder>.Get();
             urlBuilder.Clear();
             urlBuilder.Append(GetLambdasUrl());
+            urlBuilder.Append('/');
 
             var endPointSpan = endPoint.AsSpan();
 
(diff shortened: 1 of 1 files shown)


unity-renderer/Assets/DCLServices/Lambdas/LambdasService.cs
```diff
@@ -74,6 +74,7 @@ internal string GetUrl(string endPoint, params (string paramName, string paramVa
             var urlBuilder = GenericPool<StringBuilder>.Get();
             urlBuilder.Clear();
             urlBuilder.Append(GetLambdasUrl());
+            urlBuilder.Append('/');
```
menduz, line 77: Please be defensive with this one, since many configurations may occur in different realms/worlds

```suggestion
            if(!urlBuilder.endsWith('/')) urlBuilder.Append('/');
```

menduz (changes requested):

unity-renderer/Assets/DCLServices/Lambdas/LambdasService.cs
sandrade-dcl: I have checked that we could put always this final `/` as it would be the same to request using this url:
`https://peer.decentraland.org/lambdas/users/...`
or this one:
`https://peer.decentraland.org/lambdas/users//...`

Anyway, if you are not convinced, I can do the check in the code.

sandrade-dcl: Done! https://github.com/decentraland/unity-renderer/pull/4684/commits/36ceb55a063072ddb02c6d4893cc68ab42639b4e

PR #4691, opened 2023-03-21
fix: sdk7 video playback

pravusjif:
* Added error logging when the media player can't play a video due to unallowed host names (also affects Audio and Textures, as those use the whole 'allowed hosts names' strategy as well)
* Updated `VideoPlayerHandler` playing conditions handling to take into account: valid video URL, rendering active state and user interaction with application (without user interaction, browsers don't start the video and it never starts until reloading/hot-reloading the scene)
* Fixed issue with 'allowedMediaHostnames' using upper case (compared to always lower-case host name of target media assets)
* Added several tests

Issue: https://github.com/decentraland/sdk/issues/653


pbosio (commented):
Why are we bypassing scene's external hosts permissions here?

pravusjif:
> > > Why are we bypassing scene's external hosts permissions here?
> > 
> > 
> > So maybe the report of the video not playing was wrong and it was because they were not using the permissions system? let me take a look [at the test scene I've been using](https://github.com/decentraland-scenes/sdk7-goerli-plaza/tree/main/video-streaming)
> 
> Looks like that scene does not have the required permissions
> 
> ```
>   "requiredPermissions": [
>     "ALLOW_MEDIA_HOSTNAMES"
>   ],
>   "allowedMediaHostnames": [
>     "somehost.com"
>   ] 
> ```

Yes, I just [corrected that in its repo](https://github.com/decentraland-scenes/sdk7-goerli-plaza/commit/281dc4c5c1e33455e3d479670984cbc2ae37ae5d), so video streaming was actually working.

I'll revert the changes in this branch but I'll try to tackle another inconsistency that I saw: videos normally only start playing after 1 hot reload, not on the first scene load... I'll use this same branch to tackle that then.

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/VideoPlayer/VideoPlayerHandler.cs
```diff
@@ -14,68 +14,120 @@ public class VideoPlayerHandler : IECSComponentHandler<PBVideoPlayer>
...
+            if (!scene.isPersistent)
+                CommonScriptableObjects.sceneNumber.OnChange += OnSceneChanged;
+            CommonScriptableObjects.rendererState.OnChange += OnRendererStateChanged;
+        }
 
         public void OnComponentRemoved(IParcelScene scene, IDCLEntity entity)
         {
+            Helpers.Utils.OnCursorLockChanged -= OnCursorLockChanged;
+            CommonScriptableObjects.sceneNumber.OnChange -= OnSceneChanged;
+            CommonScriptableObjects.rendererState.OnChange -= OnRendererStateChanged;
```
pbosio, line 56: I don't think this `CommonScriptableObjects.rendererState` is working as it used to, so probably it must be changed for the loading screen event

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/VideoPlayer/VideoPlayerHandler.cs
```diff
@@ -14,68 +14,120 @@ public class VideoPlayerHandler : IECSComponentHandler<PBVideoPlayer>
...
+
+            ConditionsToPlayVideoChanged();
+        }
+
+        private void ConditionsToPlayVideoChanged()
+        {
+            if (lastModel == null) return;
+
+            bool shouldBePlaying = lastModel.IsPlaying() && canVideoBePlayed;
+            if (shouldBePlaying != videoPlayer.playing)
```
pbosio, line 103: do we want the "video not playing if in another scene" behavior? the video is already reducing the frequency it updates if user is not in the same scene

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/VideoPlayer/VideoPlayerHandler.cs
pravusjif: Done

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/VideoPlayer/VideoPlayerHandler.cs
pravusjif: Removed

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/VideoPlayer/VideoPlayerHandler.cs
pravusjif: aparently `CommonScriptableObjects.isLoadingHUDOpen` is not being used anymore, I have to find an alternative...

pravusjif: ok ended up using `DataStore_LoadingScreen`

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/AudioStream/ECSAudioStreamComponentHandler.cs
```diff
@@ -23,17 +23,14 @@ public class ECSAudioStreamComponentHandler : IECSComponentHandler<PBAudioStream
         // Flags to check if we can activate the AudioStream
         internal bool isInsideScene = false;
         internal bool isRendererActive = false;
-        internal bool wasCursorLocked = false;
+        internal bool hadUserInteraction = false;
```
pbosio, line 26: `hadUserInteraction` could mean a lot of stuffs, but I'm not sure what could be a proper name for that field 🤷‍♂️

pbosio (approved):

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/AudioStream/ECSAudioStreamComponentHandler.cs
pravusjif: haha yeah but for me that's the most clear name for what's needed for the freaking video to start playing in the browser... user interaction.

AjimenezDCL (approved):

Issue #4628, opened 2023-03-14
User names are being constantly changed


Kinerius:
## Context

We have some reports of users constantly changing their names, we have to check if it's really a bug or an exploit that we need to mitigate its impact or usability.

https://user-images.githubusercontent.com/7646450/225112907-848765a5-d311-4d07-bbdc-cc61a4bb1935.mp4

https://user-images.githubusercontent.com/7646450/225113052-f1c64d22-119d-4665-9a8b-98ec40d945ae.mp4

pentreathm:
The issue could be related to Profile Entities having a name defined as an empty string and the client trying to constantly generate a Fake Name for the profile. 
Here are two profiles that were having this issue: 

`curl --location --request POST 'https://peer-ap1.decentraland.org/content/entities/active' \
--header 'Content-Type: application/json' \
--data-raw '{ "pointers": ["0x1e74fc6e49380b7372385061d5fdddcf01d26927"] }'`

`curl --location --request POST 'https://peer-ap1.decentraland.org/content/entities/active' \
--header 'Content-Type: application/json' \
--data-raw '{ "pointers": ["0x74538540a034b4e786cd9ceb8d675ae58d6e840b"] }'`.

kwssbocian:
Reopening the ticket as the bug is again present on prod. Only a small amount of remote users appears to be affected as the issue was not reproducible from a single machine.
https://github.com/decentraland/unity-renderer/assets/110463715/b544b1c4-0d37-4347-a18b-ab979cb142b8
https://github.com/decentraland/unity-renderer/assets/110463715/5ea18412-34c9-43e1-b1c6-df903e2f5996

PR #4694, opened 2023-03-21
fix: dont update avatar if no change was found

eordano:
Addresses https://github.com/decentraland/unity-renderer/issues/4628


popuz:
cleanup: re-open if needed

PR #4697, opened 2023-03-21
hotfix: delay and return value issues in fetch peer profile

moliva:
## What does this PR change?

Function `fetchPeerProfile` was wrongly setting a timer with redux-sagas delay which generated no delay and returned a static `CallEffect` object meant to be used by the `race` generator from that same framework, in a vanilla promise race this was interpreted as a return value and then an invalid avatar.

This in turn was generating a random profile every time the profile view was opened and thus a new name was picked for the same address every time.

This change introduces:
- `sleep` function instead of `delay` to actually create that delay inside a `Promise#race`
- Ensures that the result of the `Promise#race `is actually the avatar from the store or a `null` value from the sleep.

## Our Code Review Standards

### DIFF
--- a/browser-interface/packages/shared/profiles/sagas/comms/index.ts
+++ b/browser-interface/packages/shared/profiles/sagas/comms/index.ts
@@ -23,7 +23,7 @@ export async function fetchPeerProfile(
         return state.profiles.userInfo[userId]!.data
       }
     }),
-    sleep(COMMS_PROFILE_TIMEOUT).then(() => null)
+    sleep(COMMS_PROFILE_TIMEOUT)
   ])
   incrementCounter(result ? 'profile-over-comms-succesful' : 'profile-over-comms-failed')
   if (!result) {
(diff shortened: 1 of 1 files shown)


browser-interface/packages/shared/profiles/sagas/comms/index.ts
```diff
@@ -23,11 +23,12 @@ export async function fetchPeerProfile(
         return state.profiles.userInfo[userId]!.data
       }
     }),
-    delay(COMMS_PROFILE_TIMEOUT)
+    sleep(COMMS_PROFILE_TIMEOUT).then(() => null)
```
eordano, line 26:
```suggestion
    sleep(COMMS_PROFILE_TIMEOUT)
```
No need to return it!

eordano (approved):

Julieta11 (approved):

lorux0 (approved):

Julieta11:
Closing this PR because of CI rules, the fix has been moved to this [PR](https://github.com/decentraland/unity-renderer/pull/4702) with a renamed branch to `hotfix/`

PR #4698, opened 2023-03-21
fix: transparent objects and SSAO

AjimenezDCL:
Transparent objects should ignore SSAO. The main reason SSAO was bleeding into transparent objects is because the SSAO pass is performed after the opaque objects and before the transparent ones. This cannot be changed, and therefore the postprocessing layer was painting the AO calculated behind some transparent objects.

To fix this, I modified the shared dependency at https://github.com/decentraland/unity-shared-dependencies/pull/18.

This PR above must be merged before this one, which updates the package version.

## Test
Go to -109 -18. The "clock-man" face should no longer bleed AO.
image

Take a look at the rest of the world and look for regressions in SSAO.


Kinerius (approved):

kwssbocian (approved):
✔️Shadow related issue on the clock's face is gone. No other issues observed.

PR #4704, opened 2023-03-21
hotfix: race comms and catalyst fetching for profile with version

moliva:
## What does this PR change?

To mitigate the current timeouts when fetching profiles from comms peers and trying to optimize the use of the catalysts, this PR introduces a race between the both while also checking the version of the profile returned for the catalyst branch.

## How to test the changes?

1. Launch the explorer
2. Go to the channels tab and enter one
3. Click on the menu for a given user and show profile
4. The profile should be taking considerable less time to load (the state of things before this PR would be >4.5s for a set timeout and retry strategy nonworking as expected for the use case)

## Our Code Review Standards


Julieta11 (commented):
I tested the branch and it seems there is an improvement related to the delay for fetching the profile of a channel member. I still see the user as a guest at first, but it refreshes with the correct data faster

menduz (approved):

Issue #4539, opened 2023-03-06
Performance analysis for new passport


davidejensen:
Do some performance analysis and profiling of the new passport in order to find key areas to improve

davidejensen:
Additional minor fix, the pool where reset at every opening of the passport due to a misleading function name, fixed

PR #4705, opened 2023-03-22
chore: passport performance improvements

davidejensen:
## What does this PR change?

Improves passport performances in two key areas described in issue #4539 

...

## How to test the changes?

1. Launch the explorer https://play.decentraland.org/?explorer-branch=chore/passport-improvements
2. Use the passport and check it works as expected

## Our Code Review Standards


popuz (approved):
- I see several changes applies as overrides on the nested prefabs. Would it make sense to apply changes directly to these prefabs and don't have overrides there?
- Also, I would suggest extract some internal parts of `PlayerPassport.prefab` to prefabs as well and apply changes on them instead of PlayerPassport. It will be not about re-usability of them, but more about isolation of changes on the PRs. Then it would be easier to track, since readability of prefab changes on the version control system is hard. As well, such strategy proves to mitigate well merge conflicts on prefabs. (PassportNavigation and PlayerInfo are good candidates for such extraction)

davidejensen:
> * I see several changes applies as overrides on the nested prefabs. Would it make sense to apply changes directly to these prefabs and don't have overrides there?
> * Also, I would suggest extract some internal parts of `PlayerPassport.prefab` to prefabs as well and apply changes on them instead of PlayerPassport. It will be not about re-usability of them, but more about isolation of changes on the PRs. Then it would be easier to track, since readability of prefab changes on the version control system is hard. As well, such strategy proves to mitigate well merge conflicts on prefabs. (PassportNavigation and PlayerInfo are good candidates for such extraction)

* not really because the changes apply in that context and not in the generic prefab
* yeah I understand what you mean, in the next iteration I might as well split it in some minor prefabs, ty for the suggestion

AjimenezDCL (approved):

PR #4713, opened 2023-03-22
hotfix: nearby chat sequence sorting

lorux0:
## What does this PR change?

Avoid sorting by timestamp the messages sent to `~nearby` channel. Instead, add them sequentaly.
Comms does not calculate any server timestamp, so depending on the user's local timestamp, messages could be in the wrong order.

## How to test the changes?

1. Launch the explorer
2. Send/receive private messages. Load previous history. Messages should be correctly sorted in the list.
3. Send/receive messages in ~nearby. Messages should be correctly sorted in the list.
4. Send/receive messages in #channels. Load previous history. Messages should be correctly sorted in the list.

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDController.cs
@@ -24,10 +24,6 @@ public class ChatHUDController : IHUD
 
     public delegate UniTask<List<UserProfile>> GetSuggestedUserProfiles(string name, int maxCount, CancellationToken cancellationToken);
 
-    public event Action OnInputFieldSelected;
-    public event Action<ChatMessage> OnSendMessage;
-    public event Action<ChatMessage> OnMessageSentBlockedBySpam;
-
     private readonly DataStore dataStore;
     private readonly IUserProfileBridge userProfileBridge;
     private readonly bool detectWhisper;
@@ -46,8 +42,26 @@ public class ChatHUDController : IHUD
     private int mentionLength;
     private int mentionFromIndex;
     private Dictionary<string, UserProfile> mentionSuggestedProfiles;
+    private Comparison<ChatEntryModel> sortingStrategy;
 
     private bool isMentionsEnabled => dataStore.featureFlags.flags.Get().IsFeatureEnabled("chat_mentions_enabled");
+
+    public Comparison<ChatEntryModel> SortingStrategy
+    {
+        get => sortingStrategy;
+
+        set
+        {
+            sortingStrategy = value;
+
+            if (view != null)
+                view.SortingStrategy = value;
+        }
+    }
+
+    public event Action OnInputFieldSelected;
+    public event Action<ChatMessage> OnSendMessage;
+    public event Action<ChatMessage> OnMessageSentBlockedBySpam;
 
     public ChatHUDController(DataStore dataStore,
         IUserProfileBridge userProfileBridge,
@@ -62,6 +76,7 @@ public class ChatHUDController : IHUD
         this.getSuggestedUserProfiles = getSuggestedUserProfiles;
         this.socialAnalytics = socialAnalytics;
         this.profanityFilter = profanityFilter;
+        SortingStrategy = (a, b) => a.timestamp.CompareTo(b.timestamp);
     }
 
     public void Initialize(IChatHUDComponentView view)
@@ -85,6 +100,7 @@ public class ChatHUDController : IHUD
         this.view.OnMentionSuggestionSelected += HandleMentionSuggestionSelected;
         this.view.OnOpenedContextMenu -= OpenedContextMenu;
         this.view.OnOpenedContextMenu += OpenedContextMenu;
+        this.view.SortingStrategy = SortingStrategy;
     }
 
     private void OpenedContextMenu()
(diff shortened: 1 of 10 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDView.cs
```diff
@@ -452,7 +453,8 @@ private void SortEntriesImmediate()
         {
             if (this.entries.Count <= 0) return;
 
-            var entries = this.entries.Values.OrderBy(x => x.Model.timestamp).ToList();
+            var entries = this.entries.Values.ToList();
+            entries.Sort((a, b) => SortingStrategy.Invoke(a.Model, b.Model));
```
eordano, line 457: Won't this cause allocations?

lorux0: I dont think there is any new allocations that we already had. We should try to avoid `ToList()` but its needed for the `SiblingIndex` later. Perhaps there is a way to avoid it. Im not super worried because the sorting will be triggered one time per frame at MOST (which is not likely to happen), and we have a max of 30 entries.

lorux0: Just uploaded these changes, i like it

eordano: Cool. Doesn't the lambda closure creates an allocation? I mean the `() => {}` syntax

kwsskulinski (approved):
Messages are correctly displayed.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDView.cs
```diff
@@ -452,13 +453,8 @@ private void SortEntriesImmediate()
...
 
-            var entries = this.entries.Values.OrderBy(x => x.Model.timestamp).ToList();
-
-            for (var i = 0; i < entries.Count; i++)
-            {
-                if (entries[i].transform.GetSiblingIndex() != i)
-                    entries[i].transform.SetSiblingIndex(i);
-            }
+            foreach (var entry in entries.Values.OrderBy(obj => obj.Model, SortingStrategy))
+                entry.transform.SetAsLastSibling();
```
AjimenezDCL, line 457: OrderBy allocates lot of memory as well because it duplicates the collection.

The core issue that prevents this code from being optimal is that `entries` is a Dictionary. Do we need accessing by `messageId`? Because then we could have a List and apply `List.Sort` which is super efficient. https://stackoverflow.com/a/53385164

Another solution is to use a SortedDictionary https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.sorteddictionary-2?view=net-7.0. It allocates more memory than the SortedList but both would be better than the current implementation.

AjimenezDCL (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDController.cs
```diff
@@ -46,9 +43,27 @@ public class ChatHUDController : IHUD
...
+    {
+        get => sortingStrategy;
+
+        set
+        {
+            sortingStrategy = value;
+
+            if (view != null)
+                view.SortingStrategy = value;
+        }
```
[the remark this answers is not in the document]
lorux0: The controller doesn't pass a list of messages. They comes one by one. If the controller would need to pass all these information would require more iterations over the same list

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDView.cs
lorux0: We need to index them by id because sometimes the message arrives duplicated. Depending if its already added we do an update or an addition to the list. I will try with the `SortedDictionary` and let you know the results

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDView.cs
lorux0: Not the lambda. But `OrderBy` creates allocations

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDView.cs
lorux0: I tried with a `SortedDictionary` but it only allows to sort by key (string), which is not useful for this case, because the sorting depends on the message information, not the id

AjimenezDCL: yeah, let's try to figure this out later.

AjimenezDCL: If your only use case for the dictionary is to avoid duplicating entries, you can always use an additional Hashset to lookup. Anyway, approving.

lorux0: We need to update the message information if its duplicated. `HashSet` will throw an exception when adding a duplicate. However, its better than a `List`

PR #4718, opened 2023-03-22
fix: sdk7 playground

pbosio:
lack of parcels array in global scenes was triggering a null exception on the scene boundaries checker while running playground


unity-renderer/Assets/DCLPlugins/ECS7/Systems/SceneBoundsCheckerSystem/ECSSceneBoundsCheckerSystem.cs
```diff
@@ -54,6 +54,9 @@ public void Dispose()
 
         private void OnSceneAdded(IParcelScene scene)
         {
+            if (scene.isPersistent)
```
menduz, line 57:
```suggestion
            // global scenes do not have boundaries to check
            if (scene.isPersistent)
```

menduz (approved):

PR #4719, opened 2023-03-22
hotfix: people in ground mitigation

dalkia:
## What does this PR change?

This mitigates two possible scenarios:

1) Wrong spawn point set in Y (different than 0)
2) Wrong Y position sent from client

This will not fix the issue in Genesis Plaza, since the spawn point there is not in the ground. But it is a much more less risky mitigation suitable for MVFW (comparing it to )

## How to test the changes?

1. Launch the explorer
2. ...

## Our Code Review Standards


kwsskulinski (approved):
Second client does not load below the ground.

popuz (approved):

kwssbocian (approved):
✔️At -73,-9 other avatars appear correctly at the ground level when spawning in.

PR #4720, opened 2023-03-22
hotfix: making comms loop for fetching profiles concurrently using spawn

moliva:
## What does this PR change?

With the motivation of loading near profiles quicker, this PR changes the call to fetch peer profile in the comms loop (whenever a profile is announced around the player) to make the call non-blocking, changing the `call` from redux-sagas framework to a `spawn` one.

This change makes the call to `fetchPeerProfile` from a blocking one having to wait for the answer/timeout to a non-blocking one which should still be consistent writing the responses to the common store.

## How to test the changes?

1. Launch the explorer
2. Go to a scene with many other players present (e.g. the poker scene or Genesis plaza starting point)
3. See how avatars should appear quicker than in the current released version (at least the ghost of them should appear much faster)

## Our Code Review Standards


menduz (approved):
Brilliant

eordano (approved):

dalkia (approved):

PR #4759, opened 2023-03-27
fix: gpuSkinningThrottlerService.Register() called twice

popuz:
## What does this PR change?

Avatar was registering its `gpuSkinning `each time when calling `Load() `method. This cause the problem, since gpuSkinning instance is readonly assigned on Avatar creation, but `Load `called often - for example by `CharacterPreviewController` when changing wearables from backpack.

- now, when GPUSkinning is re-register, we just modify its timeBetweenFrames
- removed duplication in Avatar and AvatarWithHologramm classes (via inheritance and variation of Template Method pattern)

[Devs Only]
1. Launch the explorer
2. Change wearables in Backpack
3. [DevsOnly]: Observe no LogWarning in console from `gpuSkinningThrottlerService`
4. Check that Avatar and Hologrammed avatar behaves correctly

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Rendering/Utils/GPUSkinning/GPUSkinningThrottlerService.cs
+++ b/unity-renderer/Assets/Rendering/Utils/GPUSkinning/GPUSkinningThrottlerService.cs
@@ -10,6 +10,16 @@ public class GPUSkinningThrottlerService : IGPUSkinningThrottlerService
 
     private CancellationTokenSource cts;
 
+    public static GPUSkinningThrottlerService Create(bool initializeOnSpawn)
+    {
+        GPUSkinningThrottlerService service = new GPUSkinningThrottlerService();
+
+        if (initializeOnSpawn)
+            service.Initialize();
+
+        return service;
+    }
+
     public void Initialize()
     {
         cts = new CancellationTokenSource();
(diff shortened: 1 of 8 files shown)


unity-renderer/Assets/Rendering/Utils/GPUSkinning/GPUSkinningThrottlerService.cs
```diff
@@ -21,7 +21,7 @@ public void Register(IGPUSkinning gpuSkinning, int framesBetweenUpdates = 1)
         if (!gpuSkinnings.ContainsKey(gpuSkinning))
             gpuSkinnings.Add(gpuSkinning, framesBetweenUpdates);
         else
-            Debug.LogWarning("GPUSkinningThrottlerService: Register called twice for the same IGPUSkinning");
+            Debug.LogWarning($"GPUSkinningThrottlerService: Register called twice for the same IGPUSkinning {gpuSkinning}", gpuSkinning.renderer.gameObject);
```
AjimenezDCL, line 24: I'd remove this warning and instead just modify the current throttling.

AjimenezDCL (changes requested):

AjimenezDCL (changes requested):

AjimenezDCL (approved):

dalkia:
Really like the refactor used in `Avatar` and `AvatarWithHologram`! Nice use of the Template method. As much we can save on those convuleted class, the better.

`gpuSkinningThrottlerService` is not being called twice, but the PR introduced a visual regression. Other players are in T position before appearing, instead of idle

PR
Screen Shot 2023-03-30 at 18 43 42

PRD
Screen Shot 2023-03-30 at 18 45 09

dalkia (approved):

kwsskulinski (approved):
No issues have been found.

PR #4760, opened 2023-03-27
hotfix: address instead of username in chat messages

lorux0:
## What does this PR change?

Fetches the profile, if not available, when adding a chat message.
Fixes that sometimes you see the address of the user instead of the name in the message, either sender or recipient. 

## How to test the changes?

1. Launch the explorer
2. Send and receive messages in ~nearby
3. Send and receive messages in private conversations
4. Send and receive messages in #channels (recommended to use #test)
5. If at any point you see an address instead of the user name, it should change to a proper name after a small period of time

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDController.cs
@@ -165,6 +165,7 @@ public class ChatHUDController : IHUD
             else
             {
                 model.recipientName = GetEllipsisFormat(message.recipient);
+                model.isLoadingNames = true;
 
                 // sometimes there is no cached profile, so we request it
                 // dont block the operation of showing the message immediately
@@ -186,6 +187,7 @@ public class ChatHUDController : IHUD
             else
             {
                 model.senderName = GetEllipsisFormat(message.sender);
+                model.isLoadingNames = true;
 
                 // sometimes there is no cached profile, so we request it
                 // dont block the operation of showing the message immediately
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDController.cs
@@ -1,5 +1,4 @@
 using Cysharp.Threading.Tasks;
-using DCL;
 using DCL.Chat;
 using DCL.Chat.HUD;
 using DCL.Interface;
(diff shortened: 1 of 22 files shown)


sandrade-dcl (approved):
LGTM, although I see that the code flow of `EnsureProfileThenUpdateMessage` and its retries approach is a bit complex and could be also a bit delicated.

Is it really needed to implement a retries system here?, why could be needed to call `RequestFullUserProfileAsync` several times?, because of possible connection issues for example?.

We're going to have to do a deep test of this feature 👍

lorux0:
> LGTM, although I see that the code flow of `EnsureProfileThenUpdateMessage` and its retries approach is a bit complex and could be also a bit delicated.
> 
> Is it really needed to implement a retries system here?, why could be needed to call `RequestFullUserProfileAsync` several times?, because of possible connection issues for example?.
> 
> We're going to have to do a deep test of this feature 👍

I added the retries system because there is a recursive call to `AddChatMessage` in `EnsureProfileThenUpdateMessage` which may freeze the app if not handled correctly

sandrade-dcl (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDController.cs
```diff
@@ -109,29 +118,131 @@ public void SetVisibility(bool visible)
...
+            {
+                Debug.LogWarning($"Profile fetching max retries limit reached for {profileId}");
+                return;
+            }
+
+            try
+            {
+                UserProfile requestedProfile = await userProfileBridge.RequestFullUserProfileAsync(profileId, cancellationToken);
+
+                if (requestedProfile != null)
```
mikhail-dcl, line 139: Just a highlight: according to the logic underneath `userProfile` can never be `null`

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/ChatHUDController.cs
```diff
@@ -109,29 +118,131 @@ public void SetVisibility(bool visible)
...
+                : ChatEntryModel.SubType.RECEIVED;
+        }
+        else if (message.messageType == ChatMessage.Type.PUBLIC)
+        {
+            model.subType = message.sender == ownProfile.userId
+                ? ChatEntryModel.SubType.SENT
+                : ChatEntryModel.SubType.RECEIVED;
+        }
+
+        AddChatMessage(model, setScrollPositionToBottom, spamFiltering, limitMaxEntries,
```
mikhail-dcl, line 222: This recursive flow is hard to follow. I would like to clarify how it should work:
1. It is called `AddChatMessage` but when called multiple times it acts as `update`?
2. Why do we call right the same method from `EnsureProfileThenUpdateMessage` when we already have `model` created and the only thing we need to do is update `recipientName` ? It leads to additional allocations and logical misintentions.

mikhail-dcl (commented):
I am not going to block the hotfix but I think the code could be written in a neater way

lorux0:
Base branch changed to `dev`. Check this pr: https://github.com/decentraland/unity-renderer/pull/4800

PR #4761, opened 2023-03-28
fix: raycast protocol update

pravusjif:
* Changed timestamp usage going back to having the same raycast's timestamp in the raycast result
* Implemented support for 3 new direction type (local direction, global direction, target entity, global target)
* Moved raycasting logic to a system instead of the RaycastHandler, to ensure:
    * Rays are casted after entity transforms resolving (using LateUpdate)
    * Being able to cope with the new `continuous` property that casts the ray on every frame regardless of having a model update
 * Created new 'SDKCustomLayer' in unity renderer and implemented sdk raycast component`collisionMask` property parsing into renderer layers (any `CL_CUSTOMX` => adds 'SDKCustomLayer' to the mask; `CL_POINTER` => 'OnPointerEvent'; `CL_PHYSICS` => 'CharacterController'; `CL_POINTER | CL_PHYSICS` => 'Default')
 * Updated `MeshCollider` collision layer handling in the renderer to take into account new `CL_CUSTOMX` layers (using 'SDKCustomLayer' in the renderer)
 * Updated raycast in the renderer to always use `RaycastAll` because of the case of different `CL_CUSTOM` layers all using the same 'SDKCustomLayer' in the renderer (and already existent CL_PHYSICS + CL_POINTER using the same 'Default' unity layer), and they are filtered out when the raycast result is constructed
 * Fixed layermask separate detection of CL_PHYSICS and CL_POINTER not being distinguished when setting only 1 of those in the raycast layermask
 
**NOTE: Tests will be handled in another PR**


unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Utils/ProtoConvertUtils.cs
```diff
@@ -168,5 +169,49 @@ public static DCLImageScaleMode ToDCLImageScaleMode(this BackgroundTextureMode t
...
+            if(hasPointerLayer && hasPhysicsLayer)
+                unityLayerMask |= (1 << PhysicsLayers.defaultLayer);
+
+            if (LayerMaskHasAnySDKCustomLayer(sdkLayerMask))
+            {
+                unityLayerMask |= (1 << PhysicsLayers.sdkCustomLayer);
+            }
+
+            return unityLayerMask;
+        }
```
pbosio, line 200: this function got me confused:
1- why are we creating a `new LayerMask()` if function returns an `int`?
2- is this function only used to calculate the layer for the raycast? 
maybe we should move it to where it is used instead of here and rename it for something like `GetRaycastTargetLayerMask` or something like that. WDYT?

unity-renderer/Assets/DCLPlugins/ECS7/Systems/RaycastSystem/ECSRaycastSystem.cs
```diff
@@ -0,0 +1,228 @@
...
+
+                if (model.QueryType == RaycastQueryType.RqtNone) continue;
+
+                // If the entity has a raycastResult, the first ray was already cast
+                if (raycastResultComponent.HasComponent(scene, entity))
+                {
+                    bool isContinuous = model.HasContinuous && model.Continuous;
+                    if (!isContinuous && raycastResultComponent.Get(scene, entity).model.Timestamp == model.Timestamp)
+                        continue;
+                }
```
pbosio, line 63: maybe we can add an internal component for raycast, create it or update it on raycast component update and removed after being processed (if it is not continuous)
that way we avoid always checking for this

pbosio (commented):
do we need to add the colliders on custom layer to the scene boundaries checker?

unity-renderer/Assets/DCLPlugins/ECS7/Systems/RaycastSystem/Tests/ECSRaycastSystemShould.cs
```diff
@@ -0,0 +1,173 @@
+namespace ECSSystems.ECSRaycastSystem.Tests
+{
+    public class ECSRaycastSystemShould
```
AjimenezDCL, line 3: We should be doing something with this. Either remove or fix the tests.

AjimenezDCL (approved):

unity-renderer/Assets/DCLPlugins/ECS7/Systems/RaycastSystem/ECSRaycastSystem.cs
pravusjif: Done, check [this commit](https://github.com/decentraland/unity-renderer/pull/4761/commits/08ae36fd041fa1d19a932a19bd9f14a24bd5dd10)

unity-renderer/Assets/DCLPlugins/ECS7/Systems/RaycastSystem/Tests/ECSRaycastSystemShould.cs
pravusjif: Removed tests folder until we implement them

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Utils/ProtoConvertUtils.cs
```diff
@@ -168,5 +169,49 @@ public static DCLImageScaleMode ToDCLImageScaleMode(this BackgroundTextureMode t
...
+        {
+            return IsInLayerMask(layerMask, (int)ColliderLayer.ClCustom1)
+                   || IsInLayerMask(layerMask, (int)ColliderLayer.ClCustom2)
+                   || IsInLayerMask(layerMask, (int)ColliderLayer.ClCustom3)
+                   || IsInLayerMask(layerMask, (int)ColliderLayer.ClCustom4)
+                   || IsInLayerMask(layerMask, (int)ColliderLayer.ClCustom5)
+                   || IsInLayerMask(layerMask, (int)ColliderLayer.ClCustom6)
+                   || IsInLayerMask(layerMask, (int)ColliderLayer.ClCustom7)
+                   || IsInLayerMask(layerMask, (int)ColliderLayer.ClCustom8);
+        }
```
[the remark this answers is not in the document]
pravusjif: Done: https://github.com/decentraland/unity-renderer/blob/3fe807a429140f31247c849dc18f608cb1da45bc/unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Utils/LayerMaskUtils.cs

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Utils/ProtoConvertUtils.cs
pravusjif: Done

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/MeshCollider/MeshColliderHandler.cs
```diff
@@ -105,21 +121,25 @@ private void SetColliderLayer(PBMeshCollider model)
...
 
-            if ((colliderLayer & pointerAndPhysicsMask) == pointerAndPhysicsMask)
+            // The order in this checks is important: from broader to more specific
+            if ((colliderLayer & LAYER_CUSTOM1) != 0 || (colliderLayer & LAYER_CUSTOM2) != 0
+              || (colliderLayer & LAYER_CUSTOM3) != 0 || (colliderLayer & LAYER_CUSTOM4) != 0
+              || (colliderLayer & LAYER_CUSTOM5) != 0 || (colliderLayer & LAYER_CUSTOM6) != 0
+              || (colliderLayer & LAYER_CUSTOM7) != 0 || (colliderLayer & LAYER_CUSTOM8) != 0)
+            {
+                colliderGameObject.layer = PhysicsLayers.sdkCustomLayer;
+            }
```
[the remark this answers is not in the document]
pravusjif: Done, updated everything to comply with that.

pbosio (approved):

PR #4766, opened 2023-03-28
fix: force a compatible texture format for the platform

AjimenezDCL:
Use GraphicsFormatUtility to get a compatible texture for every platform.

## Test
 No "Texture creation failed" should appear, check WebGL, Windows and Mac builds. Additionally, the infinite floor should look correct.

### DIFF
--- a/unity-renderer/Assets/DCLPlugins/MapTexturePlugin/MapTexturePlugin.cs
+++ b/unity-renderer/Assets/DCLPlugins/MapTexturePlugin/MapTexturePlugin.cs
@@ -2,8 +2,8 @@ using System;
 using System.Threading;
 using Cysharp.Threading.Tasks;
 using DCL;
-using System.Linq;
 using UnityEngine;
+using UnityEngine.Experimental.Rendering;
 using UnityEngine.Networking;
 
 public class MapTexturePlugin : IPlugin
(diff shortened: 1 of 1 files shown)


unity-renderer/Assets/DCLPlugins/MapTexturePlugin/MapTexturePlugin.cs
```diff
@@ -52,7 +47,7 @@ private static async UniTaskVoid DownloadTexture(string url, BaseVariable<Textur
                 }
                 else
                 {
-                    Texture2D tex = new Texture2D(0, 0, TextureFormat.ETC2_RGBA8, false, true);
+                    Texture2D tex = new Texture2D(0, 0, GraphicsFormatUtility.GetGraphicsFormat(TextureFormat.RGBA32, false), TextureCreationFlags.None);
```
mikhail-dcl, line 50: Could not we just use `textureFormat` instead? It was the only oversight here I believe?

AjimenezDCL: What I dont like about that approach is that we rely on a fixed set of textures formats and it doesnt scale that much. With this one we just ask the system to return a compatible GraphicsFormat with the needs we have.

mikhail-dcl: Can we delete `PRIORITIZED_FORMATS` then?

mikhail-dcl (approved):

kwssbocian (approved):
✔️No issues observed during testing. Additionally no occurrences of "Texture creation failed" error.

PR #4781, opened 2023-03-28
fix: sdk7 scene bounds checker transform scale

pravusjif:
### WHY
When a transform `scale` or `rotation` is changed, the Scene Bounds Checker is not recalculating the mesh bounds and it should.

### WHAT

implemented meshesDirty change when a transform scale or rotation is changed

### DIFF
--- a/unity-renderer/Assets/DCLPlugins/ECS7/InternalECSComponents/Interfaces/Extensions/InternalSceneBoundsCheckExtensions.cs
+++ b/unity-renderer/Assets/DCLPlugins/ECS7/InternalECSComponents/Interfaces/Extensions/InternalSceneBoundsCheckExtensions.cs
@@ -26,12 +26,32 @@ namespace DCL.ECS7.InternalComponents
 
             // Update children position in their SBCComponent as well
             IList<long> childrenId = entity.childrenId;
-
             for (int i = 0; i < childrenId.Count; i++)
             {
                 if (scene.entities.TryGetValue(childrenId[i], out IDCLEntity childEntity))
                 {
                     sbcInternalComponent.SetPosition(scene, childEntity, childEntity.gameObject.transform.position, createComponentIfMissing);
+                }
+            }
+        }
+
+        public static void OnTransformScaleRotationChanged(this IInternalECSComponent<InternalSceneBoundsCheck> sbcInternalComponent,
+            IParcelScene scene, IDCLEntity entity)
+        {
+            var model = sbcInternalComponent.GetFor(scene, entity)?.model ?? new InternalSceneBoundsCheck();
+
+            // Mesh bounds need to be recalculated
+            model.meshesDirty = true;
+
+            sbcInternalComponent.PutFor(scene, entity, model);
+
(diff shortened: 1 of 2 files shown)


pbosio (approved):

unity-renderer/Assets/DCLPlugins/ECS7/InternalECSComponents/Interfaces/Extensions/InternalSceneBoundsCheckExtensions.cs
```diff
@@ -36,6 +35,27 @@ public static class InternalSceneBoundsCheckExtensions
             }
         }
 
+        public static void OnTransformScaleRotationChanged(this IInternalECSComponent<InternalSceneBoundsCheck> sbcInternalComponent,
+            IParcelScene scene, IDCLEntity entity)
+        {
+            var model = sbcInternalComponent.GetFor(scene, entity)?.model ?? new InternalSceneBoundsCheck();
```
eordano, line 41: Can we avoid the allocation and use always the same instance?

pravusjif: Sure, sounds good to me, done!

unity-renderer/Assets/DCLPlugins/ECS7/InternalECSComponents/Interfaces/Extensions/InternalSceneBoundsCheckExtensions.cs
```diff
@@ -7,11 +7,12 @@ namespace DCL.ECS7.InternalComponents
 {
     public static class InternalSceneBoundsCheckExtensions
     {
+        private static InternalSceneBoundsCheck model;
```
pbosio, line 10: why are we adding this static field?

unity-renderer/Assets/DCLPlugins/ECS7/InternalECSComponents/Interfaces/Extensions/InternalSceneBoundsCheckExtensions.cs
pbosio: there are no extra allocations on that piece of code, after a new instance is created is being stored, so next call will be using the stored instance instead of creating a new one.
that is what that line `var model = sbcInternalComponent.GetFor(scene, entity)?.model ?? new InternalSceneBoundsCheck();` does, it get the stored instance or it creates a new one if there is none stored,
2 lines later the instance get stored `sbcInternalComponent.PutFor(scene, entity, model);`

unity-renderer/Assets/DCLPlugins/ECS7/InternalECSComponents/Interfaces/Extensions/InternalSceneBoundsCheckExtensions.cs
pravusjif: reverted

lorux0 (approved):

PR #4782, opened 2023-03-28
feat: profile messages need not be reliable

eordano:
According to this report, a large number of requests are waiting for a reliable response, thus increasing the EventEmitter listener count.

Upon taking a look at the comms protocol, it looks like profile requests are asking to be reliably transmitted. This includes the profile updates and the "RPC" protocol to request profiles from peers. Since the profile updates get sent every second, and the RPC protocol has a retry mechanism, there is no need for these three kinds of messages to be sent with the reliable flag.

image


marianogoldman:
Fixed in https://github.com/decentraland/unity-renderer/pull/4801.

PR #4792, opened 2023-03-29
fix: duplicated key in emote animation tracker + reenabled tests

AjimenezDCL:
Fix some inconsistencies if the same emote was requested twice. Also fixed and reenabled some tests.

# Test
Triggering the error was super tricky, go to places with many people and look for a `An item with the same key has already been added` error. It shouldnt be present in the console.

Also take a look at emotes en general.

## Our Code Review Standards


Kinerius (approved):

popuz (approved):

kwsskulinski (approved):
Issue looks to be fixed. No 'An item with the same key has already been added' errors found.

PR #4797, opened 2023-03-29
hotfix: add fatal error when no valid catalyst are found

kuruk-mm:
Possible edge case: If your internet goes down, the client will try to reconnect to a different realm, and it could set all the realms to already connected, and the realm selected could be null. So on that case, we bring a fatal error.

## Our Code Review Standards

### DIFF
--- a/browser-interface/packages/shared/dao/sagas.ts
+++ b/browser-interface/packages/shared/dao/sagas.ts
@@ -172,7 +172,9 @@ function* selectRealm() {
     // cached in local storage
     (yield call(getRealmFromLocalStorage, network))
 
-  if (!realm) debugger
+  if (!realm) {
+    BringDownClientAndReportFatalError(new Error('No valid realm found'), 'comms#init')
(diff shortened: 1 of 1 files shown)


marianogoldman (approved):

sandrade-dcl (approved):

browser-interface/packages/shared/dao/sagas.ts
```diff
@@ -172,7 +172,9 @@ function* selectRealm() {
     // cached in local storage
     (yield call(getRealmFromLocalStorage, network))
 
-  if (!realm) debugger
+  if (!realm) {
+    BringDownClientAndReportFatalError(new Error('No valid realm found'), 'comms#init')
```
eordano, line 176:
```suggestion
    BringDownClientAndReportFatalError(new Error('Could not connect to any catalyst servers. Please check your internet connection and try again.'), 'comms#init')
```

PR #4800, opened 2023-03-29
fix: usernames in chat messages

lorux0:
## What does this PR change?

Fetches the profile, if not available, when adding a chat message.
Fixes that sometimes you see the address of the user instead of the name in the message, either sender or recipient. 

## How to test the changes?

1. Launch the explorer
2. Send and receive messages in ~nearby
3. Send and receive messages in private conversations
4. Send and receive messages in #channels (recommended to use #test)
5. If at any point you see an address instead of the user name, it should change to a proper name after a small period of time

## Our Code Review Standards


mikhail-dcl (approved):

kwsskulinski (approved):
Issue has not been found.

Issue #4513, opened 2023-03-02
Change Island is interpreted as Realm Reconnection


eordano:
Related to: https://github.com/decentraland/unity-renderer/issues/4489

It looks like the catalyst decided to change my island, and the kernel interpreted it as a disconnection+reconnection.

The likely function responsible for this behavior is [handleCommsReconnectionInterval](https://github.com/decentraland/unity-renderer/blob/dev/browser-interface/packages/shared/comms/sagas.ts#L388).

<img width="521" alt="image" src="https://user-images.githubusercontent.com/42750/222294132-11373e74-0c5b-4c9c-88c5-bcc4db96330e.png">

Additionally, this triggers many profile fetches at once, probably hitting some query limit:
<img width="514" alt="image" src="https://user-images.githubusercontent.com/42750/222294268-3bed495d-ba44-47e6-88c9-a5929183883d.png">
<img width="958" alt="image" src="https://user-images.githubusercontent.com/42750/222294327-576ba610-8442-4129-b02b-bf51b338297c.png">
<img width="828" alt="image" src="https://user-images.githubusercontent.com/42750/222294362-1e247d5c-d1d8-4df2-81e4-0907af74ea1b.png">

xavi-dcl:
This is causing a lot of trouble during the event.

"I’m not being teleported to another space, but rather being constantly teleported to the spawn points in each space/scene"
"Anyone else having a chat issue? It just shows me changing realms over and over but I can’t get the actual chat to load fully (it shows it greyed out not in the chat box)--- on desktop client"

More details on the [issues tracker](https://docs.google.com/spreadsheets/d/1AttGJQli1cQJV1nLLNK1y2RJQP8b4VhQHz6NxX2ejjU/edit#gid=1672801883&range=B10) and on [Slack](https://decentralandteam.slack.com/archives/C04EBK18XS6/p1680043110531339)

![image.png](https://images.zenhubusercontent.com/636b99a62654f2e351ceea55/3766dbcd-e72f-402d-8077-b09c7ab86337)

PR #4804, opened 2023-03-29
hotfix: bring down client on id duplicated

moliva:
## What does this PR change?

The current PR partially addresses #4513 bringing down the client and removing disabling comms from retrying to connect to the realm when it was initially kicked due to a duplicated id error (i.e. two tabs open with the same wallet address).

## How to test the changes?

1. Launch the explorer on one tab with a given wallet
2. Launch the explorer on another tab with the same wallet
3. At least one of the two should get an error explaining the issue, bringing down the client and disconnecting from comms
4. The tab with no issue or a new tab if the previous two erred should be available to connect successfully with no realm issues

## Our Code Review Standards

### DIFF
--- a/browser-interface/packages/shared/loading/ReportFatalError.ts
+++ b/browser-interface/packages/shared/loading/ReportFatalError.ts
@@ -87,6 +87,8 @@ export function BringDownClientAndReportFatalError(
 ) {
   let sagaStack: string | undefined = payload['sagaStack']
 
+  debugger
+
   if (sagaStack) {
     // first stringify
     sagaStack = '' + sagaStack
--- a/browser-interface/packages/shared/loading/ReportFatalError.ts
+++ b/browser-interface/packages/shared/loading/ReportFatalError.ts
@@ -87,8 +87,6 @@ export function BringDownClientAndReportFatalError(
 ) {
   let sagaStack: string | undefined = payload['sagaStack']
 
-  debugger
-
   if (sagaStack) {
     // first stringify
     sagaStack = '' + sagaStack
@@ -105,6 +103,8 @@ export function BringDownClientAndReportFatalError(
     stack: getStack(error).slice(0, 10000),
     saga_stack: sagaStack
   })
+
+  store.dispatch(fatalError(error.message || 'fatal error'))
 
   globalObservable.emit('error', {
     error,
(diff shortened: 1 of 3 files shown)


marianogoldman (approved):
LGTM

marianogoldman:
Testing it with 2 tabs... what happened is both tabs got to the same state. They both show the error, they keep playing the music (from Genesis Plaza) and they both keep high activity in the console while trying to reconnect.
Screenshot

browser-interface/packages/shared/loading/ReportFatalError.ts
```diff
@@ -87,8 +87,6 @@ export function BringDownClientAndReportFatalError(
 ) {
   let sagaStack: string | undefined = payload['sagaStack']
 
-  debugger
```
kuruk-mm, line 90: Why are we removing this debugger? It's useful for debugging. The `debugger` only triggers if you have the console opened.

kuruk-mm (changes requested):

browser-interface/packages/shared/loading/ReportFatalError.ts
moliva: > Testing it with 2 tabs... what happened is both tabs got to the same state. They both show the error, they keep playing the music (from Genesis Plaza) and they both keep high activity in the console while trying to reconnect. Screenshot

Yes, I see two things with the current approach:
1. It takes some time before clients realize they need to go down
2. There's still reconnecting logic connected but that due to having a fatal error set, [should not be actually invoking the connection logic here](https://github.com/decentraland/unity-renderer/blob/hotfix/brind-down-client-on-id-duplicated/browser-interface/packages/shared/comms/sagas.ts#L413)

moliva:
@marianogoldman 
> Testing it with 2 tabs... what happened is both tabs got to the same state. They both show the error, they keep playing the music (from Genesis Plaza) and they both keep high activity in the console while trying to reconnect. Screenshot

Yes, I see two things with the current approach:
1. It takes some time before clients realize they need to go down
2. There's still reconnecting logic connected but that due to having a fatal error set, [should not be actually invoking the connection logic here](https://github.com/decentraland/unity-renderer/blob/hotfix/brind-down-client-on-id-duplicated/browser-interface/packages/shared/comms/sagas.ts#L413)

PR #4816, opened 2023-03-30
fix: hiccup opening friends hud

### LINKED ISSUE #4680: Clicking on Friends do not open the panel right away
The Friends section is not opened as soon as I click on it, as it seems to be loading the data before showing the layout with the spinner.

It feels like the click failed, and I click again, actually closing the Friends panel, and never showing up.

If you click on any player inside the friend list, it opens it conversation. Clicking on the back button also does not respond immediately, and it feels like an error.

**Acceptance Criteria**
- Friends panel should open immediately on click in the icon of the social bar
- This also apply when going back from a conversation
- If the friend list is not loaded, it should show the spinner
- As soon as the spinner finishes loading, it should show the friend list

sandrade-dcl:
## What does this PR change?
Fix #4680 

We were having a huge hiccup each time we opened the FriendsHUD, specially if we had some friends with a very long list of `muted` or `blocked` users in their profiles.

### Cause
I noticed that the problem was in the way how we were comparing, inside `UserProfileModel.Equals(...)`, if 2 profiles had the same list of `blocked`/`muted` users. We were using:
```
if (!model.blocked.All(blocked.Contains) || model.blocked.Count != blocked.Count)
    return false;
...
if (!model.muted.All(muted.Contains) || model.muted.Count != muted.Count)
    return false;
```
But these `All` operations together with `Contains` to check if a list contains the same elements than another one is very expensive and, in the cases of friends that have a huge amount of blocked and/or muted users set in their profiles (I have seen profiles with more than 10k elements there), was making the loading of the friends list get freezed for several seconds.

### Solution
To find a more efficient way to compare the lists:
```
if (model.blocked.Count != blocked.Count ||
    model.blocked.Except(blocked).Any() ||
    blocked.Except(model.blocked).Any())
    return false;
...
if (model.muted.Count != muted.Count ||
    model.muted.Except(muted).Any() ||
    muted.Except(model.muted).Any())
    return false;
```
#### Edited
We finally decided (read [this](https://github.com/decentraland/unity-renderer/pull/4816#issuecomment-1491662691)) to compare only the lenght of the lists:
```
if (model.blocked.Count != blocked.Count)
    return false;
...
if (model.muted.Count != muted.Count)
    return false;
```

## How to test the changes?
1. Go to: https://play.decentraland.org/?explorer-branch=fix/huge-hiccup-opening-friends-hud
2. Open the FriendsHUD clicking on the corresponding button in the taskbar.
3. Check the list of friends loads quickly just after opening the HUD.

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/UserProfile/UserProfileModel.cs
```diff
@@ -93,15 +93,19 @@ public bool Equals(UserProfileModel model)
 
         if (model.blocked != null && blocked != null)
         {
-            if (!model.blocked.All(blocked.Contains) || model.blocked.Count != blocked.Count)
+            if (model.blocked.Count != blocked.Count ||
```
lorux0, line 96: i think we could gain more performance by using a `HashSet` for the `blocked` & `muted` lists, what do you think?

sandrade-dcl: I have tried but using directly `HashSet` as part of the `UserProfileModel` model, it cannot serialize well the `blocked` and `muted` lists that comes from kernel.

I could try to convert the lists to `HashSet` just before doing the comparisions, but I'm not sure if it would worth it since we would have to duplicate the data. Wdyt?

sandrade-dcl:
### IMPORTANT UPDATE
Even with these optimizations, I have checked that some users (for example `olavra`) are still experiencing huge hiccups of several seconds when open the friend HUD. This raises the question of whether it really worth it to take into account the comparison of these lists, element by element, for comparing 2 instances of `UserProfileModel`.

Taking into account that both lists shouldn't be part of the profile in the catalysts (we were already discussing it in the past) and that the impact in terms of performance right now is too much, I suggest to remove the comparison of these 2 lists and maybe only leave a simple comparison of theirs lengths.

WDYT @lorux0?

unity-renderer/Assets/Scripts/MainScripts/DCL/UserProfile/UserProfileModel.cs
lorux0: We've discussed to remove the content comparison of both lists and only do the size checking. We need a proper fix in the backend. The lists being too big could cause hiccups, not only here, but in many places when processing the profile

dalkia (approved):

lorux0 (approved):

AjimenezDCL (approved):

sandrade-dcl:
### Extra changes requested by Product
Product has requested the next extra changes:
- [x] Fix the scrollbar velocity of the chat members hud.
- [x] Temporally make the [GO TO CROWD] button link to `0,0` instead of the real most filled place.

Changes done here: https://github.com/decentraland/unity-renderer/pull/4816/commits/116c3f4644cb0a95370dfef61715a7990e9f0f99

kwssbocian (approved):
✔️No issues observed regarding friendsHUD functionality.

Issue #4133, opened 2023-01-26
Genesis Plaza "Timeout" that is not a timeout


eordano:
Testing the new onboarding and flow, I stumble with this with a ~20% repro rate

![image](https://user-images.githubusercontent.com/42750/214832835-1c157ff2-af24-4e6a-a86a-469a910ea7a5.png)

Steps:
1. Clear cache and application storage
2. Enter the world as guest
3. Skip the tutorial by clicking on "Go to Genesis City"
4. After a few seconds, a sign saying "Loading timeout" appears and Genesis plaza is missing key models

PR #4819, opened 2023-03-30
chore: move timeout control to renderer

dalkia:
## What does this PR change?

1. Moves the timeout control to the renderer. Now we dont rely in the kernel call, avoiding issue like this #4133 
2. There was a race condition when LOS=1 where the scene was ready before the teleport was requested. This left a scenario were loading was stuck at 0%. This was fixed by checking the scene complete on teleport requested

### <samp>🤖 Generated by Copilot at 4a89752</samp>

Added a timeout feature to the loading screen that notifies the user if the scene is not loaded in time. Refactored and simplified some loading screen and loading status logic using `async` tasks and `CancellationToken`s. Modified `LoadingScreenController.cs` and related files.

## How to test the changes?

1. Test the basic teleport method (goto coordinates, crowd, magic). Loading screen should appear
2. Test the teleport methods from the explorer menu.  Loading screen should appear
3. Change realm without teleporting. Loading screen should not appear
4. Go to different worlds (mgoldman.dcl.eth; dclonboarding.dcl.eth) and back to genesis city. Check that the loading screen appears correctly
5. Select offline in the throtlling devtools console to stop the loading of a scene while the loading screen is on. The timeout should appear after 120 seconds

image

6. Do the signup flow. Everything should work as expected

## Our Code Review Standards


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/LoadingScreen/Scripts/LoadingScreenController.cs
```diff
@@ -148,28 +154,39 @@ private bool IsNewRealm()
...
             loadingScreenDataStore.decoupledLoadingHUD.visible.Set(false);
         }
+
+        private void DoTimeout()
+        {
+            notificationsController.ShowNotification(new Model
+            {
+                message = "Loading scene timeout",
+                type = Type.GENERIC,
+                timer = 10f,
```
eordano, line 186: Let's make this 60s
```suggestion
                timer = 60f,
```

eordano (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/LoadingScreen/Scripts/LoadingScreenController.cs
dalkia: That one is the time the notification remains on, I think you meant to modify LOAD_SCENE_TIMEOUT, which is currently at 120 seconds. Right?

popuz (approved):
LGTM. However, stopping with the `devtools->offline` didn't work for me. If I enable offline on "downlading" phase, then 120 timeout do not show up. If I enable it on "loading" phase, then it loads just normally

eordano:
@dalkia merged dev to try out copilot

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/LoadingScreen/Scripts/LoadingScreenController.cs
eordano: Thanks!

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/LoadingScreen/Scripts/LoadingScreenController.cs
```diff
@@ -113,14 +116,17 @@ private void TeleportRequested(Vector3 current, Vector3 previous)
...
-            }
+                StartTimeoutCounter(timeoutCTS.Token);
+            }else if (IsSceneLoaded(currentDestinationCandidate))
+                HandlePlayerLoading();
+        }
+
+
+        private async UniTaskVoid StartTimeoutCounter(CancellationToken ct)
+        {
+            await UniTask.Delay(LOAD_SCENE_TIMEOUT, cancellationToken: ct);
```
Kinerius, line 128: I would change `LOAD_SCENE_TIMEOUT` to `TimeSpan.FromMilliseconds(LOAD_SCENE_TIMEOUT)` for the sake of readability

Kinerius (approved):
Nice! This PR is also fixing the Timeout caused by leaving the tab in the background, sometimes it was much less than 120 seconds.

kwssbocian (approved):
✔️All teleportation methods work correctly.
✔️Changing realms does not cause the player character to be teleported.
✔️Loading screen correctly appears when teleporting to and from worlds.
✔️After 2 minutes of loading the timeout correctly occurs.
Additionally all previous issues regarding incorrect timeout messages have been fixed.

PR #4848, opened 2023-04-03
refactor:  Simplify base avatar and remove revealer

AjimenezDCL:
Refactor BaseAvatar to include the revealing process. This should simplify the process and remove nasty race conditions and bugs.

## Test
- Other players do the hologram transition properly
- In 3rd person camera, you can see the hologram in-world when changing wearables (not in the backpack).

## Our Code Review Standards


kwssbocian (approved):
✔️No issues observed regarding the process of other avatar transitioning from hologram outline to normal avatar upon loading.
✔️Own avatar briefly shows as a hologram in 3rd person after changing wearables but quickly transitions to the regular avatar.

kwssbocian (approved):
✔️No changes observed since previous testing. Own and other avatars all transition properly from the hologram to the normal form once loading is completed.
✔️No invisible avatars observed during testing.

PR #4850, opened 2023-04-03
chore: remove empty virtual updates

### LINKED ISSUE #4828: Remove Empty Virtual Updates
### Description
The game uses the BaseComponentView class as a parent to many scripts, which declares a virtual
Update() function. Even though it might look like the Update in Figure 6 is not doing anything if not
overridden, the reality is that every event function that is declared in a script will cause an overhead by
Unity having to iterate and call all the registered functions. See: https://blog.unity.com/engine-platform/10000-update-calls

We have over 100 similar empty virtual calls.

### Deliverables
- [ ] Remove unnecessary virtual Update calls
- [ ] If there are any which would be better suited to a manager (with no mono behaviour) create tickets to explain these.

popuz:
## What does this PR change?
Closes #4828

- removed virtual in a base class
- adjusted inherited overrides where needed

## How to test the changes?

1. In general verify that all UI works as previously

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 5ea82e1</samp>

Removed the `Update` method from the `BaseComponentView` class and changed the `Update` methods in its subclasses from `override` to `new` to avoid inheritance conflicts and improve performance. Also removed some unnecessary blank lines from various classes to improve code style.


kwsskulinski (approved):
UI works as previously. No issue has been found.

Kinerius (approved):
Free win :)

eordano (approved):

PR #4853, opened 2023-04-03
fix: check for livekit room status before publishing data

hugoArregui:


kuruk-mm:
Replaced by https://github.com/decentraland/unity-renderer/pull/4869

PR #4855, opened 2023-04-03
feat: sdk7 add more component group types

pbosio:
add more component gorups types:
* group of entities that match three component
* group of entities with a component but without a specific component
* group of entities with two component but without a specific component


pravusjif (approved):

Kinerius (approved):
LGTM!

PR #4913, opened 2023-04-07
Fix/minimap data doesnt support deletion

popuz:
## What does this PR change?

...

## How to test the changes?

1. Launch the explorer
2. ...

## Our Code Review Standards

## Copilot summary

copilot:summary


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/InputController/InputController.cs
```diff
@@ -144,7 +144,8 @@ private void Update_Trigger(InputAction_Trigger[] triggerTimeActions)
 #endif
                     break;
                 case DCLAction_Trigger.ToggleNavMap:
-                    InputProcessor.FromKey(action, KeyCode.M, modifiers: InputProcessor.Modifier.FocusNotInInput);
+                    if (!DataStore.i.common.isWorld.Get())
```
mikhail-dcl, line 147: It looks like a hack but we can live with it

popuz: Agree, I didn't like it as well. I saw several similar code-lines in that InputController and decided to go with it. Also because this is a temporary solution as I understand, until we clarify what to do with the Map in the World

dalkia (changes requested):

kwssbocian (approved):
✔️In worlds map cannot be opened using the M key and is not available in the Explore menu.
✔️After teleporting back to the main world the map is fully functional. POI in the popup is also shown without issues.

dalkia (approved):

PR #4920, opened 2023-04-10
feat: add force realm and version links to realm picking algorithm

aleortega:
## What does this PR change?

It adds a two new links to the Realm Picking algorithm. The first is named `VERSION_PEERS` and expects minimum versions of either comms, content, lambdas or bff through configuration file to pick the candidates that meet those version requirements (e.g: configuration files says that content versions needs to be at least 6.0.0 so this link filters out all the catalyst that are using lower versions than 6.0.0). The second link is named `FORCE_PEERS` and will be used to force access to candidates.

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 37627da</samp>

This pull request adds new features and tests to the pick realm algorithm, which is used to select the best catalyst server and realm for a given parcel. It also updates the `semver` dependency and the `fetchCatalystStatus` function to include the version information of the services provided by the catalyst servers.


Kinerius (approved):

eordano:
Can we rename "PEERS" to "CATALYST"?

aleortega:
> Can we rename "PEERS" to "CATALYST" or "REALM"? I find the term extremely confusing

Agree! I'm changing all the links names to CATALYST before merging it. Thanks!

aleortega:
> > Can we rename "PEERS" to "CATALYST" or "REALM"? I find the term extremely confusing
> 
> Agree! I'm changing all the links names to CATALYST before merging it. Thanks!

Keep Peers for ALL_PEERS_SCORE and CLOSE_PEERS_SCORE since those links refers to analyzing the amounts of users connected to the realms at that time, IMO peers is the correct word in those cases. Let me know if you disagree with this, I can change them later.
Besides, I renamed the two links added by me to VERSION_CATALYST and FORCE_CATALYST thanks!

PR #4924, opened 2023-04-10
chore: sdk7 raycast test coverage

pravusjif:
* Fixed raycast bug that allowed to return a raycast result with hits with no entities in certain edge cases
* Corrected raycast default direction to use when invalid direction values are received
* Added RaycastHandler and RaycastSystem test coverage

### <samp>🤖 Generated by Copilot at c150c2a</samp>

This pull request adds unit tests for the `ECSRaycastSystem` class and fixes a bug in the `CreateSDKRaycastHit` method. The tests use the `ECS7TestUtilsScenesAndEntities` helper class and the NSubstitute library to mock the ECS7 components and the component writer interface. The bug fix prevents a null reference exception when the hit entity is null and ensures the hit entity id is always valid. The changes affect the files `ECSRaycastSystem.cs`, `ECSRaycastSystemShould.cs`, and their corresponding meta and assembly definition files.


unity-renderer/Assets/DCLPlugins/ECS7/Systems/RaycastSystem/ECSRaycastSystem.cs
```diff
@@ -161,8 +161,10 @@ private Ray CreateRay(IParcelScene scene, IDCLEntity entity, PBRaycast model)
                     break;
                 case PBRaycast.DirectionOneofCase.TargetEntity:
                     // Target entity to cast the ray towards
-                    IDCLEntity targetEntity = scene.GetEntityById(model.TargetEntity);
-                    rayDirection = targetEntity.gameObject.transform.position - entityTransform.position;
+                    if(scene.entities.TryGetValue(model.TargetEntity, out IDCLEntity targetEntity))
+                        rayDirection = targetEntity.gameObject.transform.position - entityTransform.position;
+                    else
+                        Debug.LogError("Raycast error: can't find targetEntity, direction calculation failed.");
```
menduz, line 167: In this case we fall back to the root entity

pravusjif: So in that case the direction should be calculated with the scene's 0,0,0 ?

menduz: Yes!

In ECS all entities exist all the times, components don't. So if we cant find a transform component for the entity, but it has i.e. a MeshRenderer, some transform must be used. That is the Identity transform: position=0, scale=1 parent=root


It is the same for this case. If no entity exists, we assume a position of an identity transform. Which (should) equal the root entity position

pravusjif: Cool, I just changed it and added a test: https://github.com/decentraland/unity-renderer/pull/4924/commits/a141fcc68e90bab31f311c1cae0f9c05dc3a4ed4

pravusjif: OK then the default direction should always be the GLOBAL forward vector? or the LOCAL forward vector (AKA transform.forward vector) of the raycast entity?
I'll change it again

pravusjif: Corrected at https://github.com/decentraland/unity-renderer/pull/4924/commits/7992a536b8f204c59b0cf52af72dead6942b7819 and added a test for every direction type defaulting

lorux0 (approved):

PR #4942, opened 2023-04-12
fix: importing gltf with empty library fails with a null ref

Kinerius:
## What does this PR change?

Updated gltfast package with the fix.

The fix: https://github.com/decentraland/glTFast/commit/600a6c7dc0ca608685997519e2f356839abba3a2

Also removed the GLB as a dependency, extracted its mesh into an asset with a new script, and replaced the reference in the prefab.

## Our Code Review Standards


kuruk-mm (approved):

AjimenezDCL (approved):

unity-renderer/Assets/Rendering/LoadingAvatar/Editor/BaseAvatarTest.cs
```diff
@@ -0,0 +1,18 @@
+using NUnit.Framework;
+using UnityEditor;
+using UnityEngine;
+
+namespace Rendering.LoadingAvatar.Editor
+{
+    public class BaseAvatarTest
```
dalkia, line 7:
```suggestion
    [TestFixture]
    [Category("EditModeCI")]
    public class BaseAvatarTest
```

Remember to add the EditModeCI category, otherwise is gonna be ignored by the CI

PR #4959, opened 2023-04-14
feat: avatar dynamic slot list for new backpack

davidejensen:
## What does this PR change?

...

## How to test the changes?

Cannot be tested

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 3696ab2</samp>

This pull request adds the AvatarSlot folder and its contents, which are the files and assets that implement the avatar slot feature for the BackpackEditorHUDV2. The avatar slot feature allows the user to select and edit the components of their avatar, such as the head, the body, and the wearables. The files and assets include the data models, the views, the controller, the interfaces, the scriptable objects, and the meta files for the avatar slot feature.


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/AvatarSlot/AvatarSlotsHUDController.cs
```diff
@@ -0,0 +1,49 @@
...
+    public class AvatarSlotsHUDController : IHUD
+    {
+        internal AvatarSlotsDefinitionSO avatarSlotsDefinition;
+        internal readonly IAvatarSlotsView avatarSlotsView;
+        internal string lastSelectedSlot;
+
+        public AvatarSlotsHUDController(IAvatarSlotsView avatarSlotsView)
+        {
+            this.avatarSlotsView = avatarSlotsView;
+            avatarSlotsView.OnToggleAvatarSlot += ToggleSlot;
```
sandrade-dcl, line 17: You should unsubscribe it in the `Dispose()` method.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/AvatarSlot/AvatarSlotsHUDController.cs
```diff
@@ -0,0 +1,49 @@
...
+            avatarSlotsDefinition = Resources.Load<AvatarSlotsDefinitionSO>("AvatarSlotsDefinition");
```
sandrade-dcl, line 18: As we were discussing in our last meeting with @lorux0, you and me, this specific behavior of getting the slots configuration from a SO should be abstracted into an interface, for example `IAvatarSlotsConfigProvider`, and have a specific implementation for getting the config from a SO (in the future we could have another specific implementation that would use the same interfaz but for getting the config from backend for example). The idea would be to inject that provider interface in this controller and abstract it from the implementation (it would be basically to follow a Strategy pattern).

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/AvatarSlot/AvatarSlotsHUDController.cs
```diff
@@ -0,0 +1,49 @@
...
+        public void GenerateSlots()
+        {
+            for (var i = 0; i < avatarSlotsDefinition.slotsDefinition.Length; i++)
+            {
+                SerializableKeyValuePair<string, List<string>> section = avatarSlotsDefinition.slotsDefinition[i];
+                avatarSlotsView.CreateAvatarSlotSection(section.key, i < avatarSlotsDefinition.slotsDefinition.Length - 1);
+                foreach (string avatarSlotSection in section.value)
+                    avatarSlotsView.AddSlotToSection(section.key, avatarSlotSection);
+            }
+            avatarSlotsView.RebuildLayout();
```
sandrade-dcl, line 30: Is this completely needed? We should try to avoid as much as possible to force the layouts rebuildings.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/AvatarSlot/AvatarSlotsView.cs
```diff
@@ -0,0 +1,62 @@
...
+        public void RebuildLayout()
+        {
+            // Needed because adding elements to a transform that contains a
+            // layout does not refresh the child placements
+            Utils.ForceRebuildLayoutImmediate(avatarSlotsContainer);
+        }
+
+        public void AddSlotToSection(string sectionName, string slotCategory)
+        {
+            IAvatarSlotComponentView avatarSlot = Instantiate(avatarSlotPrefab, avatarSlotSections[sectionName]).GetComponent<IAvatarSlotComponentView>();
```
sandrade-dcl, line 39: I think we should use object pooling to manage the slots instantiations. Maybe it's not currently so important since we have a fixed list of items defined in a SO, but if it's going ot change in the future to be more dynamic, we should be prepared for that. Wdyt? (we can discuss it among the 3 of us and make a decision if you want).

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/AvatarSlot/IAvatarSlotsView.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace DCL.Backpack
+{
+    public interface IAvatarSlotsView
+    {
+        public event Action<string, bool> OnToggleAvatarSlot;
```
sandrade-dcl, line 10:
```suggestion
        event Action<string, bool> OnToggleAvatarSlot;
```

sandrade-dcl (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/AvatarSlot/AvatarSlotsHUDController.cs
davidejensen: Sadly it is needed because when adding objects as children of a vertical layout it doesn't refresh automatically

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/AvatarSlot/AvatarSlotsView.cs
davidejensen: We can do it but when we add support in the future for dynamic avatars, now it's done on start and never changes, so the pool would be "pointless" and wouldn't add any improvements

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/AvatarSlot/AvatarSlotsHUDController.cs
davidejensen: Yes I understand and it makes sense but I would do that once we go into the next phase of the avatar v2, if ever

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/AvatarSlot/AvatarSlotsHUDController.cs
sandrade-dcl: Ok, understood... although I think it would be a good practice to abstract the logic. It would be an easy change. Anyway, I agree with this option if we all decide to follow it 👍 Wdyt @lorux0?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/AvatarSlot/AvatarSlotsView.cs
sandrade-dcl: Ok, in this case there is no problem as it is going to be instantiated only at the beggining 👍

sandrade-dcl (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/AvatarSlot/AvatarSlotsHUDController.cs
sandrade-dcl: Ok, understood... although I think it would be a good practice to abstract the logic. It would be an easy change. Anyway, I agree with this option if we all decide to follow it 👍 Wdyt @lorux0?

PR #4994, opened 2023-04-18
fix: map showing in worlds

### LINKED ISSUE #4842: Do not bring the minimap when pressing the U key inside a World scene
The minimap feature is not supported in World scenes, when using the U shortcut twice you can bring it back and as a result, you get the Go to Genesis City button overlapped with the minimap.

Screenshot 2023-04-03 at 09 32 46

### Acceptance criteria
- [ ] Bug described above is fixed

### Note:
Could this PR that was merged, then reverted, have overlapping code base? https://github.com/decentraland/unity-renderer/pull/4913
Let's coordinate to avoid conflicts.

### Related issue:
Adding a related issue we could maybe fix together with this:
> On production, the minimap immediately appeared for me only on the first login for a fresh guest when I spawned in the onboarding. I then went to Genesis, then other worlds and the minimap wasn't immediately visible, I had to hide+unhide the UI to make it appear.
- [ ] Expected behavior when fixed: "Go to Genesis City" button is shown, and Minimap is NOT shown. 

If not being fixed together, please split this section to a separate issue.

dalkia:
## What does this PR change?

Fixes #4842

Adds a condition to the ShowHideUIByTrigger class and does a refactor to the MinimapHUD.

A further and more clean refactor would be to move the responsibility of showing/hiding the canvas group from the `ShowUIByTrigger` class into the respective view, which could work through the `BaseComponentView` abstract class. Therefore, leaving only the listener in the `ShowUIByTrigger` class. 

This can take a little bit more time (given that there are many views, and not all of them inherit from `BaseComponentView`, ie `MinimapHudView`). IMHO, its an overkill for just this issue, so Im creating a new ticket to track it as a possible enhacement. 

If we continue creating UIs that are specific for scenarios (like more worlds UI, or maybe quests?) then it makes sense to take full on the refactor 

## How to test the changes?

1. Launch the explorer in a catalyst realm. Minimap should be there
2. Press ad unpress U to see that the minimap toggles correctly
3. Use `/changerealm redeemer.dcl.eth` and check that the minimap dissapeared and the "Go to Genesis City" button is there
4. Press ad unpress U to see that the button toggles correctly

Repeat the process but starting directly from a realm

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 23a2bb1</samp>

Refactored the logic for showing or hiding the minimap and the jump to home button based on the `DataStore` values. Created new subclasses of `ShowHideUIByTrigger` to handle the specific UI behaviors for each HUD element. Updated the `MinimapHUD.prefab` and the `HUDCommon.asmdef` files to reflect the changes.

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Common/ShowHideUIByTriggerMinimap.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Common/ShowHideUIByTriggerMinimap.cs
@@ -1,0 +1,14 @@
+using DCL;
+using UnityEngine;
+
+[RequireComponent(typeof(CanvasGroup))]
+public class ShowHideUIByTriggerMinimap : ShowHideUIByTrigger
+{
+
+    protected override void SetUIVisibility(bool isVisible)
+    {
+        if (!DataStore.i.HUDs.minimapVisible.Get())
+            return;
+        base.SetUIVisibility(isVisible);
+    }
+}
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Common/ShowHideUIByTriggerMinimap.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Common/ShowHideUIByTriggerMinimap.cs
@@ -7,11 +7,6 @@ public class ShowHideUIByTriggerMinimap : ShowHideUIByTrigger
 
     protected override void SetUIVisibility(bool isVisible)
     {
-        if (!DataStore.i.HUDs.minimapVisible.Get())
-        {
-            base.SetUIVisibility(false);
-            return;
-        }
-        base.SetUIVisibility(isVisible);
+        base.SetUIVisibility(isVisible && DataStore.i.HUDs.minimapVisible.Get());
     }
 }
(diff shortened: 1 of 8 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Common/ShowHideUIByTriggerMinimap.cs
```diff
@@ -0,0 +1,14 @@
...
+using UnityEngine;
+
+[RequireComponent(typeof(CanvasGroup))]
+public class ShowHideUIByTriggerMinimap : ShowHideUIByTrigger
+{
+
+    protected override void SetUIVisibility(bool isVisible)
+    {
+        if (!DataStore.i.HUDs.minimapVisible.Get())
+            return;
```
menduz, line 11: Maybe ensure its visibility?
```suggestion
            { base.SetUIVisibility(false);
            return; }
```

kwsskulinski (approved):
Minimap is not seen in the worlds.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/Common/ShowHideUIByTriggerMinimap.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+
+    protected override void SetUIVisibility(bool isVisible)
+    {
+        if (!DataStore.i.HUDs.minimapVisible.Get())
+        {
+            base.SetUIVisibility(false);
+            return;
+        }
+        base.SetUIVisibility(isVisible);
```
AjimenezDCL, line 15:
```suggestion
    protected override void SetUIVisibility(bool isVisible)
    {
        base.SetUIVisibility(isVisible && DataStore.i.HUDs.minimapVisible.Get(););
```

AjimenezDCL (changes requested):
Small request!

davidejensen (approved):
Nice

AjimenezDCL (approved):

PR #5014, opened 2023-04-20
chore: refactor livekit voice handler, reuse loopback interface across handlers

hugoArregui:
## What does this PR change?

Refactor livekit voice handler:
- only one is created per adapter and it's only destroyed when the adapter itself is destroyed
- use a global loopback buffer in order to reuse the PeerConnections
- take advantage of livekit auto subscription, no keeping track of streams is necessary 
- bump livekit-client
- fix error in which person A and B are in the same island, A leaves voice chat and joins again, before this change B couldn't hear A anymore, now it works

## Test

1. Open two clients with https://play.decentraland.org/?explorer-branch=voice-test
2. Test if A and B can use voice chat
3. A leave and join the Voice Chat. Try if A and B can use voice chat normally
4. B leave and join the Voice Chat. Try if A and B can use voice chat normally
5. Repeat steps 3 and 4, sometimes, to check if there is no race condition
6. Check if you can use voice chat with 3 participants

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at d72a23a</samp>

This pull request updates the browser-interface project to use the latest version of the `@dcl/protocol` dependency and the archipelago protocol for realm communication. It also refactors the voice communication logic to use a global audio stream and removes some unused or legacy code.


browser-interface/packages/shared/voiceChat/sagas.ts
```diff
@@ -89,6 +89,8 @@ function* handleConnectVoiceChatToRoom() {
           yield put(setVoiceChatError(err.toString()))
         }
       }
+    } else if (prevHandler) {
+      prevHandler.setMute(true)
```
kuruk-mm, line 93: What is the idea about this? Can we add a comment with the expected behavior?

kuruk-mm (approved):
Looks great! I didn't test it.

I added some steps to test for QA. @hugoArregui please check those, and add some tests if necessary.

Waiting for QA approval, @kwssbocian

browser-interface/packages/shared/voiceChat/sagas.ts
hugoArregui: well, since the prevHandler is not destroyed, just not used, we need to mute it in the meantime

PR #5063, opened 2023-04-26
feat: sdk7 `GltfContainerLoadingState` component

pbosio:
## What does this PR change?
add `GltfContainerLoadingState` component
fixes https://github.com/decentraland/sdk/issues/736
adr https://adr.decentraland.org/adr/ADR-215

*most of the files modifications are related to a type changed on the protocol

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 4e9a408</samp>

This pull request adds a new internal component `InternalGltfContainerLoadingState` to the ECS7 plugin, which tracks the loading and removal of `GltfContainer` components for each entity. It also updates the existing ECS7 components and systems to use this new component and changes the type of some collider fields from `int` to `uint` for consistency and performance. It modifies the assembly definitions and meta files accordingly.


unity-renderer/Assets/DCLPlugins/ECS7/Systems/GltfContainerLoadingStateSystem/GltfContainerLoadingStateSystem.cs
```diff
@@ -11,6 +11,7 @@ public class GltfContainerLoadingStateSystem
     {
         private readonly IInternalECSComponent<InternalGltfContainerLoadingState> gltfContainerLoadingStateComponent;
         private readonly IECSComponentWriter componentWriter;
+        private int timestamp = 0;
```
menduz, line 14: This timestamp should be entity aware. I think you are setting it globally, is that correct?

pbosio: that was because I wasn't storing it at the crdt state, but I'm changing it right now, so timestamp will be at entity level, being `prevComponentStateForEntity.timestamp +1`

unity-renderer/Assets/DCLPlugins/ECS7/Systems/GltfContainerLoadingStateSystem/GltfContainerLoadingStateSystem.cs
```diff
@@ -0,0 +1,59 @@
...
+        public GltfContainerLoadingStateSystem(IECSComponentWriter componentWriter,
+            IInternalECSComponent<InternalGltfContainerLoadingState> gltfContainerLoadingStateComponent)
+        {
+            this.componentWriter = componentWriter;
+            this.gltfContainerLoadingStateComponent = gltfContainerLoadingStateComponent;
+        }
+
+        public void Update()
+        {
+            var components = gltfContainerLoadingStateComponent.GetForAll();
```
pravusjif, line 24: Currently this implementation traverses all the entities with `IInternalECSComponent<InternalGltfContainerLoadingState>`, so basically EVERY entity with a GLTF (since we are adding that internal component to all).
Wouldn't it be better to just traverse the entities that have the new (**not internal**) [GltfContainerLoadingState](https://github.com/decentraland/unity-renderer/blob/781b320216e8aa4aa61bc04c4dfcc98641eab6bc/unity-renderer/Assets/Scripts/MainScripts/DCL/DecentralandProtocol/GltfContainerLoadingState.gen.cs) component instead? since those would be the only ones that we are interested in updating and reporting its loading state...

Maybe I'm missing the point here as I thought that to use the new`GltfContainerLoadingState` component from a scene with the SDK it should be added from the scene to an entity and then read it later... if we are adding that data to EVERY entity with GltfContainer why not add and update that loading state property in the GltfContainer component itself???

pbosio: > Wouldn't it be better to just traverse the entities that have the new (not internal) [GltfContainerLoadingState](https://github.com/decentraland/unity-renderer/blob/781b320216e8aa4aa61bc04c4dfcc98641eab6bc/unity-renderer/Assets/Scripts/MainScripts/DCL/DecentralandProtocol/GltfContainerLoadingState.gen.cs) component instead?

No, since scene won't be sending the `GltfContainerLoadingState` to the renderer, the renderer will be sending the `GltfContainerLoadingState` to the scene. It's not sent by request, every `GltfContainer` expect to receive a `GltfContainerLoadingState`

> Maybe I'm missing the point here as I thought that to use the newGltfContainerLoadingState component from a scene with the SDK it should be added from the scene to an entity and then read it later... if we are adding that data to EVERY entity with GltfContainer why not add and update that loading state property in the GltfContainer component itself??? and avoid having a new unneeded component.

`GltfContainer` is sent from the scene to the renderer, not the other way around
`GltfContainerLoadingState` is sent from the renderer to the scene, not the other way around
so we need two different components to avoid any write conflict.
plus, they are related but conceptually different components, one contains data of the gltf and the other the data of it loading state

pravusjif (approved):

PR #5096, opened 2023-05-02
Added tests for hiding category handling in backpack v2

### LINKED ISSUE #5094: Add tests for backpack hiding functionality
Add missing tests to cover the backpack hiding functionality

davidejensen:
## What does this PR change?

Fix #5094 
Adds avatar slot view hiding test

...

## How to test the changes?

Cannot be tested

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at a22818a</samp>

Changed the access modifiers of some fields in `AvatarSlotComponentView` and `AvatarSlotsView` classes from private to internal. This allows other classes to manipulate the avatar slot components and implement features related to hiding, showing, and synchronizing avatar slots with wearables.


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/AvatarSlot/AvatarSlotsView.cs
```diff
@@ -15,7 +15,7 @@ public class AvatarSlotsView : BaseComponentView, IAvatarSlotsView
         [SerializeField] private GameObject sectionSeparator;
 
         private readonly Dictionary<string, Transform> avatarSlotSections = new ();
-        private readonly Dictionary<string, IAvatarSlotComponentView> avatarSlots = new ();
+        internal readonly Dictionary<string, IAvatarSlotComponentView> avatarSlots = new ();
```
sandrade-dcl, line 18: I'm seeing this variable is not being using out of this class, right? Why do we need it to be internal?

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/AvatarSlot/AvatarSlotComponentView.cs
```diff
@@ -27,7 +27,7 @@ public class AvatarSlotComponentView : BaseComponentView<AvatarSlotComponentMode
         [SerializeField] internal Image focusedImage;
         [SerializeField] internal Image selectedImage;
         [SerializeField] private GameObject emptySlot;
-        [SerializeField] private GameObject hiddenSlot;
+        [SerializeField] internal GameObject hiddenSlot;
         [SerializeField] internal RectTransform tooltipContainer;
         [SerializeField] internal TMP_Text tooltipCategoryText;
         [SerializeField] internal TMP_Text tooltipHiddenText;
```
sandrade-dcl, line 33: I'm not sure if we should continue setting our private variables as `internal` only because we want to check them from our tests. If we make our tests check the functionality instead of each piece of code, we should be able to have all the internal variables as private.

Lets discuss this also with @lorux0 and lets reach a decision, because I think it is something that is happening a lot overall and it would be fine to have an agreement as team 👍

sandrade-dcl (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/AvatarSlot/AvatarSlotComponentView.cs
lorux0: I think that in this case it is correct. Because the GameObject/Components's state is the result/output of the processing of the view in each test case. Otherwise, there is nothing else to validate.
It is slightly different when we try to expose internal methods or internal fields just for the tests, because they are not part of the validation result.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/AvatarSlot/AvatarSlotsView.cs
lorux0: +1

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/Tests/AvatarSlotComponentViewShould.cs
```diff
@@ -121,5 +121,59 @@ public void SetTooltipText2()
             Assert.AreEqual(avatarSlot.tooltipCategoryText.text, $"{TEST_CATEGORY}");
             Assert.False(avatarSlot.tooltipHiddenText.gameObject.activeInHierarchy);
         }
+
+        [Test]
+        public void SetHiddenBy()
+        {
+            avatarSlot.model = new AvatarSlotComponentModel()
```
lorux0, line 128: We should be using a method like `avatarSlot.SetModel(...)` which internally does the `RefreshControl()` like the rest of the UI components. Goes for the rest of the tests aswell

lorux0 (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/AvatarSlot/AvatarSlotsView.cs
davidejensen: Done :)

sandrade-dcl (approved):

lorux0 (approved):

Issue #4369, opened 2023-02-13
Remove duplicated assets in addressables


aixaCode:
#### Description
After moving more assets to addressables, we have noticed that some assets are duplicated
Here are the findings of a few problems:
- If a prefab is directly referenced in the scene and also added to local Addressables, it'll create duplicates. This happens because we have a pattern of using a Main prefab that references all the others. To fix this, we need to move the direct references to Asset References and load them asynchronously to avoid duplicates.
- Another problem happens with assets in the Resource folder. If a prefab uses a texture or sprite that's also used in another prefab in a bundle, it'll create duplicates. For example, the `Player.prefab` uses the sprite CursorInteractive, which is also used in the `EmbededEmote.asset`. I'm not even mentioning here atlases. We can also fix that by using asset reference, and when more assets will be moved out of Resources folder this will be improved.

#### Acceptance criteria
- [ ] Major culprits for duplication of assets are removed

PR #5131, opened 2023-05-05
Chore: move huge duplicated resrources to addressables

popuz:
## What does this PR change?
References #4369 

- created test to check size of duplicated assets between Resources and Addressables
- moved to Addressables all Resources that caused duplicates with size more than 4,35 MB 
- fixed tests

## How to test the changes?

1. Launch the explorer
2. Check Avatar editor - animations, wearables, reactions, audio, voices
3. Check that Notification HUD works as previously
4. Check that Profile HUD works as previously 
5. Check that Tutorial (old and new) works as previously

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at c48fccd</samp>

This pull request adds a new tutorial scene as an addressable asset, and updates the unit tests and the code for the Addressable Asset System. It also moves some animation files to a different folder, and uses a shorter syntax for default values.


dalkia (changes requested):
I found the following issues in webGL 

1. I have no audio in the backpack; neither voices or sound effects
2. I have no Alice audio effects when triggering the yutorial
3. The Alice tutorial is not responding to the E key on the second step

kwssbocian (changes requested):
✔️All basic UI elements of the menus appear and work the same as before. This also applies to the settings menu.
❌Some audio events missing: changing wearables (shuffling sound), character voice exclamation on wearable change (ex. "Aha!"), sounds of the robot Alice in the old tutorial
❌First prompt of the old tutorial is not interactable and does not respond to E or F keys being pressed. This makes the old tutorial not possible to progress once triggered.

PR #5138, opened 2023-05-08
fix: duplicated key exception when resolving avatar wearables

lorux0:
## What does this PR change?

Solves a duplicated key exception when solving wearables.

## How to test the changes?

1. Launch the explorer
2. Navigate through populated scenes
3. Avatars should load as expected
4. Open the backpack and change your wearables
5. The profile and the avatar should be updated correctly

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at b98ecab</samp>

Fix a bug in `WearableItemResolver.cs` that could cause an exception when adding duplicate wearableIds. Add a TODO comment to handle potential memory leaks.


davidejensen (approved):
Code looks good, will test as soon as CI finishes

PR #5139, opened 2023-05-08
chore: add `use-social-client` feature flag

Julieta11:
In the Friends saga, there are messages that should be placed behind a Feature Flag to avoid duplication of information sent to Unity.


browser-interface/packages/shared/friends/sagas.ts
```diff
@@ -644,12 +650,18 @@ function* refreshFriends() {
...
     let token = client.getAccessToken()
 
-    if (token) {
-      getUnityInstance().InitializeMatrix(token)
+    // TODO!: remove FF validation once the RPC flow is the only one
+    const isUseSocialClientEnabled: boolean = yield select(getFeatureFlagEnabled, 'use-social-client')
+    if (!isUseSocialClientEnabled) {
+      getUnityInstance().InitializeFriends(initFriendsMessage)
+
+      if (token) {
```
jmoguilevsky, line 658: This needs to be enabled when the feature flag is enabled, not the other way around

Julieta11: The description of the feature flag says: "Enables the usage of the new RPC social client to consume friends instead of using the kernel implementation". So if the flag is enabled (set to true), we shouldn't notify Unity about the initialization of friends, right?

popuz:
@lorux0 clarify the status

PR #5142, opened 2023-05-08
fix: boundaries not being in sync with the SBC

### LINKED ISSUE #4866: SCB outer algorithm not returning the correct result
When leaving and entering the Colon museum scene (23,78), the theatre is not rendered correctly

Using `DEBUG_MODE`, we detected that the SCB is considering that the theatre is out of bounds while making the outer check. Removing the outer check renders the theatre correctly.

Working with the creator, it has also been detected that removing animations from the gltf also solves the issue.

We are making a debug of the SCB algorithm to refine this situation, as it may also be affecting other scenes where objects are not visible when they should.

### LINKED ISSUE #5113: 3d models not loading into the scene

Kinerius:
### Unity updates the boundaries of colliders in the `FixedUpdate`, so we **always had incorrect values** on all collider boundaries for half a frame since the SBC worked on the Update loop.

- The `SceneBoundsChecker` immortal task now waits for the next `FixedUpdate` to re-check all entities
- The preliminary check of the `SceneBoundsChecker` now also waits for the next `FixedUpdate`
- `MeshesInfo.RecalculateBounds` is now async and waits for `FixedUpdate`
- All `SceneBoundsChecker` routines and tests are now Tasks

Fixes #5113
Fixes #4866 

## How to test the changes?

For these coordinates `-109,-7` `23,78` do:
- Set the Scene load radius to 2
- Wait for the scene to load correctly
- Move somewhere else until the target scene is unloaded
- Move back until the scene loads correctly
- Those scenes shouldn't have invisible objects compared to the initial loading. 

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneBoundariesController/SceneBoundsChecker.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneBoundariesController/SceneBoundsChecker.cs
@@ -1,7 +1,8 @@
+using Cysharp.Threading.Tasks;
 using DCL.Models;
 using System;
-using System.Collections;
 using System.Collections.Generic;
+using System.Threading;
 using UnityEngine;
 
 namespace DCL.Controllers
@@ -11,7 +12,7 @@ namespace DCL.Controllers
         private const float NERFED_TIME_BUDGET = 0.5f / 1000f;
         private const float RECHECK_BUDGET = 0.5f;
         public event Action<IDCLEntity, bool> OnEntityBoundsCheckerStatusChanged;
-        public bool enabled => entitiesCheckRoutine != null;
+        public bool enabled => isEnabled;
         public float timeBetweenChecks { get; set; } = RECHECK_BUDGET;
         public int entitiesToCheckCount => entitiesToCheck.Count;
 
@@ -21,7 +22,6 @@ namespace DCL.Controllers
         private HashSet<IDCLEntity> checkedEntities = new ();
         private HashSet<IDCLEntity> persistentEntities = new ();
         private ISceneBoundsFeedbackStyle feedbackStyle;
-        private Coroutine entitiesCheckRoutine;
         private float lastCheckTime;
 
         private Service<IMessagingControllersManager> messagingManagerService;
@@ -29,6 +29,8 @@ namespace DCL.Controllers
 
         private bool isNerfed;
         private bool isInsistent;
+        private bool isEnabled;
+        private CancellationTokenSource cancellationTokenSource = new ();
 
         public void Initialize()
         {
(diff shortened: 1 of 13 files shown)


AjimenezDCL (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneBoundariesController/SceneBoundsChecker.cs
```diff
@@ -179,16 +187,28 @@ public void AddEntityToBeChecked(IDCLEntity entity, bool isPersistent = false, b
...
+
+            // The outer bounds check is cheaper than the regular check
+            RunEntityEvaluation(entity, onlyOuterBoundsCheck: true);
+
+            // No need to add the entity to be checked later if we already found it outside scene outer boundaries.
+            // When the correct events are triggered again, the entity will be checked again.
+            if (!isInsistent && !isPersistent && !entity.isInsideSceneOuterBoundaries)
+                return;
+
+            AddEntity(entity, isPersistent);
```
popuz, line 207: I would revert the if here to be om positive side (and reduce lines number :) )

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneBoundariesController/Tests/SBC_Asserts.cs
```diff
@@ -20,7 +22,7 @@ namespace SceneBoundariesCheckerTests
 {
     public static class SBC_Asserts
     {
-        public static IEnumerator EntitiesAreBeingCorrectlyRegistered(ParcelScene scene)
+        public static async Task EntitiesAreBeingCorrectlyRegistered(ParcelScene scene)
```
popuz, line 25: Is it possible to pass cToken and handle cancelations for all these tasks (maybe at least in some Dispose?)

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneBoundariesController/Tests/SBC_Asserts.cs
```diff
@@ -42,34 +44,36 @@ public static IEnumerator EntitiesAreBeingCorrectlyRegistered(ParcelScene scene)
...
         }
 
-        public static IEnumerator PShapeIsInvalidatedWhenStartingOutOfBounds(ParcelScene scene)
+        public static async Task PShapeIsInvalidatedWhenStartingOutOfBounds(ParcelScene scene)
         {
             var boxShape = TestUtils.CreateEntityWithBoxShape(scene, new Vector3(20, 2, 20));
-            yield return null;
+
+            await UniTask.WaitForFixedUpdate(); // preliminary check
+            await UniTask.WaitForFixedUpdate(); // immortal process catches up
```
popuz, line 71: I would suggest to group it in one `WaitForTwoFixedUpdates()` method, since you used it a lot

popuz (approved):
Great work 💪Some minor suggestions ;)

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneBoundariesController/Tests/SBC_Asserts.cs
Kinerius: Ah yes that's smart

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneBoundariesController/Tests/SBC_Asserts.cs
Kinerius: All those tasks in SBC_Asserts are being used by 2 different test fixtures, we don't handle cancellations in tests at least it explicitly required

PR #5197, opened 2023-05-16
test: separated downloading and creation in map chunk loading

### LINKED ISSUE #5188: Investigate map loading process that could be wasting time
## Problem 

During the profiling session we found out that when we load the map we are

- Downloading a few chunks
- Processing those few chunks
- Download more chunks
- Loop

If we analyze this image from the Firefox debugger, we can see 5 different columns of blue bars that are the network requests towards those chunk images, when we finish the first batch, we create those sprites and move on to download more.

## Solution

We should try downloading all those chunks first and then process all of them, this should unlock the web requests limit and make the loading a bit faster.

Kinerius:
## What does this PR change?

Separated the logic that downloads the map chunks and creates the sprite, to avoid halting the requests too much.

Fixes #5188

## Our Code Review Standards


Kinerius:
This didnt work as I've expected

PR #5292, opened 2023-06-01
fix: remove infinite floor for worlds

### LINKED ISSUE #5253: Disable Infinite Floor in Worlds
### Description
The infinite floor doesn't make much sense to be enabled in worlds, we should disable it by default.

### Deliverables
- [ ] Disable infinite floor in worlds.

### Notes
We already have code to disable specific features for worlds so we should hook into the same logic.

dalkia:
## What does this PR change?

Fixes #5253 

- Responsibility of changing the infinite map image now is handled by the realm change. This could offer the possibility to add infinite floors in worlds as well. 

- Nullifies the image of the infinite floor when going into a world. This shows the orange plaza floor for any given position. The other option was to turn off the infinite floor quad, but it gave a feeling that everything was floaty; so for this PR the orange floor is presented as an option.

- Refactors and corrects some test that weren't properly implemented

## How to test the changes?

1. Launch the explorer in a catalyst realm. Check that the infinite floor work as expected
2. Go to a world. Check that you only have orange floor.
3. Go to a catalyst realm. Check that the infinite floor is back 
4. Do the same process but starting from q world. Check that the floor is orange, go to catalyst and check for the infinite floor, then go back to world and check that the floor is orange.

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 90e6e4e</samp>

This pull request implements a new feature that allows the infinite floor to show different textures depending on whether the player is in a world or a catalyst realm. It does so by adding a new `RealmsInfiniteFloorModifier` class that sets the appropriate textures from the `DataStore_HUDs` to the infinite floor when the realm changes. It also refactors some tests and modifies the `MapTexturePlugin` and the `InfiniteFloorController` to use the new variables from the `DataStore_HUDs`.


AjimenezDCL (approved):

aixaCode (approved):

kwsskulinski (approved):
No issues have been found.

PR #5297, opened 2023-06-02
fix: close searchbox close icon when thext is empty

sandrade-dcl:
## What does this PR change?
The close icon button in our searchbox UI component should be hidden when the input text is empty.

image

## How to test the changes?
1. Launch the explorer.
2. Open the backpack.
3. Notice the searchbox component (in the top right side of the screen) is not showing the close icon.
4. Input some text in the searchbox.
5. Notice the close icon appears after the search.
6. Clean the searchbox text.
7. Notice the close icon is hidden.

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 576fc6d</samp>

This pull request improves the UI and testing of the `SearchBarComponentView` class. It fixes a bug with the clear search button visibility and adds a new test case for different input field states.

### DIFF
--- a/unity-renderer/Assets/UIComponents/Scripts/Components/SearchBar/SearchBarComponentView.cs
+++ b/unity-renderer/Assets/UIComponents/Scripts/Components/SearchBar/SearchBarComponentView.cs
@@ -197,7 +197,7 @@ public class SearchBarComponentView : BaseComponentView, ISearchBarComponentView
 
     internal void SetSearchMode()
     {
-        clearSearchButton.gameObject.SetActive(true);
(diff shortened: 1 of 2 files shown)


Kinerius (approved):

kwssbocian (approved):
✔️Search box component is present in the top-right section of BackpackV2 and is shown without X close button by default. Icon appears only when the box contains text.
✔️Putting in text in the searchbox causes X close icon to appear and the wearables to be filtered accordingly.
✔️Clearing text input causes the icon to disappear. Also clicking the X icon causes the text to be cleared. In both cases wearable filtering goes away as a result.

PR #5317, opened 2023-06-06
style: backpack textures optimised

RominaMarchetti:
@popuz @sandrade-dcl I updated the of the backpack which is now a single texture.
Additionally, I turned off the setting Read/Write in all the backpack textures.


RominaMarchetti:
I forgot to turn it off in that texture D: Thanks for fixing it ;)

Regarding compression: actually, the .jpeg file size is 54kb and the best I could make once imported into unity is 7.3MB sacrificing quality using selecting a minor Max Texture Size. The sum of the previous files was 14.1 MB so there is an improvement but If there is a way to make it better I don't know it :(

popuz (changes requested):
I am not sure that we can compress it more. Should be good enough for now. Probably we can set `AlphaSource` to `None` in this big texture, since it not using any alpha. 
I saw just small issues with some icons not being dividable by 4. Can we make them something like 64x64 instead of 52x50? You can find them by scouting yellow warnings in their settings
image

RominaMarchetti:
@popuz after changing its size for making it multiple of 4 the size of the file increases but I ignore if anyways more performant. Do you know how to measure that?

Screenshot 2023-06-07 at 10 05 44

popuz:
> With an atlas the draw calls number is smaller but the `saved by batching` triplicates. Is it better anyways? I don't know what the saved by batching means.
> 
> Without atlas WithoutAtlas
> 
> With atlas WithAtlas

yes, both a better - less draw calls and more saved by batching 💪

popuz:
> Okay, pals! I'm waiting your approves now 👯

I am not sure that you need to include everything in this atlas. @lorux0 probably meant only small icons, but not background elements. So I suggest to move all these small icons in the sub-folder, from which create an atlas. And UI->BackpackV2 folder will only contain atlas itself + background images (that already dividable by 4, so we good there)

You can press this rotated arrows to re-request the review ;) 
image

RominaMarchetti:
Rsult after removing _only_ the background texture from the atlas `225`
Screenshot 2023-06-07 at 14 35 01

Result after removing everything but icons `240`
Screenshot 2023-06-07 at 14 30 47

popuz:
> 240

I wouldn't say it is an equal comparison, because you have 30 Visible skinned meshes in second screenshot (probably players avatars). However, for me both are fine.

PR #5355, opened 2023-06-15
fix: failing requests because of json conversion incompatibility

Kinerius:
## What does this PR change?

We changed the default Json deserializer for the one from Newtonsoft.

We had to fix a few DTO objects that used unity attributes

## How to test the changes?

https://play.decentraland.org/?explorer-branch=fix/json-req

Names and Lands should appear correctly in the player's passport

## Our Code Review Standards


popuz (approved):

kwssbocian (approved):
✔️For users that own Names and LANDs these are correctly being shown when viewing passports of those users.

PR #5356, opened 2023-06-15
fix: avatar animations in backpack & passport

Kinerius:
## What does this PR change?

Added an old code that made sure that the avatar animator had its correct target, this workaround should be removed once we get rid of the old ABs

## How to test the changes?

- test in https://play.decentraland.org/?explorer-branch=fix/avatar-preview
- Open backpack and see that the character animates correctly
- Open passport and see that the character animates correctly

## Our Code Review Standards


sandrade-dcl (approved):

popuz (approved):

kwssbocian (approved):
✔️Avatar previews are correctly animated in Passport view and inside Backpack. T-posing of avatars no longer occurs.

PR #5384, opened 2023-06-23
chore: clean console

Kinerius:
## What does this PR change?

Removed a few `OperationCancelledExceptions` in an effort to fix https://github.com/decentraland/unity-renderer/issues/5344

Removed warnings about GLTF not supporting certain features in prod.

Fixed a Missing Reference Exception at the culling controller

## Our Code Review Standards

### DIFF
--- a/unity-renderer/Assets/Rendering/Culling/CullingController.cs
+++ b/unity-renderer/Assets/Rendering/Culling/CullingController.cs
@@ -161,7 +161,7 @@ namespace DCL.Rendering
 
             foreach (Renderer r in renderers)
             {
-                if (!r)
+                if (r == null)
                     continue;
 
                 if (Time.realtimeSinceStartup - currentStartTime >= CullingControllerSettings.MAX_TIME_BUDGET)
--- a/unity-renderer/Assets/Rendering/Culling/CullingController.cs
+++ b/unity-renderer/Assets/Rendering/Culling/CullingController.cs
@@ -161,7 +161,7 @@ namespace DCL.Rendering
 
             foreach (Renderer r in renderers)
             {
-                if (r == null)
+                if (!r)
                     continue;
 
                 if (Time.realtimeSinceStartup - currentStartTime >= CullingControllerSettings.MAX_TIME_BUDGET)
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/GLTFast/AssetPromise_GLTFast_Loader.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/GLTFast/AssetPromise_GLTFast_Loader.cs
@@ -5,8 +5,10 @@ using DCL.GLTFast.Wrappers;
 using GLTFast;
 using GLTFast.Logging;
 using GLTFast.Materials;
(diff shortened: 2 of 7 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/GLTFast/AssetPromise_GLTFast_Loader.cs
```diff
@@ -143,6 +143,50 @@ private async UniTaskVoid ImportGltfAsync(Action onSuccess, Action<Exception> on
...
+    {
+        public LogCode LastErrorCode { get; private set; }
+
+        public void Error(LogCode code, params string[] messages)
+        {
+            LastErrorCode = code;
+            Debug.LogError(LogMessages.GetFullMessage(code, messages));
+        }
+
+        public void Warning(LogCode code, params string[] messages)
```
popuz, line 156: you can try to use `[Conditional("UNITY_EDITOR")]` attribute in these cases. It will be almost the same. Some info about it:

> The [Conditional] attribute in C# is an feature that enables developers to control whether the call to a particular method should be included or excluded at compile time, based on the presence or absence of a specified preprocessor symbol. 
This is extremely useful in cases where you have code that is only relevant in debug mode or during testing, such as writing debug logs to the console.

The main points to remember about the Conditional attribute are:
- It only affects whether calls to the method are included, not whether the method itself is included in the compiled code.
- The method that has the Conditional attribute must return void.
- You can apply the Conditional attribute to a method or attribute class, but not to an instance constructor, destructor, operator, or get or set property accessor.
- Any arguments passed to a method that is conditionally excluded are still fully compiled and executed; the call to the method is the only thing that is excluded.

unity-renderer/Assets/Rendering/Culling/CullingController.cs
```diff
@@ -161,7 +161,7 @@ internal IEnumerator ProcessProfileWithEnabledCulling(CullingControllerProfile p
 
             foreach (Renderer r in renderers)
             {
-                if (r == null)
+                if (!r)
```
popuz, line 164: please revert it. _**Don't omit comparison to null**_ is in our [code-guidlines](https://github.com/decentraland/unity-renderer/blob/dev/docs/style-guidelines.md#general)

popuz (approved):
less noise in Sentry, nice 💪 Left 1 cosmetical request (code style) and 1 cosmetical suggestion (that one is optional)

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/AssetManager/GLTFast/AssetPromise_GLTFast_Loader.cs
Kinerius: I was not aware of that Conditional Attribute, It also leaves the code more readable, nice suggestion! 👍

0xD-Fabio: Amazing stuff! Didn't know either!

PR #5389, opened 2023-06-27
chore: bump libs

Julieta11:


Julieta11:
**catalyst-client dependency update.**

closing this PR as: we tried to address the broken catalog saga, specifically the fetches of wearables and emotes.

It turns out that the newest version of the lib no longer includes the includeDefinitions param in the endpoints, which is currently in the process of being deprecated. So we tried to manually fetch everything including the "to be deprecated" includeDefinitions param. Anyway, we realized that we have this huge saga doing weird requests with lots of filters and it'd be a waste of time to try to make them manually and in kernel. Instead we think we should take time to understand the requests and make them properly, and we could even try to include all in a Separate initiative.

The motivation behind updating the lib is related to the entity building process, specifically for the use of EntityType.OUTFITS (which is defined on newer versions of the lib). Now, the deployment of Outfits works by utilizing 'as unknown' to build the entity:
 
```
const entity = {
    type: EntityType.OUTFITS,
    pointers: [identity.address + ':outfits'],
    files: contentFiles,
    metadata: outfits
  } as unknown as BuildEntityOptions
```

The thing is: I'll leave the 'as unknown' and we'll try to address all the fetches as a separate initiative.

PR #5415, opened 2023-06-30
style: fix profile card scroll bar texture and collectibles scroll view mask

RominaMarchetti:
This PR fixes
- Texture used for the handle of the scroll bar and erases the tint animation.
- Collectibles section scroll view max heigh size increase to reduce the gap between the section buttons and the content bellow.

BEFORE
Screenshot 2023-06-30 at 15 12 01

AFTER
Screenshot 2023-06-30 at 15 15 28

### **How to test**
1- Go to https://play.decentraland.org/?explorer-branch=style/ProfileCardPolish
2- Open the profile card of any user, ideally with many collectibles and with a wide description. 
3- Check the scroll bar doesn't look clunky or has any tint animation with mouse interaction, in both sections.
4- Go to the `collectibles` section and check the scroll bar is anchored in X axis in the same position than in `About`.


kwssbocian (approved):
✔️Scroll max height and gap between section buttons and content has been reduced.
✔️No tint animation occurs for scroll bar on mouse over or interaction.
✔️Scroll bar is anchored in the same position along X axis for both About and Collectibles sections. Scroll boundaries are consistent for both sections.

Kinerius (approved):

PR #5425, opened 2023-07-03
fix: request profile from unknown peer if profile announcement was missed

hugoArregui:
## What does this PR change?

I saw sometimes you join a world and don't  see people around you, when joining a realm we send a profile announcement message with the version of the profile we are currently using. If for some reason that message is lost, a given peer won't request the profile again although other messages arrive and the original peer will be forever unknown. This is a pragmatic fix, if you get a position message from an unknown peer, request their profile.

## How to test the changes?

There is no specific way to reproduce this issue (specially if this fix actually works), but for QA this should be a basic comms test: go to the same place and you should be able to see people, talk to them, chat with them.

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 79c7925</samp>

This pull request enhances the profile synchronization feature for users in the same comms room. It modifies `handlers.ts` and `peers.ts` to send and receive profile requests and updates more efficiently.

### DIFF
--- a/browser-interface/packages/shared/comms/handlers.ts
+++ b/browser-interface/packages/shared/comms/handlers.ts
@@ -54,7 +54,7 @@ export async function bindHandlersToCommsContext(room: RoomConnection) {
   pingRequests.clear()
 
   // RFC4 messages
-  room.events.on('position', processPositionMessage)
+  room.events.on('position', (e) => processPositionMessage(room, e))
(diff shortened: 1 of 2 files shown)


kwssbocian (approved):
✔️During testing other avatars in the same place were present visible and interactable. Communication and messages went through without issues.

pentreathm (approved):

PR #5444, opened 2023-07-05
feat: portable experiences confirmation

lorux0:
## What does this PR change?

Adds a confirmation popup whenever a portable experience attempts to start. It will only start if the user accepts. It also allows to save your response by portable experience id.
Implements the UI hiding feature for sdk7 scenes. Required for portable experiences.
Toggled by a feature flag: `px_confirm_enabled`.

Screen Shot 2023-07-11 at 16 18 37

## How to test the changes?

1. Go to: https://sdk-test-scenes.decentraland.zone/?position=8%2C8&explorer-branch=feat%2Fportable-experiences&ENABLE_px_confirm_enabled=&realm=LocalPreview
2. Spawn a portable experience.
3. Check that the popup is shown with the correct information.
4. A PX may contain permissions, icon, name, description. Everything should be correctly in place.
5. Try accepting or rejecting.
6. Experience enabling/disabling the "dont ask again" toggle.
7. Open the portable experiences viewer. The current PXs should be displayed correctly in its enabled/disabled status.
8. Try disabling/enabling the UI from the PX viewer.
9. Try disabling/enabling the PX from the PX viewer.

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at c2a9dd6</samp>

This pull request adds the feature of spawning portable experiences from ENS names in the browser interface and the unity renderer. It updates the dependencies of `@dcl/protocol` and `@dcl/scene-runtime` to use branch URLs, and adds some validation, error handling, and logging for the ENS feature. It also adds a new UI component called `ExperiencesConfirmationPopup`, which shows a confirmation dialog for the user before loading a portable experience. It implements the logic and the state management for the confirmation popup, and exposes some functions and events for the communication between the browser interface and the unity renderer.


unity-renderer/Assets/Scripts/MainScripts/DCL/DataStore/DataStore_HUDs.cs
```diff
@@ -34,7 +34,8 @@ public class DataStore_HUDs
         public readonly BaseVariable<bool> shouldShowNotificationPanel = new BaseVariable<bool>(true);
         public readonly BaseVariable<Transform> notificationPanelTransform = new BaseVariable<Transform>(null);
         public readonly BaseVariable<Transform> topNotificationPanelTransform = new BaseVariable<Transform>(null);
-        public readonly BaseVariable<bool> isSceneUIEnabled = new BaseVariable<bool>(true);
+        public readonly BaseVariable<bool> isCurrentSceneUiEnabled = new BaseVariable<bool>(true);
+        public readonly BaseDictionary<int, bool> isSceneUiEnabled = new ();
```
pbosio, line 38: quick questions, sorry, I'm having troubles to understand it.
why are 2 variables required to handle scene's ui visibility?
where or under what conditions is `BaseDictionary<int, bool> isSceneUiEnabled` being modified? I couldn't find it

lorux0: At first, i tried setting `isCurrentSceneUiEnabled` but only works for the current scene. It doesn't work for portable experiences, it seems they don't become the current scene. In order to disable/enable the UI for the PX specifically, i need to find the scene's root element in the hierarchy.
The variable is being set in `ExperiencesViewerComponentController` whenever the user clicks the hide/show portable experience UI button.

pbosio: I see. Thanks.
I would suggest using just one variable to toggle ui visibility, `isCurrentSceneUiEnabled` could iterate loaded scenes and set the visibility value for non-global scenes and also set them when scenes a new scene is loaded.
But I don't feel that should be a blocker, so I'll leave the suggestion at your discretion if you see it fit and if you got the time

pbosio (approved):

browser-interface/packages/shared/portableExperiences/reducer.ts
```diff
@@ -29,6 +29,8 @@ export function portableExperienceReducer(
     return state
   }
 
+  console.log(action)
```
AjimenezDCL, line 32: This should be removed

kwsskulinski (changes requested):
Most of the time no confirmation popup message appears after spawning the PX. When the confirmation message does appear the image does not. Also, the PX spawns even after you select 'Don't allow'. If you untick the 'Don't ask me again', popup does not appear.
Additionally pressing 'F' to kill the PX spawns next one without killing previous.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/ExperiencesConfirmation/ExperiencesConfirmationPopupController.cs
```diff
@@ -0,0 +1,120 @@
...
+        private void OnConfirmRequested(ExperiencesConfirmationData current, ExperiencesConfirmationData previous)
+        {
+            string pxId = current.Experience.ExperienceId;
+
+            ExperiencesConfirmationData.ExperienceMetadata metadata = current.Experience;
+
+            experienceId = pxId;
+            acceptCallback = current.OnAcceptCallback;
+            rejectCallback = current.OnRejectCallback;
+
```
AjimenezDCL, line 78: It seems this could be simplified using an async approach:

OnConfirmationRequest should spawn a task that: shows the UI and wait for either accept or reject (UniTaskCompletionSource should help you a lot). It would allow you to remove those ugly `acceptCallback` and `rejectCallback` variables.

AjimenezDCL (changes requested):
Looks good! left a small review

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/ExperiencesConfirmation/ExperiencesConfirmationPopupController.cs
lorux0: Did a refactor from callback management to async task. Not sure if its simpler but follows our standards

browser-interface/packages/shared/portableExperiences/reducer.ts
lorux0: Done

unity-renderer/Assets/Scripts/MainScripts/DCL/DataStore/DataStore_HUDs.cs
lorux0: I did analyze this solution and i think that the current one is better. Its harder to get a consistent state of the visibility of each scene by managing all of them in `isSceneUiEnabled`. Now i can override the visibility state of a specific scene. If by any means `isCurrentSceneUiEnabled` changes, its more complicated to keep the visibility state of a PX.

lorux0:
> Most of the time no confirmation popup message appears after spawning the PX.
> Also, the PX spawns even after you select 'Don't allow'.
> If you untick the 'Don't ask me again', popup does not appear.

Could you provide a video and test steps?

> When the confirmation message does appear the image does not
Fixed.

> Additionally pressing 'F' to kill the PX spawns next one without killing previous.
This is not a problem of this feature, its a problem of the scene.

AjimenezDCL (approved):

PR #5470, opened 2023-07-13
fix: mentions are not always working correctly

### LINKED ISSUE #4861: [QA] @Mentions issue for wallet logins on webGL
Description:

On web @mentions for the wallet users do not show the messages with their name as highlighted in yellow and there is no @ icon in the conversation list for the channel the mention was posted in and also in the notifications window. It appears as just a normal message.

Reproduction rate: 

5/5

Steps to reproduce:    

1. Launch DCL on webGL. 
2. Mention a wallet user on ~nearby chat.
3. Observe the issue.
 
Results:    

Mentions are displayed as normal messages.
 
Expected results:

Mentions are highlighted in yellow.
 
Notes:    
- Issue occurs for production. 

sandrade-dcl:
## What does this PR change?
Fix #4861 

The background color of the chat entry was yellow for ALL mentions instead of only for the messages that contains a mention of OUR OWN USER.

image

## How to test the changes?
1. Launch 2 explorers with 2 different users (it doesn't matter if one of them is a guest).
2. Open the chat in both browsers (for example `nearby`).
3. With one of the users, mention to the other one.
4. Check that the color of the sent message is green (instead of yellow).
5. From the browser of the other user, check that the received message is yellow.
6. Do the same but starting the process from the other user and check that it works in the same way.

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 2699c73</samp>

Fix chat entry background color issue when hovering over mention links. Reset the color in `DefaultChatEntry.cs` after link detection.

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/DefaultChatEntry.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/ChatWidgetHUD/DefaultChatEntry.cs
@@ -38,7 +38,6 @@ namespace DCL.Chat.HUD
         [SerializeField] private MentionLinkDetector mentionLinkDetector;
         [SerializeField] private Color autoMentionBackgroundColor;
         [SerializeField] private Image backgroundImage;
-        [SerializeField] internal UserProfile ownUserProfile;
 
         private float hoverPanelTimer;
         private float hoverGotoPanelTimer;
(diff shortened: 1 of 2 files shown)


kwssbocian (approved):
✔️From own perspective all messages, including mentions, are now colored green.
✔️From peer perspective only the messages mentioning that user are colored yellow. Mentions of other users are colored white and therefore appear like any other normal message.

AjimenezDCL (approved):

PR #5476, opened 2023-07-13
style: backpack background texture

RominaMarchetti:
This PR optimizes the backpack background texture. Now it is power of two, so now the compression should be properly applied.

Before: 2048x1248
Now: 2048x1024


kwssbocian (approved):
✔️Backpack background texture present and appears correctly. No other issues observed

sandrade-dcl (approved):

PR #5484, opened 2023-07-14
feat: portable experiences scene feature toggles

lorux0:
## What does this PR change?

Adds the support of the current portable experience feature toggle configuration:
```
"featureToggles": {
	"portableExperiences": "enabled" | "disabled" | "hideUi" 
}
```

Supports both sdk6 & sdk7.

- enabled: allows all portable experiences in the scene
- disabled: denies all portable experiences from running in the scene
- hideUi: disables the user interface only for the running portable experiences in the scene

## How to test the changes?

1. Configure your `scene.json` with a PX feature toggle
2. Go to https://sdk-test-scenes.decentraland.zone/?position=8%2C8&explorer-branch=feat%2Fpx-scene-permissions&ENABLE_px_confirm_enabled=&realm=LocalPreview
3. Check that the described behaviour works as expected

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 6e2c550</samp>

This pull request adds two new plugins to the `PluginSystem` that allow disabling the portable experiences feature and hiding the UI of portable experiences scenes based on feature toggles. It also updates the `ECSScenesUiSystem` to respect the hide UI feature toggle and cleans up some code in the `ExperiencesViewerComponentController`.


unity-renderer/Assets/DCLPlugins/ECS7/Systems/ScenesUiSystem/ECSScenesUiSystem.cs
```diff
@@ -378,6 +387,22 @@ private void SetSceneUiVisibility(int sceneNumber, bool enabled)
                 new StyleEnum<DisplayStyle>(enabled ? DisplayStyle.Flex : DisplayStyle.None);
         }
 
+        private void UpdatePortableExperiencesVisibilityByFeatureToggle(IParcelScene currentScene, IEnumerable<IParcelScene> scenes)
```
pbosio, line 390:
```suggestion
        private void UpdatePortableExperiencesVisibilityByFeatureToggle(IParcelScene currentScene, IReadOnlyList<IParcelScene> scenes)
```
I would prefer to avoid any `IEnumerable` or `foreach` iteration if it can be avoided and in this scenario it looks like we could use `IReadOnlyList`and the good old `for (..)`

lorux0: `foreach` operations doesnt generate any GC nowdays thus the performance gain would be imperceptible. Anyway the change has been applied

pbosio (approved):

PR #5523, opened 2023-07-25
fix: removing redundant calls to confirmation dialog for unfriending actions

moliva:
## What does this PR change?

This PR will only remove a couple of redundant confirmation calls that were used internally

## How to test the changes?

1. Launch the explorer
2. Unfriend someone from different ways (passport view, DM window, friends hud)
3. Everything should work as usual (this PR will only remove a couple of redundant confirmation calls that were used internally)

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 6d9dad7</samp>

The pull request modifies the `FriendsHUDController` and the `PrivateChatWindowController` to bypass the unfriend logic and log the actions instead. The purpose is to debug the friends HUD and the private chat window features.


lorux0 (approved):

kwssbocian (approved):
✔️No issues observed upon unfriending action. No differences observed when compared to already existing behavior.

PR #5531, opened 2023-07-27
fix pex spawn every time

gonpombo8:
fix avatar schema validation for links encoded

### How we handle the Global PEX
We have some PEX that are always active via variant feature flags.
This pex has some urn like: `urn:entity:someid?baseUrl=someUrlHere` and that queryparam was breaking the logic to handle the enable/disable logic for pex.
This PR add a logic to remove that queryParam so the name/id of the pex is just the urn: `urn:entity:someid`

...

## How to test the changes?

1. Launch the explorer
2. ...

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at df760f3</samp>

This pull request fixes a bug with encoded links in profiles and refactors the handling of scene URNs. It decodes the link URLs in `profileToServerFormat.ts` and removes query parameters from the scene URN in `portableExperiencesUtils.ts`.

### DIFF
--- a/browser-interface/packages/lib/decentraland/profiles/transformations/profileToServerFormat.ts
+++ b/browser-interface/packages/lib/decentraland/profiles/transformations/profileToServerFormat.ts
@@ -23,6 +23,10 @@ export function ensureAvatarCompatibilityFormat(profile: Readonly<Avatar | OldAv
   avatarInfo.emotes = avatar?.emotes
   avatarInfo.snapshots = avatar?.snapshots
 
+  if (profile.links) {
+    profile.links.map(($) => ({ ...$, url: decodeURIComponent($.url) }))
+  }
+
   if (avatar && 'eyeColor' in avatar) {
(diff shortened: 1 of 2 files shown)


browser-interface/packages/lib/decentraland/profiles/transformations/profileToServerFormat.ts
```diff
@@ -23,6 +23,10 @@ export function ensureAvatarCompatibilityFormat(profile: Readonly<Avatar | OldAv
   avatarInfo.emotes = avatar?.emotes
   avatarInfo.snapshots = avatar?.snapshots
 
+  if (profile.links) {
```
gonpombo8, line 26: this is not related to the PX but the avatar schema validation was throwing errors because the link was `'https%3A%2F%2Fboedo.casla'` instead of `https://boedo.casla`.

davidejensen (approved):

aixaCode (approved):

PR #5548, opened 2023-08-02
feat: Integrate single sign on

fzavalia:
## What does this PR change?

Integrates Single Sign On into the explorer.
This consists on fetching and storing the user identity in an iframe instead of the current window's local storage.
Allowing the same identity to be used on different dapps without the need to ask the user for multiple signatures.

The identity is still stored in the dapp local storage as well because I didn't want to risk breaking too much stuff. If what was previously stored was the AuthIdentity and not a different object like the ExplorerIdentity I would have just removed it but considered it an unnecessary risk which can be tackled as tech debt.

## How to test the changes?

1. Launch the explorer.
2. Clear the cache, remove everything from local storage as to have a clean new session. Also clean in case you already used the SSO already, the local storage of id.decentraland.org. You can easily do so in the console in the explorer. 
image Make sure there is nothing stored there.
3. Login with your wallet. It will ask you for a signature to login as usual. Accept it.
4. Check the local storage for id.decentraland.org, you should see a key `single-sign-on-{address}` with the identity
5. Once inside the explorer refresh the page and you should login as usual without the need of signing again.
6. Log out and the storage in id.decentraland.org should dissapear. Next time you log in you will be asked to sign the identity again.

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 8011a1b</samp>

This pull request implements single sign-on (SSO) functionality for the Decentraland renderer using the `@dcl/single-sign-on-client` package. It updates the `package-lock.json` and `package.json` files to include the new dependency and the latest versions of related packages. It also modifies the browser interface and the session management modules to integrate the SSO service with the Decentraland identity system.

7. 

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 8011a1b</samp>

This pull request implements single sign-on (SSO) functionality for the Decentraland renderer using the `@dcl/single-sign-on-client` package. It updates the `package-lock.json` and `package.json` files to include the new dependency and the latest versions of related packages. It also modifies the browser interface and the session management modules to integrate the SSO service with the Decentraland identity system.


browser-interface/packages/entryPoints/index.ts
```diff
@@ -31,6 +32,8 @@ import { isWebGLCompatible } from './validations'
 declare const globalThis: { DecentralandKernel: IDecentralandKernel }
 globalThis.DecentralandKernel = {
   async initKernel(options: KernelOptions): Promise<KernelResult> {
+    SingleSignOn.init('https://id.decentraland.org')
```
fzavalia, line 35: Creates the iframe.
Maybe the url could be provided by query params

browser-interface/packages/shared/session/index.ts
```diff
@@ -23,6 +24,8 @@ export const storeSession: (session: StoredSession) => Promise<void> = async (se
 export const deleteSession = async (userId?: string) => {
   if (userId) {
     await removeFromPersistentStorage(sessionKey(userId))
+
+    await SingleSignOn.clearIdentity(userId)
```
fzavalia, line 28: Afaik the userId is the user address.
When logging out this will clear the identity from the iframe

browser-interface/packages/shared/session/index.ts
```diff
@@ -32,18 +35,35 @@ export const deleteSession = async (userId?: string) => {
...
+ * From a list of sessions return the first one with the SSO identity
+ */
+async function getSessionWithSSOIdentity(sessions: StoredSession[]) {
+  if (!sessions.length) {
+    return null
+  }
+
+  const session = sessions[0]
+
+  const ssoIdentity = await SingleSignOn.getIdentity(session.identity.address)
```
fzavalia, line 58: Attempts to obtain the session from localstorage and the iframe. 
Source of truth is the iframe so if it doesn't have an identity it return null

kwssbocian (approved):
✔️Even with a previously existing session first login attempt brings up a wallet sign pop-up.
✔️New session is stored in id.decentraland.org entry in LocalStorage. Entry corresponds with the wallet's public key
✔️Subsequent login attempts do not ask for Sign from wallet
✔️Logging out removes the id.decentraland.org entry from LocalStorage

Kinerius (approved):

PR #5549, opened 2023-08-02
fix: sdk6 colliders and gltf visibility

pbosio:
## What does this PR change?
while toggling off/on a gltf visibility, in certain scenarios, pointer event colliders where being skip and never enabled,
because it was assumed that only one collider exist and that it was the first child in the hierarchy

the following hierarchy, for example, was not working
```
gltf
-child
-child
-pointer event collider
```

## How to test the changes?
pointer events should work as usual

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at f2a3f6a</samp>

Simplified and optimized the code for shape visibility in `BaseShape.cs`.


pravusjif (approved):

kwsskulinski (approved):
Pointer events work without any issues.

PR #5625, opened 2023-08-22
fix: avatar not loading anymore after the loading process was cancelled

lorux0:
## What does this PR change?

Fixes an endless loading in the the avatar preview during the avatar first-time creation process.
This may affect any other inconsistent avatar displaying if the loading process was cancelled by any reason.

## How to test the changes?

1. Launch the explorer
2. Clean cache
3. Enter as a guest
4. Follow the avatar creation process
5. Navigate through the world and check that the avatars are being loaded correctly around you

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at d9dd33c</samp>

This pull request refactors and improves the code quality and error handling of the avatar system and the backpack editor HUD. It simplifies the cancellation logic of avatar loading, adds debug warnings and exception logging, and follows code style and formatting conventions. It also adds a new assembly reference for the avatar system.

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/AvatarSystem/Avatar.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/AvatarSystem/Avatar.cs
@@ -72,12 +72,8 @@ namespace AvatarSystem
             {
                 // Cancel any ongoing process except the current loadCancellationToken
                 // since it was provoking a double cancellation thus inconsistencies in the flow
-                status = IAvatar.Status.Idle;
-                avatarCurator?.Dispose();
-                loader?.Dispose();
-                visibility?.Dispose();
-                lod?.Dispose();
-                gpuSkinningThrottlerService?.Unregister(gpuSkinning);
+                // TODO: disposing collaborators is an anti-pattern in the current context. Disposed objects should not be reused. Instead all collaborators should handle OperationCancelledException by their own so the internal state is restored
+                CancelAndRestoreOngoingProcessesExceptTheLoading();
 
                 throw;
             }
@@ -85,12 +81,8 @@ namespace AvatarSystem
             {
                 // Cancel any ongoing process except the current loadCancellationToken
                 // since it was provoking a double cancellation thus inconsistencies in the flow
-                status = IAvatar.Status.Idle;
-                avatarCurator?.Dispose();
-                loader?.Dispose();
-                visibility?.Dispose();
-                lod?.Dispose();
-                gpuSkinningThrottlerService?.Unregister(gpuSkinning);
+                // TODO: disposing collaborators is an anti-pattern in the current context. Disposed objects should not be reused. Instead all collaborators should handle OperationCancelledException by their own so the internal state is restored
+                CancelAndRestoreOngoingProcessesExceptTheLoading();
 
                 Debug.Log($"Avatar.Load failed with wearables:[{string.Join(",", wearablesIds)}] " +
                           $"for bodyshape:{settings.bodyshapeId} and player {settings.playerName}");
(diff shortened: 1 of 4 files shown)


anicalbano (approved):
✅ PR reviewed by QA. 
✅ The avatar is loading in each attempt to start a new creation flow.
📓 Tested in Windows and mac, Chrome, Firefox and Edge.

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/BackpackEditorHUDController.cs
```diff
@@ -103,14 +103,15 @@ public class BackpackEditorHUDController
 
             view.SetOutfitsEnabled(dataStore.featureFlags.flags.Get().IsFeatureEnabled("outfits"));
             SetVisibility(dataStore.HUDs.avatarEditorVisible.Get(), saveAvatar: false);
-            view?.SetVRMButtonActive(this.dataStore.featureFlags.flags.Get().IsFeatureEnabled("vrm_export"));
-            view?.SetVRMButtonEnabled(true);
-            view?.SetVRMSuccessToastActive(false);
+            view.SetVRMButtonActive(this.dataStore.featureFlags.flags.Get().IsFeatureEnabled("vrm_export"));
```
dalkia, line 106: Is it safe to remove this null-checks? I could not find the relationship with the PR

unity-renderer/Assets/Scripts/MainScripts/DCL/AvatarSystem/Avatar.cs
```diff
@@ -57,37 +59,46 @@ public class Avatar : IAvatar
...
             {
-                Dispose();
+                // Cancel any ongoing process except the current loadCancellationToken
+                // since it was provoking a double cancellation thus inconsistencies in the flow
+                status = IAvatar.Status.Idle;
+                avatarCurator?.Dispose();
+                loader?.Dispose();
+                visibility?.Dispose();
+                lod?.Dispose();
+                gpuSkinningThrottlerService?.Unregister(gpuSkinning);
```
dalkia, line 80: I would suggest extracting this piece of code to a method, since its repeated in three places in this class.

dalkia (approved):
Approved with minor suggestions!

unity-renderer/Assets/Scripts/MainScripts/DCL/AvatarSystem/Avatar.cs
```diff
@@ -205,9 +216,9 @@ public virtual void AddVisibilityConstraint(string key)
         public void Dispose()
```
dalkia, line 216: One additional question. Is it safe to re-dispose elements that have been disposed?

Imagine this situation. The avatar gets disposed while loading, therefore the `OperationCanceledException` is catched. But inside there, items are re-disposed again; as they were in this `Dispose()` method. Could that be an issue?

kwsskulinski (approved):
No infinite loading was seen. Avatars loaded without any issues.

Kinerius (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/BackpackEditorHUDController.cs
lorux0: Yes because we have direct access to the `view` in the lines above. Moreover the view will never be null

unity-renderer/Assets/Scripts/MainScripts/DCL/AvatarSystem/Avatar.cs
lorux0: I agree with your comment, i think its an anti-pattern. I didnt want to change it because so far it has been working like this.

lorux0: In addition, i think that the proper way of improving this is by every object (loader, visibility, etc) handling its own `OperationCancelledException` avoiding an explicit disposal call from the `Avatar` class

Issue #5561, opened 2023-08-08
[QA] Changing genders while the avatar is loading doesn't show the equipped clothes


m3taphysics:
**Description:**
If a guest starts the onboarding process and changes the avatar gender while it's loading, the standard clothes do not appear as equipped.

**Steps to reproduce:**
1. Enter Decentraland as a guest
2. Start the onboarding process
3. When reaching the avatar creation, change the gender of the avatar.
4. Observe 

**Expected results:**
The standard clothes are equipped no matter the gender of the avatar.

**Current results:**
When changing the gender, the avatar does not load the clothes and it appears to be naked.

**Repro rate:**
5/5 (100%)

**Notes:**
- This issue only happens the first time the user enters as guests.
- This issue was reproduced on Chrome, Firefox, Edge and Windows Client.

**Platforms:** 
- Web client, Windows client.

**Evidence:**

https://github.com/decentraland/unity-renderer/assets/139157886/65f3f08e-ddae-4d45-b595-dfde4fa4b56f

PR #5628, opened 2023-08-23
fix: fallback to default wearables when changing body shape

lorux0:
## What does this PR change?

Fixes this issue: https://github.com/decentraland/unity-renderer/issues/5561
Equips a set of default wearables when changing the body shape and the current wearables have incompatibility with the new body shape.

## How to test the changes?

1. Launch the explorer as a guest
2. Start the avatar creation flow
3. Change the gender
4. Observe that you should not be naked, instead a set of wearables should be equipped. Only applies for the next categories: UPPER_BODY, LOWER_BODY, FEET, HAIR, MOUTH, EYEBROWS, EYES
5. Launch the explorer as a registered user
6. Open the backpack
7. Equip wearables that are compatible with a specific gender
8. Change the gender
9. Observe that the specific gender wearables should have been replaced with default wearables. Only applies for the next categories: UPPER_BODY, LOWER_BODY, FEET, HAIR, MOUTH, EYEBROWS, EYES

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at e26ad1f</samp>

This pull request adds a feature and a test to the `BackpackEditorHUDController` class, which handles the logic of editing the avatar's wearables in the backpack HUD. The feature ensures that wearables that do not match the avatar's body shape are replaced with default ones when saving the changes. The test verifies that the feature works as expected with different body shapes and wearables.

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/BackpackEditorHUDController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/BackpackEditorHUDController.cs
@@ -6,6 +6,7 @@ using MainScripts.DCL.Components.Avatar.VRMExporter;
 using MainScripts.DCL.Controllers.HUD.CharacterPreview;
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 using System.Threading;
 using UnityEngine;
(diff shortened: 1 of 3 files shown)


anicalbano (approved):
✅ This PR has been tested and approved by QA


https://github.com/decentraland/unity-renderer/assets/139157886/d56364c5-4292-4512-9dfd-4e782d9e4745

davidejensen (approved):
code LGTM
only minor suggestion could be to have the fallback urns as constants somewhere generic so that it would be easy to update both code and tests and eventual additional usages in the future

PR #5638, opened 2023-08-25
fix: Voice chat icon display showing removed user

dalkia:
## What does this PR change?

Removes a player from the `playersSpeaking` list when it leaves the `player.otherPlayers` array.

There was an issue where, if a peer kept speaking when the user teleported away, the off message from kernel for the old previous voice chat handler never came. Therefore, the peer remained forever in the `playersSpeaking` array.

## How to test the changes?

You need two computers to test this :(. It cannot be tested only in one

1. Take two users into a location
2. Leave one speaking, teleport the other one to 0,0
3. As new peers start to appear for the teleported users, the voice icon should be off

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 2539b9c</samp>

Fix voice chat recording status bug. Reset the `recording` flag of a user in `VoiceChatWindowComponentView.cs` when they are removed from the voice chat window.


sandrade-dcl (approved):

anicalbano (approved):
✅ PR / Fix checked and approved by QA 😄

PR #5654, opened 2023-08-29
Release 2023-08-30

pravusjif:
Release for 30th August 2023


pravusjif:
Release aborted due to the new found problems regarding meshes not rendering with antialiasing off, using the new unity version

PR #5656, opened 2023-08-29
chore: SDK7 Entity Pooling

0xD-Fabio:
## What does this PR change?

...

## How to test the changes?

1. Launch the explorer
2. ...

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 13b5b88</samp>

This pull request marks several ECS components as internal, meaning that they are not sent to the scene but only used by the renderer. This is part of a larger refactoring to improve the performance and memory usage of the ECS system by reducing the number of components that are serialized and deserialized. The affected components are `Animator`, `AudioSource`, `AudioStream`, `AvatarAttach`, `AvatarModifierArea`, `AvatarShape`, `Billboard`, `CameraMode`, `CameraModeArea`, `EngineInfo`, `GltfContainer`, `GltfContainerLoadingState`, `Material`, `MeshCollider`, `MeshRenderer`, `NFTShape`, `PointerEventResult`, `PointerEvents`, `PointerLock`, and `Raycast`. The changes are applied in the corresponding `*Register.cs` files in the `unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents` folder.


unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ECSRuntime/Components/ECSComponent.cs
```diff
@@ -12,12 +12,28 @@ public class ECSComponent<ModelType> : IECSComponent
             new DualKeyValueSet<IParcelScene, long, ECSComponentData<ModelType>>(50);
 
         private readonly Func<IECSComponentHandler<ModelType>> handlerBuilder;
-        private readonly Func<object, ModelType> deserializer;
+        internal readonly Func<object, ModelType> deserializer;
 
+        internal readonly IECSComponentPool<ModelType> iEcsComponentPool;
+
+        IEcsComponentImplementation implementation;
```
mikhail-dcl, line 19: I think design-wise it's strange that the component itself controls the pool capabilities. Normally, I would suggest that a pool controls component creation. 
Moreover, we have 2 additional abstractions `IECSComponentPool<ModelType>` and `IEcsComponentImplementation` which make things more complex than they should be. Also it looks like they are connected with each other but represented by different interfaces, it looks like a red flag.

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Animator/AnimatorRegister.cs
```diff
@@ -11,7 +13,17 @@ public class AnimatorRegister : IDisposable
 
         public AnimatorRegister(int componentId, ECSComponentsFactory factory, IECSComponentWriter componentWriter)
         {
-            factory.AddOrReplaceComponent(componentId, AnimatorSerializer.Deserialize, () => new AnimatorComponentHandler(DataStore.i.ecs7));
+            var poolWrapper = new ECSReferenceTypeIecsComponentPool<PBAnimator>(
+                new WrappedComponentPool<IWrappedComponent<PBAnimator>>(10,
+                    () => new ProtobufWrappedComponent<PBAnimator>(new PBAnimator()))
```
mikhail-dcl, line 18: It can be further generalized as all PB Components are created by a default constructor. After all, if you don't require to specify default capacity pool creation can be a generic static one-liner

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ECSRuntime/Components/ECSValueTypeComponentPool.cs
```diff
@@ -0,0 +1,12 @@
+namespace DCL.ECSRuntime
+{
+    public class ECSValueTypeComponentPool<ModelType> : IECSComponentPool<ModelType> where ModelType : struct
```
mikhail-dcl, line 3: You can't really pool value types. I believe it should not never exist

mikhail-dcl (changes requested):

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ECSRuntime/Components/ECSValueTypeComponentPool.cs
0xD-Fabio: That's right, but I needed this to inherit from IECSComponentPool, because of the multiple abstraction layers already in place.
In fact, these Get and Release are never called because Internal ECSComponent (aka structs) are never called because we check if they are an ECSComponent with or without an instantiated pool.

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ECSRuntime/Components/ECSReferenceTypeComponentPool.cs
```diff
@@ -0,0 +1,29 @@
+using DCL.ECS7.ComponentWrapper;
+using DCL.ECS7.ComponentWrapper.Generic;
+
+namespace DCL.ECSRuntime
+{
+    public class ECSReferenceTypeIecsComponentPool<ModelType> : IECSComponentPool<ModelType> where ModelType : class
```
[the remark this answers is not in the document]
0xD-Fabio: I have tried to come up with a solution to merge them, but this cannot be done:
`WrappedComponentPool` is a more generic and utility class for any type that satisfies the `IWrappedComponent` interface.
`ECSReferenceTypeIecsComponentPool<ModelType>` is more specialized and acts as an adapter that translates the interface of `WrappedComponentPool<T>` into something that the rest of the ECS7 implementation can understand.

Also, keeping them separate adheres better to the Single Responsibility Principle, in my opinion.
Trying to merge them brought me on a path for a huge refactoring of the SDK7, and that should be done in a more strategic and thoughtful way by the SDK team (and probably would take me weeks, which can't fit in this shape-up initiative).

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ECSRuntime/Components/ECSComponent.cs
0xD-Fabio: The problem here is that the `ECSComponent `class handles bot value types and reference types, and doing it in a generic way is tricky.
Because of the way SDK7's architecture handles `ECSComponents`, we are currently discerning between structs and classes because of the injected `ModelType `pools.
I know this is not the optimal way to design a clean code, but I couldn't find any other solution without a significant refactor, and that would take more time.

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Animator/AnimatorRegister.cs
0xD-Fabio: PB Components have too many divergencies. I've tried to come up with a solution, but I would end up creating a static "ComponentPoolFactory" that would just shorten the Register method by one line.
Besides that, the pool capacities will be fine-tweaked in the future, because at the moment we don't know what would be the optimal amount for each kind of component.

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ECSRuntime/Components/ECSComponentImplementations/PooledComponentImplementation.cs
```diff
@@ -0,0 +1,39 @@
...
+    public void Deserialize(IParcelScene scene, IDCLEntity entity, object message)
+    {
+        // Since we are using pooling, we need to get an instance from the pool
+        ModelType modelInstance = component.iEcsComponentPool.Get();
+
+        // FD:: do we need to do further deserialization here due to message?
+
+
+        component.SetModel(scene, entity, modelInstance);
+    }
```
pbosio, line 38: where does the deserialization happen? I mean, setting the model with the message data

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/ECSRuntime/Components/ECSComponentImplementations/PooledComponentImplementation.cs
```diff
@@ -0,0 +1,39 @@
...
+    public PooledComponentImplementation(ECSComponent<ModelType> component)
+    {
+        this.component = component;
+    }
+
+    public void SetModel(IParcelScene scene, long entityId, ModelType model)
+    {
+        if(model != null)
+        {
+            component.SetModel(scene, entityId, model);
```
pbosio, line 20: how does this work? looks like `ECSComponent::SetModel` calls `PooledComponentImplementation::SetModel` which is calling `ECSComponent::SetModel`
am I missing something? 🤔

Issue #4426, opened 2023-02-16
[QA] Failure to load avatar face snapshot and wrong fallback image


xavi-dcl:
Often (quite randomly but I suspect it happens more when there are plenty of users in a scene) you click on some avatar and the snapshot is shown as in the following screenshot:

![image.png](https://images.zenhubusercontent.com/636b99a62654f2e351ceea55/20db932d-6c58-431a-83e1-cb5e8fc6c04a)

Proposed steps:
- [ ] Fix the default avatar image (replace by a grey default profile icon)
- [ ] Clarify why the snapshots are failing to load so often
- [ ] Fix the root of the issue and/or make the client more robust to load failure (e.g. implementing retries?) 

------------------

Causes of this issue:
- [x] [Wrong profile announcement through comms](https://github.com/decentraland/unity-renderer/pull/5696)
- [ ] Weird/incorrect snapshots generated & stored in Catalyst profiles - [Slack thread](https://decentralandteam.slack.com/archives/C01QVR7TJK1/p1695132203469979)
- [ ] Snapshots in profiles pointing to same empty image - [Slack thread](https://decentralandteam.slack.com/archives/C01QVR7TJK1/p1695133549785459)
- [ ] Correct snapshots in profile are not rendered properly - [Slack thread](https://decentralandteam.slack.com/archives/C01QVR7TJK1/p1695222552544669)

lorux0:
I've been debugging and looks like the profile is sent with the wrong information. The face256 snapshot is a default avatar. Looks like its not a renderer issue, but kernel or catalysts:
![Screen Shot 2023-03-20 at 17.47.03.png](https://images.zenhubusercontent.com/615af652539a0a743b87eaaa/7014cc46-13cb-4a4b-9396-7081cdfcb99c)

lorux0:
Continued further investigations. Making a request for a user, for example:
```
https://peer-wc1.decentraland.org/lambdas/profiles/0x3fbe902031290ff52bd449c3e32a75665d31ed3b?no-cache=0.9613263442949931
```
Returns the correct face256 information.
When renderer receives the update from kernel about the profile information, we get the data wrong:
![Screen Shot 2023-03-20 at 18.02.02.png](https://images.zenhubusercontent.com/615af652539a0a743b87eaaa/e96b89b5-1c23-450f-8bd6-f720620862c7)

So its gotta be something in kernel

anicalbano:
[...]

PR #5696, opened 2023-09-08
fix: set correct snapshots on comms profile messages

moliva:
Addresses #4426.

## What does this PR change?

Fixes the snapshot hashes being sent through comms messages. These were being stripped of their original snapshots (both body and face) to match the default ones that appear in the issue screenshots.

The cause of this seems to be that we are checking for an address (path or URL) or a valid IPFSv2 hash, when for the current user profile we are currently storing a base64 serialization of the snapshots. These were ultimately invalid for the above checks and where changed when announcing our own profiles to the rest of the players, thus showing the defaults instead.

## How to test the changes?

1. Launch the explorer with different accounts using this same version
2. Gather them together so they are connected through comms
3. Check that the snapshots received (as seen in the nearby chat and other interfaces) should match the ones in their profiles (for the people using *this specific version*, the other players still using the release branch won't show up properly due to the announcing issue)

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at b188db4</samp>

The pull request adds functionality to convert base64 snapshots in profiles to content files and upload them to the catalyst server. It also adds caching and validation logic for the snapshots. The pull request modifies `comms/sagas.ts`, `handleDeployProfile.ts`, and adds a new function `isBase64` to `base64ToBlob.ts`.

### DIFF
--- a/browser-interface/packages/shared/comms/sagas.ts
+++ b/browser-interface/packages/shared/comms/sagas.ts
@@ -53,6 +53,8 @@ import { RootState } from 'shared/store/rootTypes'
 import { now } from 'lib/javascript/now'
 import { getGlobalAudioStream } from './adapters/voice/loopback'
 import { store } from 'shared/store/isolatedStore'
+import { buildSnapshotContent } from 'shared/profiles/sagas/handleDeployProfile'
+import { isBase64 } from 'lib/encoding/base64ToBlob'
 
 const TIME_BETWEEN_PROFILE_RESPONSES = 1000
 // this interval should be fast because this will be the delay other people around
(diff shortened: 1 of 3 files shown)


kuruk-mm (approved):
Not tested. Code looks great

mikhail-dcl (approved):

browser-interface/packages/shared/comms/sagas.ts
```diff
@@ -370,18 +373,36 @@ function getInformationForCommsProfileRequest(state: RootState) {
   }
 }
 
-function stripSnapshots(profile: Avatar): Avatar {
+const snapshotsCache: Map<string, string> = new Map()
```
hugoArregui, line 376: feel free to ignore this comment because I'm not 100% how this works, but couldn't this cache get out of hands? entries won't be ever evevicted AFAIU, maybe use lru-cache instead?

hugoArregui (approved):

anicalbano (approved):
✅ Tested and approved by QA

image

image

browser-interface/packages/shared/comms/sagas.ts
moliva: This should not be a critical case since the items added are linked to an effective change in profile and the values stored are four strings per change (a key base64 and a string for body and face256).

I'll come back to this issue though, in a separate PR. Thanks!

PR #5699, opened 2023-09-11
Chore: remove screencapture awaited initialization

popuz:
## What does this PR change?

`InitializeAwait` from `IService` is used in `.WhenAll`. Since in `ScreencaptureService `initialization awaits other things and not needed to be ready on the first frame, I set its initialization to be not awaited

## How to test the changes?

1. Check screencapturing works as previously

## Our Code Review Standards

## Copilot summary

copilot:summary


kwssbocian (approved):
✔️Screen capture behavior works exactly like before from user perspective. No changes observed.

unity-renderer/Assets/DCLServices/ScreencaptureCamera/Service/ScreencaptureCameraService.cs
```diff
@@ -50,7 +53,7 @@ public void Dispose()
                 Object.Destroy(enableCameraButtonCanvas.gameObject);
         }
 
-        public async UniTask InitializeAsync(CancellationToken cancellationToken)
+        private async UniTask InitializeAsyncInternal(CancellationToken cancellationToken)
```
sandrade-dcl, line 56: I think it's would be better to put the `Async` world at the end of the methods. Wdyt?
```suggestion
        private async UniTask InitializeInternalAsync(CancellationToken cancellationToken)
```

sandrade-dcl (approved):
LGTM!

unity-renderer/Assets/DCLServices/ScreencaptureCamera/Service/ScreencaptureCameraService.cs
popuz: good point, especially because we have it like that in our code conventions

mikhail-dcl (approved):

PR #5702, opened 2023-09-12
feat: resolve archipelago from comms.adapter

pedrotambo:
This should work for:

- [x] new /about using `adapter` instead of `fixedAdapter`
- [x] old /about without `adapter` and `fixedAdapter` (currently catalysts)
- [x] old /about using `fixedAdapter` (currently worlds) 
- [ ] offline adapters (how can I test this?)


browser-interface/packages/shared/realm/resolver.ts
```diff
@@ -45,8 +45,22 @@ export async function adapterForRealmConfig(
     ...about.configurations
   }
 
+  if (
```
lorux0, line 48: What is this check for?

hugoArregui: sorry, left over, removing it

lorux0 (approved):

anicalbano (commented):
✅ The feature has been tested by QA with the FF. It'll be reviewed once more in prod to check the stability with a bigger influx of users.

Issue #5532, opened 2023-07-27
[QA] Participating in games at Tominoya causes persistent periodic freezing


kwssbocian:
Description:
Participating in various activities at Tominoya scene for a while (at least 10min) results in periodic triggering of asset unloading by the client which causes a freeze every time it occurs. 
Every occurrence of unloading unused serialized files/unloading unused assets results in a noticeable freeze (`MarkObjects` is reported to be taking more than 2000ms even when reported unloaded assets is 0). After the freezing starts to occur it persists even after teleporting outside of Tominoya to an empty remote scene - it goes away only after refreshing the browser tab.
Freezes appear to be longer on lower powered machines. Tested on Win PC with i7-10700K CPU, 16GB RAM, RTX 2070 SUPER GPU

Reproduction rate:
5/5 

Steps to reproduce:    
1. Log in using a wallet account on webGL and navigate to Tominoya (-121,136) scene.
2. Open up the browser DevTools on the Console tab and play any of the available games for at least 10-15min.
3. Observe the console for occurrences of `Unloading X Unused Serialized files (Serialized files now loaded: X)
Unloading X unused Assets to reduce memory usage. Loaded Objects now: X.` warnings. These should be accompanied with a few second freeze of the renderer window.
4. Teleport away to a remote empty scene (for example 150,150). Wait some time while standing still and observe the console window. After a while freezes will continue to occur periodically.

Results:    
Playing games at Tominoya causes periodic freezes to occur for the rest of the session.

Expected results: 
Playing games at Tominoya causes no freezing to occur.

Notes:    
- Platforms: web, wallet login. 
- Issue occurs for both production and dev branches.
- Please see the attached screenshots and video for reference.
![image.png](https://images.zenhubusercontent.com/62e91e01ed547029fd717bd4/46d0e2c6-5212-4113-8560-f529ebd6706b)
![image.png](https://images.zenhubusercontent.com/62e91e01ed547029fd717bd4/7fc43ddc-9889-44ca-bf96-349dcf73dd27)
https://images.zenhubusercontent.com/337227404/d1feec77-ecd7-490d-989b-e8425312458c/2023_07_27_12_56_00.mp4
https://images.zenhubusercontent.com/337227404/ebeaecc8-a1e5-4df9-9907-31a69b231148/2023_07_27_12_58_37.mp4

davidejensen:
Current findings:
[...]

PR #5708, opened 2023-09-15
chore: remove unused scene messages pool queue

pravusjif:
* Removed SceneController's sceneMessagesPool usage, since that queue is being infinitely filled with every "entity component" message generating the memory leak reported at https://github.com/decentraland/unity-renderer/issues/5532#issuecomment-1718661066
* Increased memory manager minimum memory check, as the current explorer supports 4gbs max instead of the old 2gb (and triggering the unload of assets always introduces a big hiccup)
* updated webgl max memory size to 4096 in project settings just in case (should already be overwritten by [emscripten linker flag](https://github.com/decentraland/unity-renderer/blob/4c5f3d00a02b3a2f9a04bf76f589066817da1813/unity-renderer/Assets/Scripts/Editor/PreBuildProcessing.cs#L24))
* Minor code improvement

--------------------------

### TESTING
This PR shoul be tested as thoroughly as a Release Candidate test.

--------------------------

### <samp>🤖 Generated by Copilot at a3ca31c</samp>

This pull request refactors the scene message handling logic in the `SceneController` class and its related interfaces. It removes the unused and inefficient `sceneMessagesPool` and simplifies the creation of `QueuedSceneMessage_Scene` instances. It also cleans up some unnecessary `using` directives and improves the code style.

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneController.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneController.cs
@@ -53,24 +53,11 @@ namespace DCL
             enabled = true;
 
             DataStore.i.debugConfig.isDebugMode.OnChange += OnDebugModeSet;
-
-            // SetupDeferredRunners(); // TODO: is this needed ?? (WebGL/Desktop appears to run OK without it)
-
             DataStore.i.player.playerGridPosition.OnChange += SetPositionDirty;
             CommonScriptableObjects.sceneNumber.OnChange += OnCurrentSceneNumberChange;
 
             Environment.i.platform.updateEventHandler.AddListener(IUpdateEventHandler.EventType.Update, Update);
             Environment.i.platform.updateEventHandler.AddListener(IUpdateEventHandler.EventType.LateUpdate, LateUpdate);
-        }
-
-        private void SetupDeferredRunners()
-        {
-#if UNITY_WEBGL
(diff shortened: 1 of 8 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneController.cs
```diff
@@ -54,14 +54,11 @@ public void Initialize()
 
             DataStore.i.debugConfig.isDebugMode.OnChange += OnDebugModeSet;
 
-            SetupDeferredRunners();
+            // SetupDeferredRunners(); // TODO: is this needed ?? (WebGL/Desktop appears to run OK without it)
```
pravusjif, line 57: @Kinerius from what I tested (WebGL + Desktop) everything works fine without having this call, do you know if we really need this? if I can remove it I can also remove more code associated to that.

Kinerius (approved):
GFJ ❤️

unity-renderer/Assets/Scripts/MainScripts/DCL/WorldRuntime/SceneController.cs
Kinerius: It was needed in the past when this was the cornerstone of the SDK6, I think the old MessageQueue system has to go

Kinerius (approved):

anicalbano (approved):
✅ Reviewed by QA

popuz (approved):
Amazing work 🦅

PR #5714, opened 2023-09-15
fix: filter catalyst node on realm picking when it does not contains realm name

aleortega:
## What does this PR change?

This PR updates the criteria when picking a catalyst node through the realm-picking algorithm. From now on, all those catalysts that does not return its realm name when fetching its /about will be excluded from the candidates.

### <samp>🤖 Generated by Copilot at d189774</samp>

Filter out catalysts without realm name and add test case. The changes in `dao/index.ts` and `fetchCatalystStatuses.test.ts` improve the client's ability to select a valid catalyst by ignoring those that do not provide a realm name. The test case verifies the filtering logic.


anicalbano (approved):
✅ Approved by QA

PR #5724, opened 2023-09-20
wip

leanmendoza:


browser-interface/packages/shared/apis/host/runtime7/serialization/crdt/index.ts
```diff
@@ -0,0 +1,6 @@
+export * from './deleteComponent'
```
gonpombo8, line 1: cant we use @dcl/sdk here so we avoid duplicating all these files ? 
And use the ones that are already tested ?

gonpombo8 (commented):
what about the perf of this when we have many users on multiple scenes ? 
Does it affects in something ? It seems costly but im not sure

browser-interface/packages/shared/apis/host/runtime7/avatar/ecs.ts
```diff
@@ -0,0 +1,278 @@
...
+import { Entity } from '../engine/entity'
+import { ReadWriteByteBuffer } from '../serialization/ByteBuffer'
+import { AppendValueOperation } from '../serialization/crdt/appendValue'
+import { DeleteEntity } from '../serialization/crdt/deleteEntity'
+import { PutComponentOperation } from '../serialization/crdt/putComponent'
+
+export { PBPointerEventsResult } from '../../../../protocol/decentraland/sdk/components/pointer_events_result.gen'
+
+const MAX_ENTITY_VERSION = 0xffff
+const AVATAR_RESERVED_ENTITY_NUMBER = { from: 10, to: 200 }
```
gonpombo8, line 17: this is kind of a limitation right ? We can't support more that 190 users 🤔  ?
What when we have the sdk7 avatar scene and foreing entities ? How this two things are going to live together ?

leanmendoza: Actually, the range defined in the ADR is from 32 to 256 (224 players limit). AFAIK, the current limit in the Foundation client is 100 players. I'm not sure how realistic it is to think about more than 224 players but the ADR-245 is in `Review` state and comments and modifications are welcome.
Regarding the foreign entities, I'm not sure about how they and this are going to live together as I don't know what is the ETA to get the foreign entities a fact. The first (and original) idea was to do this with the Foreign entities (in the Protocol Squad), but due to the complexity and the scope of this, we decided to move with this approach.

Bevy and Godot implementation doesn't have a JavaScript `Scene` for avatars, and probably Explorer-Alpha goes with that approach, it'd be embedded in the explorer itself.

browser-interface/packages/shared/apis/host/runtime7/serialization/crdt/index.ts
leanmendoza: Absolutely we can, I didn't try to iterate fast, this is not going to be implemented by me, only giving a PoC to show it could work

leanmendoza:
> what about the perf of this when we have many users on multiple scenes ? Does it affects in something ? It seems costly but im not sure

Each player is only in one Scene, so the Transform is only sent to it. I don't see any perf issue but we'd need to do some tests.

PR #5732, opened 2023-09-21
fix: blocked by peer status

lorux0:
## What does this PR change?

Re-enables the propagation of the `blocked` list through comms and improves UI (user context menu & passport) flows when the peer has blocked you.

## How to test the changes?

1. Launch the explorer
2. Block a user or get blocked by a user
3. Check that the flows in the passport are correct. If the user has blocked you, you shouldn't be able to interact. If you blocked the user, you should be able to unblock it
4. Open any user context menu. If the user has blocked you, you shouldn't be able to interact, for example send a friend request. If you blocked the user, you should be able to unblock it
5. Try send messages. You shouldn't be able to receive or send to/from blocked users.

## Our Code Review Standards

## Copilot summary

copilot:summary

### DIFF
--- a/browser-interface/packages/shared/comms/sagas.ts
+++ b/browser-interface/packages/shared/comms/sagas.ts
@@ -410,6 +410,8 @@ async function stripSnapshots(profile: Avatar): Promise<Avatar> {
     ...profile,
     avatar: { ...profile.avatar, snapshots: newSnapshots as Snapshots },
     // THIS IS IMPORTANT, the blocked and muted sizes are too big for the network and are unnecesary
+    // blocked status propagation has been re-enabled so the current user knows when a peer has blocked it
+    // use case: you shouldn't be able to send messages or friend requests to peers that blocked you
     // blocked: [],
     muted: []
   }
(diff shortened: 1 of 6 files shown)


browser-interface/packages/shared/comms/sagas.ts
```diff
@@ -410,7 +410,9 @@ async function stripSnapshots(profile: Avatar): Promise<Avatar> {
     ...profile,
     avatar: { ...profile.avatar, snapshots: newSnapshots as Snapshots },
     // THIS IS IMPORTANT, the blocked and muted sizes are too big for the network and are unnecesary
-    blocked: [],
+    // blocked status propagation has been re-enabled so the current user knows when a peer has blocked it
```
hugoArregui, line 413: can't messages be ignored locally instead? I'm worried about the big warning above

hugoArregui: I meant the warning in the comment "the blocked and muted sizes are too big for the network", can we simply ignore the messages if we get one from a blocked user?

lorux0: Is not only the messages. You shouldnt be able to interact with friend requests or preview the passport of peers that blocked you. We have special states in the UI for this.

hugoArregui: ok, understood, but I don't think we can simply send this through comms, some users have thousands of addresses in the block list, and this message is broadcasted to everyone around the user.

lorux0: I will try to make a workaround by fetching the user profile from the catalyst when this information is required

lorux0: Trying to fetch the blocked list from the catalyst when needed adds too much complexity thus a mess to handle with the current architecture. A new proposal is coming on which profile information will always be fetched from the catalyst instead of being broadcasted through comms. This way would fix the problem

PR #5733, opened 2023-09-22
Chore: Impostor vulnerability prevention

0xD-Fabio:
## What does this PR change?

Adds a validation for impostors on profiles that are broadcasted through comms.
A new endpoint is used to request the profiles. The response comes with a `hash` and `signedHash` which later will be used for the impostor checking.
The about response of the catalysts now includes a `publicKey` which is used to validate the `hash` & `signedHashed`.
It is required a request to the catalyst for every local profile change so the hashes are updated and correctly propagated. Also it is required to force a profile request to the catalyst at the start of the session, to ensure that the hashes are updated.
Protocol dependency: https://github.com/decentraland/protocol/pull/179

## How to test the changes?

1. Start DCL with two different users: https://play.decentraland.zone/?explorer-branch=chore/impostor-prevention
2. Check that you are connected to any of these realms: hera, hela, heimdallr, baldr, artemis
3. You should be able to see both users in the world. Take in consideration that you will not be able to see any other peers that are not using this branch
4. Change wearables, name or anything in your profile
5. Check that the changes are propagated to the peers
6. Do the same tests on any world
7. Perform a `/changerealm`, you should keep seeing other peers
8. Check the onboarding realm flow
9. Check the LocalPreview realm

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 275eb7b</samp>

This pull request adds functionality to request and validate user profile data from the Catalyst server using checksums and public keys. It introduces a new interface and class `UserProfileAPIClient` to handle the HTTP requests and responses, and modifies the `UserProfileController`, `UserProfile`, and `UserProfileModel` classes to handle the realm information updates and the data validation. It also updates the `UserProfile.asmdef` file to add references to other assemblies, and the `UserProfileWebInterfaceBridge` class to expose a public method for the web interface to request and validate user profiles.


browser-interface/packages/shared/dao/types.ts
```diff
@@ -22,6 +22,8 @@ type BaseCandidate = {
   version: { bff: string; comms: string; lambdas: string; content: string }
   elapsed: number
   status: ServerConnectionStatus
+  publicKey: string
+  address: string
```
hugoArregui, line 26: we are not including the address in the about anymore, so maybe we should not include it here?

browser-interface/packages/shared/profiles/impostorValidation.ts
```diff
@@ -0,0 +1,19 @@
+import {Avatar} from "@dcl/schemas";
+import {sha3} from "eth-connect";
+import { ecdsaVerify } from 'ethereum-cryptography/secp256k1-compat'
+
+export function isImpostor(avatar: Avatar, profileHash: string, profileSignedHash: string, signerPublicKey: string): boolean {
+  let checksum = getProfileChecksum(avatar);
+  return checksum != profileHash || !verifySignature(profileHash, profileSignedHash, signerPublicKey)
```
hugoArregui, line 7: I think you should use `!==` instead of `!=`

browser-interface/test/dao/pick-realm-algorithm/test-utils.ts
```diff
@@ -121,6 +121,8 @@ function buildCandidate(params: Partial<Candidate>): Candidate {
     version: { content: '1.0.0', lambdas: '1.0.0', bff: '1.0.0', comms: 'v3' },
     acceptingUsers: true,
     usersParcels: circularSlice(usersCount, someRandomParcels),
+    publicKey: '',
+    address: '',
```
hugoArregui, line 125: I think we need to remove address

hugoArregui (approved):

Kinerius (approved):

lorux0:
This PR stays on hold until a new approach is resolved:
1. Fetch all profiles from the catalyst
2. Send address and signature through comms instead of the full profile

PR #5741, opened 2023-09-26
Feat: map setallite view

popuz:
## What does this PR change?

Implements Satellite View for the navMap. 
- done in similar (but slightly) different way as our Parcels for map
- UI to switch between Satellite and Parcel View
- HyperLink UI at the right bottom to jump to genesis.city web-site

## How to test the changes?

- **Satellite View**
1. Launch the explorer
2. Observe satellite view on minimap
3. Open NavMap and observe new Satellite View. 
4. Check that other NavMap functionality remains as was
5. Click Satellite and Parcels view buttons at the top and verify they are working
6. Click on the Hyper-link for genesis.city and verify it leads to the respective web-site
7. Verify player coordinates are no longer presented

- **Feature flag**
1. Run https://play.decentraland.zone/?explorer-branch=feat/map-settalite-view
2. Verify there is no Satellite view, related buttons or hyper-link presented
3. Note: player coordinates at the bottom right could be also not presented, but that is fine
 
## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 09d431a</samp>

This pull request adds a new satellite layer to the map rendering system, which allows the user to switch between the satellite and the parcels views. It also refactors some of the existing map rendering classes and adds new classes and components to handle the satellite layer logic and UI. The main files affected are `MapConfiguration.prefab`, `MapRenderer.cs`, `SatelliteChunkAtlasController.cs`, `NavMapChunksLayersController.cs`, and `NavMapChunksLayersView.cs`.


unity-renderer/Assets/DCLServices/MapRendererV2/MapRenderer.cs
```diff
@@ -98,13 +99,35 @@ private void ReleaseCamera(IMapCameraControllerInternal mapCameraController)
             mapCameraPool.Release(mapCameraController);
         }
 
+        private bool isSatelliteViewMode = true;
+
+        public void SetSatelliteViewMode(bool isActive)
```
mikhail-dcl, line 104: You should not manipulate with layers directly: there is already logic for this in `EnableLayers` and `DisableLayers`: `ActivityOwners` means that there could be several consumptions of the map: e.g. `MiniMap` and `Map`, it should be respected.
In order to implement it properly you should:
Slightly modify the aforementioned methods to: 
 - allow disable/enable layers incrementally from the active customer dynamically.
 - expose them
 - Make `ActivityOwners` respect reconfiguration of layers from the same consumer (currently it supposes methods are called only once on rental and on disposal respectively)

unity-renderer/Assets/DCLServices/MapRendererV2/MapRenderer.cs
```diff
@@ -98,13 +99,35 @@ private void ReleaseCamera(IMapCameraControllerInternal mapCameraController)
             mapCameraPool.Release(mapCameraController);
         }
 
+        private bool isSatelliteViewMode = true;
```
mikhail-dcl, line 102: Looks like a hack, what is this?

unity-renderer/Assets/DCLServices/MapRendererV2/MapRenderer.cs
mikhail-dcl: @davidejensen you should have a similar implementation with layers' capability being activated/deactivated

unity-renderer/Assets/DCLServices/MapRendererV2/MapLayers/SatelliteAtlas/SatelliteChunkController.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        private async UniTask<UnityWebRequest> GetTextureFromWebAsync(string url, int maxAttempts = 3, CancellationToken ct = default)
+        {
+            UnityWebRequest webRequest;
+
+            try { webRequest = await webRequestController.Ref.GetTextureAsync(url, cancellationToken: linkedCts.Token); }
+            catch (Exception e)
+            {
+                if (webRequestAttempts < maxAttempts)
```
mikhail-dcl, line 84: It seems redundant as repetition logic is already in ` webRequestController.Ref.GetTextureAsync`. This way you repeat the request 9 times.

unity-renderer/Assets/DCLServices/MapRendererV2/MapLayers/Atlas/ChunkAtlasController.cs
```diff
@@ -9,26 +9,21 @@ namespace DCLServices.MapRendererV2.MapLayers.Atlas
 {
     internal class ChunkAtlasController : MapLayerControllerBase, IAtlasController
     {
-        public delegate UniTask<IChunkController> ChunkBuilder(Vector3 chunkLocalPosition,
-            Vector2Int coordsCenter,
-            Transform parent,
-            CancellationToken ct);
+        public delegate UniTask<IChunkController> ChunkBuilder(Vector3 chunkLocalPosition, Vector2Int coordsCenter, Transform parent, CancellationToken ct);
 
         public const int CHUNKS_CREATED_PER_BATCH = 10;
```
mikhail-dcl, line 14: I think you should lower this number for both this class and your new one as it will be executed in parallel and cause additional hiccups than expected initially. Consider reducing it by twice (as you have 2 implementations)

mikhail-dcl (changes requested):
Overall looks good but layers changing logic should be implemented accordingly to the existing flow

unity-renderer/Assets/DCLServices/MapRendererV2/MapLayers/SatelliteAtlas/SatelliteChunkController.cs
popuz: I actually tried it with repetition logic from `webRequestController.Ref.GetTextureAsync` but it didn't fix the SSL connection exception I was getting in Editor. I can make branching here, so this part will be Editor-only. What do you think?

unity-renderer/Assets/DCLServices/MapRendererV2/MapLayers/Atlas/ChunkAtlasController.cs
mikhail-dcl: I would not load it on demand as we have the background capacity to do it. And you count it as mandatory as it's more readable than Sattelite. How about to start loading it after Sattelite then (not in parallel)?

mikhail-dcl (approved):

anicalbano (approved):
✅ Reviewed and approved by QA 💯 

image

image

https://github.com/decentraland/unity-renderer/assets/139157886/a76aa48d-b65a-4514-b19d-5dd7413b6fba

Issue #5698, opened 2023-09-11
[QA] A bright white light appears on screen for a number of users 


anicalbano:
**Description of the issue:**
In some ocations, while building and adding materials, some users reported a white light appearing on screen that doesn't go away after refreshing. There are no specific repro steps to catch the issue, since it happened in isolated situations.

Some of the reports came from https://play.decentraland.org/?realm=boedo.dcl.eth&position=0%2C0

And for additional info, we have the discussions on Slack:
https://decentralandteam.slack.com/archives/C01QVR7TJK1/p1694433886766899
https://decentralandteam.slack.com/archives/C01QVR7TJK1/p1694103722078469

![image](https://github.com/decentraland/unity-renderer/assets/139157886/cf9f8a7f-bad1-4530-9c10-05d0d0986fc0)

https://github.com/decentraland/unity-renderer/assets/139157886/6b1591df-11a4-47ca-b60e-5ddce329dcb0

dalkia:
The bright light is a result of a bad material forming that got introduced with the latest Unity release. The bright light can be seen only in Mac WebGL, but in other platforms the material is not being displayed correctly. More info on the different results on this [thread](https://decentralandteam.slack.com/archives/C03G496L6NN/p1694792103129199?thread_ts=1694791107.998549&cid=C03G496L6NN). 

The issue seems to occur when a mesh has scale 0 on any of its axis. This can be tested on this test [branch ](https://github.com/decentraland/unity-renderer/pull/5712). You can test it by going to -60, 0 and walking north. There are some lamp posts that should be displaying the DCL logo, but they arent. 

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBazJyIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--a11d358ecc77d0e1153c2cea78ffd47d6f84abab/image.png)
Correct material displayed

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBazZyIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--a409810c36fe8daf871f452097da2113263c7916/image.png)
Issue in Mac WebGL

[...]

PR #5750, opened 2023-09-27
fix: fix zero scale on z axis to avoid bright light artifacts

aixaCode:
## What does this PR change?

Attempt to fix issue with bright light mentioned here https://github.com/decentraland/unity-renderer/issues/5698
The root of the issue seems to be the improperly set scale of GameObjects, due to:
1. Scene code on entity transform, i.e DCL flags on the tram line.
2. Submeshes created by GLTFast (or maybe an AB version)
 
These problems come from the creators' configurations, so we need our renderer to handle these more gracefully.
There still can be a corner cases which could not be fixed, like submesh being broken

Added some extra logic to `utils.CapGlobalValuesToMax` to prevent issues with out-of-range values, without setting the z-axis to 0. This way, creators can still play around with x and y scales to hide objects or just use the visible boolean in their gltfshape.
Also brought back the changes to enable bloom in the settings.

Ideally, we’d stop these from rendering and throw back some feedback during testing. For now, consider this as a temporary fix—we’ll be looking into more robust solutions

## How to test the changes?

1. Launch the explorer.
2. Ensure the bloom is on in the settings.
5. Test location defined in the issue.
6. No bright light should be observed.

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at a62fb7b</samp>

This pull request fixes a bug with the bloom effect scale, adds a feature to disable the bloom toggle for performance reasons, and improves the consistency of the bloom toggle state. It affects the `Utils`, `BloomControlConfiguration`, `BloomDisabledAndOffByDefault`, and `BloomControlController` classes and assets.


unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/Utils/Utils.cs
```diff
@@ -134,13 +134,18 @@ public static Vector2Int StringToVector2Int(string coords)
...
 
         public static void CapGlobalValuesToMax(this Transform transform)
         {
             bool positionOutsideBoundaries = transform.position.sqrMagnitude > MAX_TRANSFORM_VALUE * MAX_TRANSFORM_VALUE;
             bool scaleOutsideBoundaries = transform.lossyScale.sqrMagnitude > MAX_TRANSFORM_VALUE * MAX_TRANSFORM_VALUE;
 
-            if (positionOutsideBoundaries || scaleOutsideBoundaries)
+            //Extra check to prevent rendering artifacts when bloom is on
+            //More description: https://github.com/decentraland/unity-renderer/issues/5698
+            bool scaleZAxisBelowMinimum = Mathf.Abs(transform.localScale.z) < MIN_TRANSFORM_SCALE_Z_VALUE;
```
pravusjif, line 146: We saw the problem with Z axis but it sounds to me like it could perfectly happen with X or Y in 0 as well, maybe we should just have a generic `MIN_TRANSFORM_SCALE_VALUE` and check the 3 axis instead of just Z, what do you think?

aixaCode: Yeah, I was thinking about this at the beginning, but when testing noticed that if you set scale on x or y the object is hidden and not being rendered at all. Because of that, I went with just z scale to avoid more breaking changes if creators are using scale in some way

pravusjif (approved):

anicalbano:
✅ Re-checked for release:

image
image

PR #5768, opened 2023-10-04
feat: add support for worlds using ens domains

marianogoldman:
## What does this PR change?
Worlds Content Server now also supports deployment of ENS names apart from currently supported DCL names. So you can now have a world deployed as `your-name.eth` if you own it.
This PR adds support for:
1. Entering the world directly via URL `https://play.decentraland.org/?realm=yourname.eth`
2. Entering the world via /changerealm chatbox command `/changerealm yourname.eth`.
3. Entering the world via /world chatbox command `/world yourname.eth`.

...

## How to test the changes?

Use any of the 3 methods listed above as 3 different ways for entering a World deployed under an ENS domain.

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 844a5cc</samp>

Improved the support for ENS names in portable experiences and realm resolution. Removed an unused import from `ArchipelagoConnection.ts`.

### DIFF
--- a/browser-interface/packages/shared/realm/resolver.ts
+++ b/browser-interface/packages/shared/realm/resolver.ts
@@ -72,11 +72,7 @@ export function prettyRealmName(realm: string, candidates: Candidate[]) {
 }
 
 export function isEns(str: string | undefined): str is `${string}.eth` {
-  return !!str?.match(/^[a-zA-Z0-9]+\.eth$/)?.length
-}
-
-export function isDclEns(str: string | undefined): str is `${string}.dcl.eth` {
-  return !!str?.match(/^[a-zA-Z0-9]+\.dcl\.eth$/)?.length
+  return !!str?.match(/^[a-zA-Z0-9.]+\.eth$/)?.length
 }
 
 export function dclWorldUrl(dclName: string) {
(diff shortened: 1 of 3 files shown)


browser-interface/packages/shared/realm/resolver.ts
```diff
@@ -71,6 +71,10 @@ export function prettyRealmName(realm: string, candidates: Candidate[]) {
   return realm
 }
 
+export function isEns(str: string | undefined): str is `${string}.eth` {
+  return !!str?.match(/^[a-zA-Z0-9]+\.eth$/)?.length
```
fzavalia, line 75: This will not support ens subdomains like `nando.company.eth`

hugoArregui (approved):

browser-interface/packages/shared/realm/resolver.ts
fzavalia: Doing that will probably make it unnecessary to use isDclEns on other places as it will pick .dcl.eth names as well

marianogoldman: Thanks @fzavalia. Much nicer and simpler.

PR #5791, opened 2023-10-11
fix: sdk7 avatar modifier area

pravusjif:
Solves https://github.com/decentraland/sdk/issues/963

* Refactored avatar modifier area implementation to use a dedicated system and simplify its componnent handler (the original implementation was too legacy)
* Fixed 'ExcludedIds' not working correctly
* Optimized area checks garbage memory allocation to allocate 100b~ instead of 320b~
* Replaced old tests with more reliable ones

-------------------------------

### <samp>🤖 Generated by Copilot at c8a5111</samp>

This pull request refactors and improves the avatar modifier area feature, which allows modifying the appearance and behavior of avatars in a specific area. It introduces a new internal component for storing and updating the state and logic of the feature, and a new system for applying or removing the modifiers. It also updates the existing component, register, and test classes to use the internal component and the test utilities. It removes some unused and obsolete code and files, and adds some new assembly definitions and references. It modifies the `ECS7ComponentsComposer` class to pass the internal components to the registers that need them. It also reorganizes and optimizes some code structures and methods.

-------------------------------

### QA TEST INSTRUCTIONS - AVATAR MODIFIER AREA "EXCLUDE IDs" FEATURE FIX

1. In a new incognito browser window open [this playground link](https://playground.decentraland.org/?explorer-branch=fix%2Fsdk7-avatar-modifier-area&code=Y29uc3QgZXhjbHVkZWRVc2VyID0gJ1VTRVItV0FMTEVULUFERFJFU1MtR09FUy1IRVJFJwpjb25zdCBhdmF0YXJNb2RpZmllckFyZWFFbnRpdHkgPSBlbmdpbmUuYWRkRW50aXR5KCkKICBUcmFuc2Zvcm0uY3JlYXRlKGF2YXRhck1vZGlmaWVyQXJlYUVudGl0eSwgeyBwb3NpdGlvbjogVmVjdG9yMy5jcmVhdGUoMiwgMSwgMil9KQogIEF2YXRhck1vZGlmaWVyQXJlYS5jcmVhdGUoYXZhdGFyTW9kaWZpZXJBcmVhRW50aXR5LCB7CiAgICBhcmVhOiBWZWN0b3IzLmNyZWF0ZSg0LCA0LCA0KSwKICAgIG1vZGlmaWVyczogW0F2YXRhck1vZGlmaWVyVHlwZS5BTVRfSElERV9BVkFUQVJTXSwKICAgIGV4Y2x1ZGVJZHM6IFtleGNsdWRlZFVzZXJdCiAgfSkKCgo%3D) and wait for the world to load.
2. Change the `const excludedUser = 'USER-WALLET-ADDRESS-GOES-HERE'` line to use a real user id (metamask wallet address) you own. For example: `const excludedUser = '0x0F5D2fB29fb7d3CFeE444a200298f468908cC942'`
3. Hit the "Run" Button
4. Notice that in the starting 4x4 area, your incognito avatar should be hidden by the avatar modifier area that exists in that zone only. If you move outside of that zone your avatar should appear again (make sure you are using third person camera, that can be toggled with the 'V' key)
5. open a non-incognito window, make sure you are logged in with metamask (with the same address that you put in the code in the incognito playground window)
6. in the new non-incognito window enter https://playground.decentraland.org/?ENABLE_WEB3 and sign-in with metamask
7. Wait for the world to load and place your logged-in avatar inside the 4x4 area. You won't see it disappear since the code running in that new window is different, that's OK.
8. Go back to the incognito window and confirm that you see the avatar you logged in in the other window, and that it's unaffected by the avatar modifier area (it should be seen at all times and not hidden)


unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/AvatarModifierArea/Handler/AvatarModifierAreaComponentHandler.cs
```diff
@@ -11,200 +9,54 @@ namespace DCL.ECSComponents
...
-                    dataStore.otherPlayers.OnAdded += OtherPlayersOnOnAdded;
-                }
+                var modifier = factory.GetOrCreateAvatarModifier(modifierKey);
 
-                this.model = model;
-                
-                // Force update due to after model update modifiers are removed and re-added
-                // leaving a frame with the avatar without the proper modifications
-                Update();
+                internalComponentModel.OnAvatarEnter += modifier.ApplyModifier;
```
lorux0, line 46: Could we get duplicated subscriptions here?

pravusjif: uhmm yes you are right, just in case I added those events cleanup before that iteration, thanks!

anicalbano (approved):
🟢 Reviewed by QA with positive results. Each one of the following screens depicts the steps of testing:

image

image

image

image

lorux0 (approved):

PR #5796, opened 2023-10-12
fix: emotes and avatar visibility

Kinerius:
## What does this PR change?

Fixed:
- Ghost avatars in T pose
- Emotes being executed and renderers when avatars are hidden within Avatar Modifier areas
- A rare exception when loading wearables
- Forced emotes to always render, the culling is broken so this is a workaround.

## How to test the changes?

- Log in to a zone with multiple avatars, they should not be T posing
- Go to an area with an Avatar Modifier area that hides the avatar ( I think the machines at Wondermine have one ) and execute an emote, the emote should not show or play a sound.

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 847f412</samp>

Refactored the emotes loading logic in the `Avatar` system and fixed some bugs related to emote sound and visibility. Also improved the handling of duplicate categories in the `WearableItem` class.


anicalbano (approved):
✅ Reviewed and approved by QA

davidejensen (approved):
nice!

PR #5797, opened 2023-10-12
fix: fetch realms info before showing the realm selector

lorux0:
## What does this PR change?

Request the list of available realms, so the user count is updated before showing the realm selector popup.
Fixes cases when the user count is displaying incorrect.

## How to test the changes?

1. Launch the explorer
2. Open the start menu
3. Open the realm selector
4. You should be able to see the realms correctly
5. Interact with the realm selector popup, everything should keep working as expected

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at 820d8b0</samp>

Added a new feature to the explore menu to allow users to select a different realm. Implemented a bridge between the web interface and the unity interface to fetch and display realms information. Updated the `RealmsInfoHandler` class to use `UniTask` and `WebInterface` for asynchronous requests. Added the necessary assembly references and dependencies to use the `RealmsInfoBridge` and `IRealmsInfoBridge` classes.

### DIFF
--- a/browser-interface/packages/shared/renderer/sagas.ts
+++ b/browser-interface/packages/shared/renderer/sagas.ts
@@ -125,7 +125,7 @@ function* reportRealmChangeToRenderer() {
   }
 }
 
-async function fetchAndReportRealmsInfo(url: string) {
+export async function fetchAndReportRealmsInfo(url: string) {
   try {
     const response = await fetch(url)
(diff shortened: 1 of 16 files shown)


anicalbano (approved):
🟢 Reviewed and approved by QA

Kinerius (approved):

PR #5800, opened 2023-10-13
chore: main realm is always available, but won't be realm picked unless FF is enabled

hugoArregui:
This means:

- If FF is disabled (is currently disabled): you won't be ever assigned to main realm by realm picking algorithm
- `/changerealm main` will work irregardless of the FF status
- jumping on the main realm from the UI will work irregardless of the FF status

### DIFF
--- a/browser-interface/packages/shared/dao/sagas.ts
+++ b/browser-interface/packages/shared/dao/sagas.ts
@@ -10,7 +10,8 @@ import {
   getAddedServers,
   getCatalystNodesEndpoint,
   getDisabledCatalystConfig,
-  getPickRealmsAlgorithmConfig
+  getPickRealmsAlgorithmConfig,
+  isMainRealmEnabled
 } from 'shared/meta/selectors'
 import { SET_REALM_ADAPTER, setOnboardingState } from 'shared/realm/actions'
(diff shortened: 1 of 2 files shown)


browser-interface/packages/shared/dao/sagas.ts
```diff
@@ -68,15 +69,20 @@ function* pickCatalystRealm() {
...
 
+  const mainRealmEnabled = yield select(isMainRealmEnabled)
+
   const filteredCandidates = candidates.filter((candidate: Candidate) => {
     const lastConnected = lastConnectedCandidates.get(candidate.domain)
     if (lastConnected && Date.now() - lastConnected < 60 * 1000) {
       return false
     }
+    if (!mainRealmEnabled) {
+      return candidate.catalystName != 'main'
```
marianogoldman, line 80:
```suggestion
      return candidate.catalystName !== 'main'
```

anicalbano (approved):
🟢 Reviewed and QA and working as expected!

https://github.com/decentraland/unity-renderer/assets/139157886/99a2f02b-5888-488b-a919-ba2dd87e8045

popuz (approved):

PR #5807, opened 2023-10-13
test: sdk ignore media hostnames for non portable experiences

pravusjif:
Implemented media-hostname permissions ignoring for non portable experiences.

### <samp>🤖 Generated by Copilot at 77c5fbd</samp>

This pull request adds support for external media sources for portable experiences, which are special scenes that can be deployed and executed from any other scene. It modifies the `TryGetMediaUrl` method and its callers in various classes to add a new parameter `isPortableExperience` that allows skipping the check for the scene permissions and the allowed media hostnames. It also updates the unit tests for the `VideoPlayerHandler` class to reflect the new feature.


pravusjif:
Closing this experiment for now until further notice.

PR #5843, opened 2023-10-24
fix: add interaction missing textures

RominaMarchetti:
This PR fixes the missing textures for key and mouse inputs on SDK7 scenes.

Screenshot 2023-10-24 at 15 16 28


pravusjif (approved):
This fixes the missing/broken interacction icon

PR #5871, opened 2023-10-30
fix: POI info button

RominaMarchetti:
This branch is created to fix the missing info button next to the POI filter in the nav map. 
Screenshot 2023-10-30 at 10 04 14

### How to test?
1- Open this [link](https://play.decentraland.org/?explorer-branch=fix/ToggleOnPoiInfo)
2- Open the Genesis City Map
3- Open the filters and check that the info button next to the POI filter is visible and the tooltip is shown when clicking it


anicalbano (approved):
🟢 Reviewed and approved by QA ✔️ 

image

PR #5883, opened 2023-11-01
fix: preview mode crash on start

pravusjif:
* Added missing nullcheck since aparently in preview-mode/localhost the `columnsOrganizerComponentView` doesn't exist or is not initialized.
* This fixes the current broken state of the latest explorer release with any scene in preview-mode (it crashes and the world doesn't start in preview mode in localhost)

### <samp>🤖 Generated by Copilot at 95bb556</samp>

Fixed a potential null reference exception in the `EmotesCustomizationComponentView` class by making a field nullable. This improves the stability of the emotes customization UI.


anicalbano (approved):
✔️  No QA

Kinerius (approved):

PR #5884, opened 2023-11-01
fix: sdk7 ui text position

pravusjif:
* Reverted initial position type for ui text
* Added position type update based on root element

-------------------------

### <samp>🤖 Generated by Copilot at 2cf7646</samp>

Refactor the UI system to fix layout issues and simplify UI element creation. Change the default position of UI elements to absolute and add a condition for auto width or height in `UIElementHandlerBase.cs`. Move the `SetElementDefaultStyle` method from `UiElementUtils.cs` to `UIElementHandlerBase.cs`.

-------------------------

### QA TEST INSTRUCTIONS

1. Enter the the residence tower at https://play.decentraland.org/?realm=main&position=116%2C-26&explorer-branch=fix/sdk7-ui-text-position
2. Enter the lobby and click on the "Elevator" button you'll see in the scene UI
3. Confirm the "Bar" button text is centered in the new panel that opens
4. If you test the same with https://play.decentraland.org/?realm=main&position=116%2C-26 (production build) you'll see it's not centered, so it's wrong in production.

Screen Shot 2023-11-01 at 15 32 50

### DIFF
--- a/unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Utils/UiElementUtils.cs
+++ b/unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Utils/UiElementUtils.cs
@@ -10,7 +10,7 @@ public static class UiElementUtils
         elementStyle.bottom = 0;
         elementStyle.width = new StyleLength(StyleKeyword.Auto);
         elementStyle.height = new StyleLength(StyleKeyword.Auto);
-        elementStyle.position = new StyleEnum<Position>(Position.Relative);
+        elementStyle.position = new StyleEnum<Position>(Position.Absolute);
         elementStyle.justifyContent = new StyleEnum<Justify>(Justify.Center);
         elementStyle.alignItems = new StyleEnum<Align>(Align.Center);
     }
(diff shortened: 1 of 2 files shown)


unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/Utils/UiElementUtils.cs
```diff
@@ -10,7 +10,7 @@ public static void SetElementDefaultStyle(IStyle elementStyle)
         elementStyle.bottom = 0;
         elementStyle.width = new StyleLength(StyleKeyword.Auto);
         elementStyle.height = new StyleLength(StyleKeyword.Auto);
-        elementStyle.position = new StyleEnum<Position>(Position.Relative);
+        elementStyle.position = new StyleEnum<Position>(Position.Absolute);
```
gonpombo8, line 13: mm its weird.
We are sending the position: relative by default on the sdk level 🤔 
https://github.com/decentraland/js-sdk-toolchain/blob/main/packages/%40dcl/react-ecs/src/components/uiTransform/index.ts#L37

pravusjif: that change right now only affects the UiText, and it only affects the inner visual element of the UiText.
Some UiComponents we create are actually composed of 2 visual elements: a root element that acts as a kind of container and an inner one. UiText has a container and a label as the child.

We were already setting this as Absolute before I implemented the auto-size support.

LucasLioyQA (commented):
Screenshot 2023-11-01 at 11 52 45

**QA approved**

PR #5891, opened 2023-11-01
fix hot reload for smart wearables

gonpombo8:
Fix hot reload for smart wearables.

It doesnt change any code that can be reached at play.decentraland.org.

It changes only a piece of code that is being called only with the Preivew mode

### how to test
- create a smart wearable
- install this version of the sdk https://github.com/decentraland/js-sdk-toolchain/pull/822#issuecomment-1789569493
- run npm start and when the explorer opens, add the following param to the url `&explorer-branch=fix/sw-hot-reload`


pravusjif (approved):
LGTM!

PR #5925, opened 2023-11-13
hotfix: if the profile doesn't come from the catalyst, use the local one

sandrade-dcl:
## What does this PR change?
Hotfix

## Our Code Review Standards

### <samp>🤖 Generated by Copilot at e16f2d7</samp>

Fixed a bug that reset the profile when switching realms by checking for a local profile when the catalyst server returns null. Modified `initialRemoteProfileLoad.ts` to handle null or undefined profiles.


Kinerius (approved):
**L**ikey **G**reat **T**ry **M**erging

PR #5950, opened 2023-11-17
hotfix: 2023-11-17

Kinerius:
## What does this PR change?

Fixed emote SFX volume not being updated when changing the volume slider.

Removed 60 seconds timeout when loading user wearables since under heavy loads this can take some time for users with hundreds of emotes and we are not even handling the timeout correctly with UX

## How to test the changes?

- Check that the volume slider in the Settings panel is called "AVATAR & EMOTE SFX".
- Check that an emote being played is updated with the new volume settings when moving the slider.
- Check with a user with hundreds of emotes that their catalog loads correctly after some time. 

## Our Code Review Standards

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 495e2c5</samp>

Added a new class `EmoteVolumeHandler` to control the volume of emote audio sources based on the audio settings and the distance from the camera. Refactored and simplified the code that used to handle the emote volume in different classes and methods. Renamed the `AVATAR SFX` setting to `AVATAR & EMOTE SFX` to reflect the change.

### DIFF
--- a/unity-renderer/Assets/DCLServices/EmotesCatalog/EmotesCatalogService/LambdasEmotesCatalogService.cs
+++ b/unity-renderer/Assets/DCLServices/EmotesCatalog/EmotesCatalogService/LambdasEmotesCatalogService.cs
@@ -1,5 +1,3 @@
-using System.Collections.Generic;
-using System.Threading;
 using Cysharp.Threading.Tasks;
 using DCL;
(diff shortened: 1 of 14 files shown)


anicalbano (approved):
🟢 Reviewed by QA and Support and working as expected.

PR #5962, opened 2023-11-17
style: update live event container on teleport prompt

RominaMarchetti:
## What does this PR change?

This PR was created to fix the copy and architecture of the event module in the Teleport prompt.
It should look like this now:
Screenshot 2023-11-17 at 13 02 13

## How to test the changes?

1- Open this [link](https://play.decentraland.org/?explorer-branch=style/LiveEventContainerTeleportPrompt)
2- Go to the events column in the genesis plaza and click on the jump in button of a live event card.
3- Check that instead of `current event` now the label says `live` and the amount of players is next to it. Note that instead of a solid container in the background now there is a gradient. All this was done in order to align its aesthetic with the discover cards (was outdated). Plus, the whole modal is bigger now.


LucasLioyQA (approved):
**QA approved** on mac M1 chipset using chrome :

Screenshot 2023-11-17 at 15 01 20

Screenshot 2023-11-17 at 15 01 39

PR #6015, opened 2023-12-10
fix: sdk7 avatar shape reposition on scene reload

pravusjif:
Added missing variable reset on cleanup: without that any sdk7 avatar shape entity gets wrongly repositioned when its scene is reloaded (by moving away or by hot reload).

Fixes https://github.com/decentraland/sdk/issues/1051.

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 396a36e</samp>

Fix avatar position bug when switching realms by resetting `initializedPosition` flag in `AvatarShape` component.


popuz (approved):
make sense :)

PR #6055, opened 2024-01-11
fix: Use sso local functions

fzavalia:
## What does this PR change?

Removes the initialization of the SSO iframe, only uses SSO local functions for compatibility. This PR is important as the iframe has been deprecated and there is no need for it to still be used.

Also call the deleteSession function before emitting the logout event.

...

## How to test the changes?

1. Launch the explorer
2. Test that connection and disconnection still work as always

## Our Code Review Standards

## Copilot summary

copilot:summary


LautaroPetaccio (approved):

aixaCode:
@fzavalia Any specific time that needs to be included in the release?

fzavalia:
@aixaCode Any time, it is not a blocker but would be great to stop initializing the iframe given that it is not used anymore.

PR #6116, opened 2024-03-04
fix: add extended urns for saving outfits

lorux0:
## What does this PR change?

Allows the outfit's wearables to be saved as extended urn, required by servers.

## How to test the changes?

1. Launch the explorer with param: `&CATALYST=https://peer.decentraland.zone`
2. Open the backpack
3. Go to outfits
4. Check that are loaded as expected
5. Play with outfits, like editing/removing/adding
6. Close the backpack and restart the client
7. Check that the outfits are persisted as expected

## Our Code Review Standards

## Copilot summary

copilot:summary

### DIFF
--- a/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/BackpackEditorHUDModel.cs
+++ b/unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/BackpackEditorHUDModel.cs
@@ -13,12 +13,25 @@ namespace DCL.Backpack
         public Color eyesColor;
         public HashSet<string> forceRender = new ();
 
-        public AvatarModel ToAvatarModel()
+        public AvatarModel ToAvatarModel(Dictionary<string, string> extendedWearableUrns = null)
         {
+            List<string> wearables;
+
+            if (extendedWearableUrns != null)
+            {
+                wearables = new List<string>();
+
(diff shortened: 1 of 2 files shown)


unity-renderer/Assets/Scripts/MainScripts/DCL/Controllers/HUD/BackpackEditorHUDV2/BackpackEditorHUDModel.cs
```diff
@@ -13,12 +13,25 @@ public record BackpackEditorHUDModel
...
         public HashSet<string> forceRender = new ();
 
-        public AvatarModel ToAvatarModel()
+        public AvatarModel ToAvatarModel(Dictionary<string, string> extendedWearableUrns = null)
         {
+            List<string> wearables;
+
+            if (extendedWearableUrns != null)
+            {
+                wearables = new List<string>();
```
popuz, line 22: Is this method called only once? We don't need to cache this List (or use Unity pool for it), right?

popuz (approved):
LGTM 🟢

marianogoldman:
I did some testing on my own and all looking good, the scenario we identified yesterday seems fixed in this branch. Thanks @lorux0.
@LucasLioyQA please run all the cases you have and then let's see if we can release this as a hot fix.

LucasLioyQA (approved):
All entity requests on the network section have provided a 200 successful response so far 

Screenshot 2024-03-05 at 10 28 51

PR #6120, opened 2024-03-06
update avatar scene to prevent rendering players when blocked

nicoecheza:
## What does this PR change?

[Shape up - Cycle 5 => Block players visually](https://www.notion.so/decentraland/Block-Players-Visually-7ecdf67bd0b64d76bb29e01b56f60f4f)

## How to test the changes?
Block a player and you should not be able to see them anymore

### DIFF
--- a/browser-interface/packages/ui/avatar/avatarSystem.ts
+++ b/browser-interface/packages/ui/avatar/avatarSystem.ts
@@ -22,7 +22,7 @@ type Position = {
   rotationW: number
 }
 
-const avatarMap = new Map<string, AvatarEntity>()
+export const avatarMap = new Map<string, AvatarEntity>()
 
 export class AvatarEntity extends Entity {
   visible = true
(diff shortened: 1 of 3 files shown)


browser-interface/packages/ui/decentraland-ui.scene.ts
```diff
@@ -1,5 +1,5 @@
 import { executeTask } from '@dcl/legacy-ecs'
-import { avatarMessageObservable } from './avatar/avatarSystem'
+import { avatarMap, avatarMessageObservable } from './avatar/avatarSystem'
```
gonpombo8, line 2: use getPeer

browser-interface/packages/ui/avatar/avatarSystem.ts
```diff
@@ -22,7 +22,7 @@ type Position = {
   rotationW: number
 }
 
-const avatarMap = new Map<string, AvatarEntity>()
+export const avatarMap = new Map<string, AvatarEntity>()
```
gonpombo8, line 25:
```suggestion
const avatarMap = new Map<string, AvatarEntity>()
```

gonpombo8: use getPeer function instead

PR #6135, opened 2024-03-15
Introduce AudioSystem to write back the current state of an audio source

m3taphysics:
## What does this PR change?

Introduces AudioSourceSystem that can write back the state of an audio source.

## Our Code Review Standards

## Copilot summary

copilot:summary

### DIFF
--- a/unity-renderer/Assets/DCLPlugins/ECS7/Systems/AudioSourceSystem/ECSAudioSourceSystem.cs
+++ b/unity-renderer/Assets/DCLPlugins/ECS7/Systems/AudioSourceSystem/ECSAudioSourceSystem.cs
@@ -1,0 +1,53 @@
+﻿using DCL.ECS7;
+using DCL.ECS7.ComponentWrapper.Generic;
+using DCL.ECS7.InternalComponents;
+using DCL.ECSComponents;
+using DCL.ECSRuntime;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace ECSSystems.AudioSourceSystem
+{
+    public class ECSAudioSourceSystem
+    {
+        private readonly IInternalECSComponent<InternalAudioSource> audioSourceComponent;
+        private readonly IReadOnlyDictionary<int, ComponentWriter> componentsWriter;
+        private readonly WrappedComponentPool<IWrappedComponent<PBAudioSource>> audioSourcePool;
+
+        public ECSAudioSourceSystem(
+            IInternalECSComponent<InternalAudioSource> audioSourceComponent,
+            IReadOnlyDictionary<int, ComponentWriter> componentsWriter,
+            WrappedComponentPool<IWrappedComponent<PBAudioSource>> audioSourcePool)
+        {
+            this.audioSourceComponent = audioSourceComponent;
+            this.componentsWriter = componentsWriter;
+            this.audioSourcePool = audioSourcePool;
+        }
+
+        public void Update()
+        {
+            var audioSourceComponentList = audioSourceComponent.GetForAll();
+
+            // Loop through every audio source component
+            foreach (var component in audioSourceComponentList)
+            {
+                var scene = component.key1;
+                var entity = component.key2;
+                var model = component.value.model;
+
+                var wrappedComponent = audioSourcePool.Get();
+                wrappedComponent.WrappedComponent.Model.Playing = model.audioSource.isPlaying;
+                wrappedComponent.WrappedComponent.Model.Loop = model.loop;
+                wrappedComponent.WrappedComponent.Model.Volume = model.volume;
+                wrappedComponent.WrappedComponent.Model.Pitch = model.pitch;
+                wrappedComponent.WrappedComponent.Model.AudioClipUrl ??= model.audioClipUrl;
+                wrappedComponent.WrappedComponent.Model.CurrentTime = model.audioSource.time;
+
+                if (componentsWriter.TryGetValue(scene.sceneData.sceneNumber, out var writer))
+                {
+                    writer.Put(entity, ComponentID.AUDIO_SOURCE, wrappedComponent);
+                }
+            }
+        }
+    }
+}
--- a/unity-renderer/Assets/DCLPlugins/ECS7/ECS7Plugin.asmdef
+++ b/unity-renderer/Assets/DCLPlugins/ECS7/ECS7Plugin.asmdef
@@ -39,7 +39,8 @@
         "GUID:7b4a5d0cf82e544daa512b41aec35b00",
         "GUID:acb0f5633d7c47aba96ee6b61c872d94",
         "GUID:1e10040ab6604cbe8ad0921047d49e65",
-        "GUID:ac62e852826a4b36aeb22931dad73edb"
+        "GUID:ac62e852826a4b36aeb22931dad73edb",
+        "GUID:f51ebe6a0ceec4240a699833d6309b23"
     ],
     "includePlatforms": [],
     "excludePlatforms": [],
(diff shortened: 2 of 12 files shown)


unity-renderer/Assets/DCLPlugins/ECS7/ECS7Plugin.cs
```diff
@@ -6,6 +6,7 @@
 using DCL.ECSRuntime;
 using RPC.Context;
 using System.Collections.Generic;
+using UnityEngine;
```
AlejandroAlvarezMelucciDCL, line 9: Maybe this is a leftover?

unity-renderer/Assets/DCLPlugins/ECS7/ECS7Plugin.cs
```diff
@@ -68,8 +69,8 @@ public ECS7Plugin()
                 new WrappedComponentPool<IWrappedComponent<PBEngineInfo>>(MAX_EXPECTED_SCENES, () => new ProtobufWrappedComponent<PBEngineInfo>(new PBEngineInfo())),
                 new WrappedComponentPool<IWrappedComponent<PBUiCanvasInformation>>(MAX_EXPECTED_SCENES, () => new ProtobufWrappedComponent<PBUiCanvasInformation>(new PBUiCanvasInformation())),
                 new WrappedComponentPool<IWrappedComponent<PBPointerEventsResult>>(MAX_EXPECTED_SCENES * 10, () => new ProtobufWrappedComponent<PBPointerEventsResult>(new PBPointerEventsResult())),
-                new WrappedComponentPool<IWrappedComponent<PBTweenState>>(MAX_EXPECTED_SCENES * 10, () => new ProtobufWrappedComponent<PBTweenState>(new PBTweenState()))
-            );
+                new WrappedComponentPool<IWrappedComponent<PBTweenState>>(MAX_EXPECTED_SCENES * 10, () => new ProtobufWrappedComponent<PBTweenState>(new PBTweenState())),
+                new WrappedComponentPool<IWrappedComponent<PBAudioSource>>(MAX_EXPECTED_SCENES * 10, () => new ProtobufWrappedComponent<PBAudioSource>(new PBAudioSource())));
```
AlejandroAlvarezMelucciDCL, line 73: Maybe it's out of the scope and not the best time for it, but can we take the opportunity to change the magic number 10 by something with a meaningful name? I'm new so maybe I'm not aware of the team's conventions yet, but something like MAX_SCENES_MODIFIER might make sense?

Also, just for personal learning, why some components use the MAX_EXPECTED_SCENES without the multiplication by 10? Maybe it's a oversight and they also need it and because of that, maybe we would just need to increase MAX_EXPECTED_SCENES?

unity-renderer/Assets/DCLPlugins/ECS7/ECSComponents/AudioSource/Handler/ECSAudioSourceComponentHandler.cs
```diff
@@ -1,10 +1,13 @@
...
+﻿
+using Cysharp.Threading.Tasks;
+using System;
 using DCL.Controllers;
 using DCL.ECS7.InternalComponents;
 using DCL.ECSRuntime;
 using DCL.Helpers;
 using DCL.Models;
 using DCL.SettingsCommon;
+using System.Collections.Generic;
```
AlejandroAlvarezMelucciDCL, line 10: I think `using Cysharp.Threading.Tasks;` and `using System.Collections.Generic;` are not needed, maybe a left over?

unity-renderer/Assets/DCLPlugins/ECS7/Systems/AudioSourceSystem/ECSAudioSourceSystem.cs
```diff
@@ -0,0 +1,53 @@
+﻿using DCL.ECS7;
+using DCL.ECS7.ComponentWrapper.Generic;
+using DCL.ECS7.InternalComponents;
+using DCL.ECSComponents;
+using DCL.ECSRuntime;
+using System.Collections.Generic;
+using UnityEngine;
```
AlejandroAlvarezMelucciDCL, line 7: I think we can remove UnityEngine and DCL.ECSRuntime 👌

unity-renderer/Assets/DCLPlugins/ECS7/Systems/AudioSourceSystem/ECSAudioSourceSystem.cs
```diff
@@ -0,0 +1,53 @@
...
+            this.componentsWriter = componentsWriter;
+            this.audioSourcePool = audioSourcePool;
+        }
+
+        public void Update()
+        {
+            var audioSourceComponentList = audioSourceComponent.GetForAll();
+
+            // Loop through every audio source component
+            foreach (var component in audioSourceComponentList)
```
AlejandroAlvarezMelucciDCL, line 32: I think it might be worth it to switch from a `foreach` to a `for` loop to reduce allocations. What do you think?

unity-renderer/Assets/DCLPlugins/ECS7/Systems/AudioSourceSystem/ECSAudioSourceSystem.cs
```diff
@@ -0,0 +1,53 @@
...
+
+                var wrappedComponent = audioSourcePool.Get();
+                wrappedComponent.WrappedComponent.Model.Playing = model.audioSource.isPlaying;
+                wrappedComponent.WrappedComponent.Model.Loop = model.loop;
+                wrappedComponent.WrappedComponent.Model.Volume = model.volume;
+                wrappedComponent.WrappedComponent.Model.Pitch = model.pitch;
+                wrappedComponent.WrappedComponent.Model.AudioClipUrl ??= model.audioClipUrl;
+                wrappedComponent.WrappedComponent.Model.CurrentTime = model.audioSource.time;
+
+                if (componentsWriter.TryGetValue(scene.sceneData.sceneNumber, out var writer))
```
AlejandroAlvarezMelucciDCL, line 46: Maybe we can re-order this check to the beginning to avoid doing all the above in case the `componentsWriter` doesn't have a writer for the given `sceneNumber`?

unity-renderer/Assets/DCLPlugins/ECS7/ECS7Plugin.cs
m3taphysics: @AlejandroAlvarezMelucciDCL I looked at this but each component has arbitrary rules as to how big to pre-allocate the pools, for example transform only has *2, so this appears to be have been determined through some sort of experiementation.

@pravusjif may be able to offer insight into what this means!

unity-renderer/Assets/DCLPlugins/ECS7/Systems/AudioSourceSystem/ECSAudioSourceSystem.cs
m3taphysics: @AlejandroAlvarezMelucciDCL you are right! The IEnumerator call will allocate, good spot

unity-renderer/Assets/DCLPlugins/ECS7/ECS7Plugin.cs
AlejandroAlvarezMelucciDCL: Thanks for the explanation!

pravusjif: Uff those hardcoded values are from before I joined the SDK team so I'm not really sure but I would also trust them to have been defined based on some manual testing, most likely. I'd leave them as they are on unity-renderer at lesat, we won't be using those values in the new explorer.

unity-renderer/Assets/DCLPlugins/ECS7/Systems/AudioSourceSystem/ECSAudioSourceSystem.cs
pravusjif: Uhmmm it would be weird to have a scene without a scene writer, to be honest

PR #6153, opened 2024-04-02
feat: sdk7 audio stream audio event component

pravusjif:


browser-interface/packages/unity-interface/audioStream.ts
```diff
@@ -1,7 +1,82 @@
...
+  'waiting',
+  'playing',
+  'pause'
+]
 
 const audioStreamSource = new Audio()
-let playToken: number = 0;
+AUDIO_EVENTS.forEach(($) => audioStreamSource.addEventListener($, listen))
+
+function listen(ev: AudioEvents[keyof AudioEvents]) {
```
gonpombo8, line 38: dont feel safe by passing the engine here.
Why dont do the same approach that we have for the comms message ?
Create a mitter and every time there is a change on a listener emit an event with the sceneId number that triggered that event.

Then on the internalEngine we can listen to those changes.
This way we have all the logic of the components INSIDE the internalEngine and no external functions modifiy the engine/components.

gonpombo8 (changes requested):

browser-interface/packages/unity-interface/audioStream.ts
gonpombo8:
```ts
// AudioGlobal.ts
const audioMitter = mitt<AudioMitter>()
function listen(ev: AudioEvents[keyof AudioEvents]) {
  audioMitter.emit('change', { sceneId, event })
})

// InternalEngine.ts
import { audioMitter } from '../global-audio.ts'
audioMitter.on('change', ({ sceneId, event}) => { // logic for the components here })
```

nicoecheza: I agree on having all the `internalEngine` updates on the same place is better (will apply those changes), but something to note about `dont feel safe by passing the engine here` is that all the `internalEngine's` are exposed via [this](https://github.com/decentraland/unity-renderer/blob/70d2e92ef2cad427801a704e3e7401fc00967247/browser-interface/packages/shared/world/parcelSceneManager.ts#L38)

so you can actually access the engine from **anywhere**

Screenshot 2024-04-09 at 2 58 39 PM

browser-interface/packages/shared/world/runtime-7/engine.ts
```diff
@@ -47,7 +61,30 @@ function getUserData(userId: string) {
...
+
+// AudioStream updates
+audioStreamEmitter.on('changeState', ({ entityId, sceneId, state }) => {
+  const scene = getScene(sceneId)
+  const engine = getEngineFromScene(scene)
+
+  if (!engine) return
+
+  const AudioEvent = defineAudioEvent(engine)
+  AudioEvent.addValue(entityId, { state, timestamp: Date.now() })
```
gonpombo8, line 85:
```suggestion
const audioEvent = AudioEvent.get(entityId)
const lastAudio = [...audioEvent.values()].pop()
const timestamp = (lastAudio.timestamp ?? 0) + 1
AudioEvent.addValue(entityId, { state, timestamp })
```

browser-interface/packages/shared/world/runtime-7/engine.ts
```diff
@@ -47,7 +61,30 @@ function getUserData(userId: string) {
...
+
+function getEngineFromScene(scene?: SceneWorker): IEngine | undefined {
+  return scene?.rpcContext.internalEngine?.engine
+}
+
 export const localProfileChanged = mitt<LocalProfileChange>()
+export const audioStreamEmitter = mitt<AudioStreamChange>()
+
+// AudioStream updates
+audioStreamEmitter.on('changeState', ({ entityId, sceneId, state }) => {
```
gonpombo8, line 78: if you move this listener inside the engine function, there is no need to expose the sceneWorker or the engine.

You have the engine there, and all the context of that scene lives in that function

browser-interface/packages/shared/world/runtime-7/engine.ts
nicoecheza: mmmm this will not be the actual timestmap in which the state changed 🤔

browser-interface/packages/shared/world/runtime-7/engine.ts
nicoecheza: yeah, but we would be applying the new state to all the scenes for the same entity right?

gonpombo8: we can put the conditionals inside the function. Check this https://github.com/decentraland/unity-renderer/pull/6163/files#diff-1562784c24abb585b205a56650940d8c13d9740a3d19f1d1853039a4764a56b3R238

`if (sceneId !== scenId) {}`

browser-interface/packages/shared/world/runtime-7/engine.ts
gonpombo8: I think the tiemstamp is a monotonic counter for ordering the values (like the CRDT lamport timestamp) , not the actual timestamp in which the event happens.

nicoecheza (approved):

gonpombo8 (approved):

PR #6157, opened 2024-04-09
update livekit-client@2.0.10

nicoecheza:
## What does this PR change?

Updated `livekit-client` dep. Since it was a major update, there was some breaking changes that needed to be addressed.
[Migration guide](https://docs.livekit.io/guides/migrate-from-v1/)

## How to test the changes?
Realtime audio, video, etc should work as expected in explorer.


AlejandroAlvarezMelucciDCL:
Looks good to me 👍

PR #6173, opened 2024-04-18
Chore/cleanup code simple

ProteanCoding:
## What does this PR change?

* Removing dead code from a duplicated library import
* Cleanup of configuration files


aixaCode:
Hi, thanks for this contribution, this is really nice clean up. We will check it, as we need to run some checks first.

aixaCode (approved):

ProteanCoding:
Obviously this cleanup is not that simple...
The code duplication is certainly an issue, if the code that is used in `circleci` is not the code that is actively maintained.
Perhaps the live `Plugins` folder should be moved out of the `Assets` folder ?

ProteanCoding:
Going to move this code-duplication removal to its own branch, in the meantime, it should be possible to merge these remaining two simple cleanups.

aixaCode:
Could you describe why you are making this changes?

ProteanCoding:
I do believe I have been following the `CONTRIBUTING.md` guide.
I will add comments to the two changes describing their purpose.

.gitattributes
```diff
@@ -1,7 +1,7 @@
 ###############################################################################
 # Set default behavior to automatically normalize line endings.
 ###############################################################################
-* text=auto
+* text=auto eol=lf
```
ProteanCoding, line 4: On all modern OSes, modern editors (including VS Code) can deal with `lf` endings just fine, `crlf` is not needed.
If `git` is allowed to set the line endings to `crlf` before commits, this leads to discrepancies that are annoyingly useless at best.
And who knows what at worse ! Some of these files end up being used in a Docker environment, which will be expecting `lf` endings.

browser-interface/.gitignore
```diff
@@ -82,6 +82,7 @@ test/**/*.js
 test/**/*.js.map
 dist
 !packages/shared/world/runtime-7/sourcemap/source-map@0.7.4.js
+/protocol-temp/
```
ProteanCoding, line 85: This folder can remain if the build process is interrupted before the end.
Its brief appearance has also been confusing my own file monitoring process.

PR #6176, opened 2024-04-18
Chore/cleanup empty scenes build

ProteanCoding:
## What does this PR change?
Removal from the repository of the temporary Empty Scenes build files.


browser-interface/.gitignore
```diff
@@ -46,7 +46,7 @@ node_modules/
 npm-debug.log
 .DS_Store
 static/dist
-public/systems
```
ProteanCoding, line 49: This folder is no longer present

aixaCode:
How are you testing those changes? You would need to provide a bit more context why are you proposing this change
As far as I understand those empty scenes are still used, and you can't remove them

ProteanCoding:
This is meant as a follow-up to #6175.
These appear to be files that are left-over from previous builds of the Empty Scenes. They will appear again at this location during the build process.

ProteanCoding:
Apologies for the lack of context, I have been hoping to converse with the team about the proper process before these PRs.
But I have overwhelmed myself trying to resolve too many issues at the same time. I see now these all should have been kept as Drafts to start with.

browser-interface/.gitignore
```diff
@@ -46,7 +46,7 @@ node_modules/
...
+public/empty-scenes/common/contents
```
ProteanCoding, line 49: This ends up containing what are in fact temporary build files.

PR #6177, opened 2024-04-18
Removing duplicate websocket sharp

ProteanCoding:
## What does this PR change?
This is to fix code duplication.
This code has been maintained here: https://github.com/decentraland/unity-renderer/commit/9ad1866469e05da610c671bf2b6a63bd4a43759c
But the wrong version has been used here: https://github.com/decentraland/unity-renderer/blob/dev/.circleci/config.yml#L18


aixaCode:
Are you sure this is correct? If you remove `websocket-sharp-master` from `unity-renderer/Assets/Plugins`, Unity project will not compile as there is code that references it.
Also I don't see anything wrong with the project path, it's working properly for all platforms. I think we can remove the `unity-renderer/Plugins` one.
Also Examples seems to be safe to remove.

ProteanCoding:
Sorry I meant to keep this as `Draft`, I was getting overwhelmed with all the PRs on my plate, so I decided to pull this one out of https://github.com/decentraland/unity-renderer/pull/6173, so that that one at least could stay `ready to merge`.

ProteanCoding:
I am indeed not sure this is correct.
But any code duplication is in itself inherently incorrect, and really should be addressed. It makes it dangerously hard to ensure that further modifications are properly propagated.
I will create an independent Issue for this.

aixaCode:
Unfortunately, you can't run our automated checks as an external contributor, and I would need to move your branches to our repo so the CI will pick it up. I'm a bit lost with all the PRs you have opened ;)

Regarding the duplicate WebSocket Sharp library, I'm sure the project will not compile. Just to confirm, I pulled your branch and opened it in Unity, and there are quite a few compiler errors. Also, our CI has failed with compilation errors.

Screenshot 2024-04-19 161203

As a suggestion, it would be good to test your changes locally before opening a PR to catch issues like this.

ProteanCoding:
Yes as I wrote, I meant to keep this one in `Draft`. All my other PRs have been properly tested.

ProteanCoding:
So I've been reviewing my train of thought regarding this removal.
For `make build-unity-local` in the `browser-interface`'s Makefile, @dcl/explorer is simply copied out of `node_modules`.
https://github.com/decentraland/unity-renderer/blob/481ffa9c6776842caabac5179d5c3b4ffbb4ca64/browser-interface/Makefile#L72
So I assumed the `unity-explorer` project was pulled in at that point.
I now see that this in fact pulls the result of the `unity-renderer` folder, which is what is being published in https://www.npmjs.com/package/@dcl/explorer.

Talk about misnomers !!

ProteanCoding:
I hadn't even compiled that folder at that point, as it needs Unity 20`22.3.6`f1, and the `unity-explorer` project itself needs Unity 20`22.3.23`f1.
A Unity engine is a hefty download 😖

PR #6202, opened 2024-05-30
fix: handling of third-party wearable on ExtendedUrnParser class

aleortega:
## What does this PR change?

This PR addresses how third-party wearables are processed by the ExtendedUrnParser class. This class is responsible for shortening URNs when they are extended (i.e., urn.Split(':').Length > 6). When an extended URN is received, it cuts off the last part and returns the shortened version. This is necessary because the /entities/active endpoint from Catalysts' content servers expects the shortened URN to return the wearable/emote metadata.

The issue was that wearables from third-party providers contain 7 parts, causing the code to mistakenly treat them as extended URNs, which is not the case. Third-party wearables have an additional part but that part do not represent a token id (_which is the part that should be removed_), so we need to call the Catalysts endpoint using the entire URN received instead of shortening it.

The changes in this PR prevent the removal of the last part of the URN when this method receives a wearable's URN from a third-party provider (i.e., urn.Contains("collections-thirdparty")).

## Our Code Review Standards

## Copilot summary

copilot:summary


marianogoldman (approved):

aleortega:
[Testing link](https://decentraland.org/play/?explorer-branch=fix%2Ftp-wearable&CATALYST=https%3A%2F%2Fpeer-testing.decentraland.org&position=0%2C-1&island=peer-testing1w)

lorux0 (approved):

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/NFTUtils/NFTUtils.cs
```diff
@@ -143,17 +142,30 @@ public static bool TryParseUrn(string urn, out string chain, out string contract
...
-                return true;
+                // Check if the URN contains "collections-thirdparty"
+                if (contractStandardSpan.ToString().Equals(COLLECTIONS_THIRDPARTY, StringComparison.Ordinal))
+                {
+                    // 4: contract address
+                    var contractAddressSpan = urnSpan;
+                    contractAddress = contractAddressSpan.ToString();
+                    chain = chainSpan.ToString();
+
+                    return true;
```
AlejandroAlvarezMelucciDCL, line 153: I would move both `return true;` out of the if-else block to have only one exit point and make it easier to read.

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/NFTUtils/NFTUtils.cs
```diff
@@ -143,17 +142,30 @@ public static bool TryParseUrn(string urn, out string chain, out string contract
...
-
-                // 5: token id
-                var tokenIdSpan = urnSpan;
-                contractAddress = contractAddressSpan.ToString();
-                tokenId = tokenIdSpan.ToString();
-                chain = chainSpan.ToString();
-
-                return true;
+                // Check if the URN contains "collections-thirdparty"
+                if (contractStandardSpan.ToString().Equals(COLLECTIONS_THIRDPARTY, StringComparison.Ordinal))
```
AlejandroAlvarezMelucciDCL, line 146: I'm not too familiar with this method and how/where it's used, but I'm concerned that in this block we're not setting a new value for `tokenId`, is that expected for third-party collections? Can we add a note stating that for future reference?

AlejandroAlvarezMelucciDCL (commented):
I left a small comment and a question. Looks good to me but I have some concerns

unity-renderer/Assets/Scripts/MainScripts/DCL/Helpers/NFTUtils/NFTUtils.cs
aleortega: That is correct yes, Third-Party wearables do not have token id at the moment. I'll add a comment.

AlejandroAlvarezMelucciDCL (approved):

LucasLioyQA:
**QA checked.**

Please, check the following issue tracker to check which issues have been found and fixed while testing this PR and what still needs to be fixed>

https://docs.google.com/spreadsheets/d/19V7h30vyjgqmhKSjvISv7aSdIrfYxDEZ/edit#gid=1851350851

No third party emotes could be tested, since no one could be obtained.

PR #6235, opened 2024-07-10
fix: third person camera collide with character only layer as well

pravusjif:
Third person camera's Cinemachine Collider was only colliding with the `Default` physics layer but it should also collide with the `CharacterOnly` physics layer, since that one is configured when setting colliders as "PHYSICS only" through the SDK's collision layers.

### HOW TO TEST

Confirm that now the camera collides with the walls at 107,44.

BUGGED ([prod](https://play.decentraland.org/?position=107,44))

https://github.com/decentraland/unity-renderer/assets/1031741/a90e9866-828f-49b7-92bc-f6c1d421860e

FIXED (this PR)

https://github.com/decentraland/unity-renderer/assets/1031741/6bc0a3df-69b8-4a64-9289-9c632d483de1

### DIFF
--- a/.github/auto_assign_config.yml
+++ b/.github/auto_assign_config.yml
@@ -10,7 +10,6 @@ reviewGroups:
     - lorux0
     - davidejensen
     - popuz
-    - Kinerius
     - mikhail-dcl
     - pravusjif
   qa:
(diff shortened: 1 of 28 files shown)


pravusjif:
Closing this as I created it from `main`...

